PR #1, opened 2018-02-19
practice exercises: find errors in code

kabaros:


week-1-practise/1-easy/a-syntax-error-3.js
```diff
@@ -0,0 +1,10 @@
+// There is a syntax error with this code. Fix it.
+
+function trimWord(word) {
+    return word.trim;
```
djgrant: Is this the same as the one above?

week-1-practise/1-easy/a-syntax-error-2.js
```diff
@@ -0,0 +1,10 @@
+// There is a syntax error with this code. Fix it.
+
+function trimWord(word) {
+    return word.trim;
```
djgrant: Is the error that the function is not called?

week-1-practise/1-easy/c-bad-code-.js
```diff
@@ -0,0 +1,9 @@
+// This function
```
djgrant: ?

PR #3, opened 2018-02-21
Islam JS-exercises Homework

islam95:
Hello! Please check out my homework.

### DIFF
--- a/week-5-practise/js/main.js
+++ b/week-5-practise/js/main.js
@@ -1,0 +1,58 @@
+// Task 1
+var blueBtn = document.querySelector("#blueBtn");
+var orangeBtn = document.querySelector("#orangeBtn");
+var greenBtn = document.querySelector("#greenBtn");
+
+var jumbotron = document.querySelector(".jumbotron");
+var donateBtn = document.querySelector(".jumbotron .buttons a:first-of-type");
+var volunteerBtn = document.querySelector(".jumbotron .buttons a:last-of-type");
+
+blueBtn.addEventListener("click", function(){
+    jumbotron.style.backgroundColor = "#588fbd";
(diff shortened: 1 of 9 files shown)


islam95:
I will create new pull request from a different branch

islam95:
No problem.

week-3/F-array-forEach/exercise.js
```diff
@@ -9,6 +9,23 @@
 
 var arr = new Array(100);
 
+for(var i = 1; i < arr.length; i++){
+  arr[i] = i;
+}
+
+arr.forEach(function(index) {
```
djgrant: The first parameter passed to `forEach` is the array item, not the index. The second parameter is the index.

So for this exercise you can get away with not filling the array up with values, and just use the index. (The first parameter is not used).

islam95: Oh.. OK. I will try another way. It's just in MDN examples function in `forEach` has only one parameter. That's why I was confused.

islam95: Also, when I tried without filling the array, the index in `forEach` was empty. I don't know, maybe I am wrong. I'll try the way you said.

islam95: Just to clarify, do you mean I should use 2 parameters in `forEach` function?
like so `arr.forEach(function(arr, index)){};` ?

djgrant: Ah, you are correct. `forEach` does not iterate over empty array items.

It could be done this way http://jsbin.com/xupiwizaxo/edit?js,console but your original solution is fine.

islam95: Now it's easy. Done!

PR #7, opened 2018-02-24
Rukmini Homework

neelunsiri:


week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,11 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+function isStudent(name) {
+  return students.indexOf(name) > -1;
```
tekul: You could also use the array method `includes` which you used in a previous exercise.

week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,11 @@
...
+}
+
+var groupIsOnlyStudents = group.every(isStudent); // complete this statement
```
tekul: Good stuff :+1:

PR #10, opened 2018-02-24
Completed all the Home work exercises

AKFessha:


week-3/A-array-find/exercise.js
```diff
@@ -3,9 +3,11 @@
   Any result over 60 is a pass. 
   Using .find() (and any other array methods necessary), find the lowest pass score
 */
-
+function findLowestPassScore(num) {
+  return num.sort((a, b) => a - b).find(num => num > 60); //using ES6
```
petersuhm: Good job! 😎 💪

week-3/C-array-every/exercise.js
```diff
@@ -4,8 +4,13 @@
 
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
+function checkingGroup(name) {
+  if (students.includes(name)) {
+    return name;
```
petersuhm: I think this should return a boolean value.

week-3/D-array-filter/exercise.js
```diff
@@ -7,8 +7,12 @@
 */
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
-
-var pairsByIndex; // Complete this statement
+function checkingValue(value) {
+  if (value instanceof Array && value.length >= 2) {
```
petersuhm: More than 2 elements doesn't make a pair. So you need to check that it has exactly 2 elements in order for it to be a pair :)

week-3/F-array-forEach/exercise.js
```diff
@@ -8,7 +8,19 @@
 */
 
 var arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
-
+function multiples(num) {
```
petersuhm: I had to do this for my first real job interview 😆

week-3/C-array-every/exercise.js
AKFessha: No, it should return either "The group contains students only or "The group does not contain students only"  have a look at the whole code, it has a console.log statement

week-3/C-array-every/exercise.js
```diff
@@ -4,8 +4,13 @@
...
+  }
+}
 
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = group.every(checkingGroup); // complete this statement
 
 if (groupIsOnlyStudents) {
   console.log("The group contains only students");
```
AKFessha: right here

week-3/D-array-filter/exercise.js
AKFessha: Noted! thanks

week-3/C-array-every/exercise.js
petersuhm: But you're returning the name. You should be able to do this:

```
function checkingGroup(name) {
    return students.includes(name);
}
```

week-3/F-array-forEach/exercise.js
AKFessha: oh really! how did you get on with it? did you manage to secure the job?

week-3/C-array-every/exercise.js
AKFessha: yeah you are right, I can do `return students.includes(name);` instead of the if statement

week-3/D-array-filter/exercise.js
```diff
@@ -8,7 +8,7 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 function checkingValue(value) {
-  if (value instanceof Array && value.length >= 2) {
+  if (value instanceof Array && value.length === 2) {
```
petersuhm: Exactly :)

week-3/C-array-every/exercise.js
```diff
@@ -5,9 +5,7 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 function checkingGroup(name) {
-  if (students.includes(name)) {
-    return name;
-  }
+  return students.includes(name);
```
petersuhm: 👍

PR #11, opened 2018-02-24
swazz homework

swazz:


week-1/J-functions/exercise2.js
```diff
@@ -1,7 +1,32 @@
...
-var result = triple(12);
-
+var result = halve(12);
 console.log(result);
+
+function triple(result) {
+  return result * 3;
+}
+var results = triple(8);
+console.log(results);
```
zero-point: This is good. Just so you know, you can also log the result of the function to console directly. E.g. console.log(triple(8));

week-1/K-functions-parameters/exercise3.js
```diff
@@ -1,5 +1,9 @@
 // Write your function here
-
-var greeting = createGreeting("Daniel");
+function createGreeting(name, age) {
+  return "hello,my name is " + name;
+  "and i am old" + age;
```
zero-point: This is a good solution to the exercise. But I see you also wanted to log the age. To do this, instead of the ";" after name you need a "+" to concatenate the strings together.

week-1/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,23 @@
+function percentage() {
```
zero-point: This exercise hasn't been finished, it still requires a bit of work to understand how to use functions. I recommend looking at your solution for the exercise at I-floats/exercise.js and then thinking about how to place that code in separate functions: one for calculating percentages and one for creating the message. You should also attempt exercise2: https://github.com/CodeYourFuture/js-exercises/tree/master/week-1/L-functions-nested

week-2/G-conditionals-2/exercise-1.js
```diff
@@ -7,7 +7,9 @@
 */
 
 function negativeOrPositive(number) {
-
+  if (number >= 0) return "postive";
+  else number <= 0;
```
zero-point: This is not quite right, if-else statements have the following shape:

```
if (condition) {
    block of code to be executed if the condition is true
} else { 
    block of code to be executed if the condition is false
}

```
In your code, when the condition is true, the code is correct. But when the condition is false, it tries to execute "number <= 0" which is wrong and uneccessary (you've already checked through your if condition whether it's >=0 and it only goes to the else block if it failed, so no need to check again). It then exits the else block (because there are no curly brackets, then only one statement is within the else block) and only after it's out of the if-else block it returns "negative" (which again is not right, it should return "negative" within the if-else block).

If this doesn't make sense, please let me know and we'll go over it together :)

week-2/G-conditionals-2/exercise-2.js
```diff
@@ -8,7 +8,9 @@
 */
 
 function studentPassed(grade) {
-
+  if (grade >= 50) return "pass";
+  else grade < 50;
```
zero-point: Same problem here as in the previous exercise, it's not a big problem, it just needs a bit more attention. You clearly understand the theory of conditionals but need more help with writing the actual code for it. Otherwise it would cause you problems later on when conditionals become more complicated.

week-2/G-conditionals-2/exercise-3.js
```diff
@@ -9,7 +9,10 @@
 */
 
 function calculateGrade(mark) {
-
+  if (mark >= 80) return "A";
+  else if (mark >= 60 && mark < 80) return "B";
+  else if (mark >= 50 && mark < 60) return "C";
+  else if (mark < 50) return "Fail";
```
zero-point: This is good, a significant improvement from the previous exercises on conditionals. :+1: 

I would still improve this by removing the last if, as it is uneccessary (if it reached the last else condition then you already know that it is less than 50 so no need to have another check). Instead of a "else if", it can simply be an "else".

week-2/K-array-methods/exercise2.js
```diff
@@ -7,7 +7,7 @@
 var mentors = ["Daniel", "Irina", "Rares"];
 var students = ["Rukmini", "Abdul", "Austine", "Swathi"];
 
-var everyone; // complete this statement
+var everyone = [mentors] + [students]; // complete this statement
```
zero-point: Concatenation of strings is done with "+", but for arrays such as mentors and students the way to concatenate is to use concat(). For example: [1, 2, 3].concat([4, 5, 6]);

week-1-practise/level-3/1-piping.js
```diff
@@ -11,6 +11,9 @@
 */
 
 var startingValue = 5;
+//var result = (format(startingValue*2)+10);
+var result = format(multiply(startingValue, 2) + add(0, 10));
```
zero-point: Pretty good! The only thing that I would change is that you shouldn't do "+ add(0, 10)" to add 10, it's a bit redundant to do it that way. Instead you should do this add(multiply(startingValue, 2), 10) so that the add function actually takes the multiplied value as its first parameter and 10 as its second parameter.

PR #12, opened 2018-02-24
Lesley week 1 first

LslyKng:


week-2/D-predicates/exercise.js
```diff
@@ -2,16 +2,19 @@
...
 
+function isNegative(number) {
+ // return typeOf number === "number";
+    return number<0;
 }
 
 // Finish the predicate function to test if the passed number is between 0 and 10
-function isBetweenZeroAnd10(number) {
+  function isBetweenZeroAnd10(number) {
+  var number = 0
```
djgrant: This function should `return` either `true` or `false` depending on whether `number` is between zero and 10 or not.

You can use an if statement to check if `number` passes the condition. 

``` 
if (condition) { 
  return true;
} else {
  return false;
}
```

What should the condition be?

There are actually two conditions: 
- bigger than or equal to zero
- less than or equal to 10

Both those conditions should be true. What logical operator can you use for that?

week-2/E-conditionals/exercise.js
```diff
@@ -5,14 +5,15 @@
...
 
+
 var name = "Daniel";
 var role = "mentor";
 
 var name2 = "Harun";
 var role2 = "student";
-
-
+ {
```
djgrant: Don't need the braces (`{` `}`) here.

week-2/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return arr [0]; // complete this statement
 }
 
 function last(arr) {
-  return; // complete this statement
+  return arr [2] ; // complete this statement
```
djgrant: What if the parameter `arr` passed to this function has 4 items? Would this still work?

Try to re-write this without hardcoding the last value.

LslyKng:
Thanks

week-2-practise/level-1/5-indexOf.js
```diff
@@ -3,12 +3,12 @@
...
+var index = itemToFind.indexOf (7); // ONLY EDIT THIS LINE
 
 console.log(index);
 
 /* 
   EXPECTED RESULT
   ---------------
   3
-*/
+*
```
tekul: Does this work without any errors if you run it? You've accidentally deleted the `/` from the comment end marker.

week-2-practise/level-1/6-map.js
```diff
@@ -11,7 +11,7 @@ function tidyUpString(str) {
 }
 
 var mentors = ["/Daniel ", "irina ", " Gordon", "ashleigh "];
-var mentorsTidy; // ONLY EDIT THIS LINE
+var mentorsTidy = str.map (tidyUpString); // ONLY EDIT THIS LINE
```
tekul: What's `str` here? You need to call `map` on the array and it will apply the function to each member.

week-2/D-predicates/exercise.js
```diff
@@ -8,13 +8,13 @@
...
 function isNegative(number) {
- // return typeOf number === "number";
+ // return typeof number === "number";
     return number<0;
 }
 
 // Finish the predicate function to test if the passed number is between 0 and 10
   function isBetweenZeroAnd10(number) {
-  var number = 0
+    var number = (5)
```
tekul: This needs to check whether `number > 0` and also `number < 10`, so we need to combine two boolean conditions using the "and" operator (which is called `&&` in Javascript).

week-2/F-logical-operators/exercise.js
```diff
@@ -11,6 +11,10 @@ var cssLevel = 4;
 
 // Finish the statement to check whether HTML, CSS knowledge are above 5
 // (hint: use the comparison operator from before)
+functions satisfiesRequirements(num) {
+  if (num>5) {return true}
+  return false 
+}
 var htmlLevelAbove5;
```
tekul: This exercise is similar to the above in that it wants you to combine booleans using the operators:

- and (`&&`) 
- or (`||`)

week-2/F-logical-operators/exercise2.js
```diff
@@ -5,18 +5,34 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(number) {
+  if (number<0) {return true}
+  else {return false}
+}
+
+//function isBetween5and10(number)
```
tekul: You've commented out the function names here? `//` is a line comment marker.

Also, remember what we said on Saturday about what to do if you find yourself writing:

```
if (something) {
    return true;
} else {
    return false;
}
```
:-) ?

week-2/M-array-map/exercise.js
```diff
@@ -10,8 +10,8 @@ function double(num) {
 }
 
 var numbers = [1, 2, 3, 4];
-var numbersDoubled; // complete this statement (use map and the double function)
-
+var numbersDoubled = numbers.map (double); // complete this statement (use map and the double function)
```
tekul: Looks good :)

week-2/L-array-methods-2/exercise2.js
```diff
@@ -7,7 +7,11 @@
   Tip: use the string method .split() and the array method .join()
 */
 
-function capitalise(str) {}
+function capitalise(str) {
+  var str = split (separator) 'h');
```
tekul: I'm a bit confused here. What's this doing? It's quite a tricky problem. A good idea is to search for the functions it suggests to use. I just did and I found [this page](https://www.w3schools.com/jsref/jsref_split.asp), which also has a useful tip:

> Tip: If an empty string ("") is used as the separator, the string is split between each character.

Playing with that and the `join` function in node:

```
> "daniel".split("");
[ 'd', 'a', 'n', 'i', 'e', 'l' ]
> "daniel".split("").join();
'd,a,n,i,e,l'
> "daniel".split("").join("");
'daniel'
```
So if you can find out how to convert a string to upper case, you can split the string into an array  of characters, upper case the first element of the array, then join them back together.

LslyKng:
Thanks Luke I'll have a look tomorrow

LslyKng:
In week-2/J-array-get-set/exercise.js
<https://github.com/CodeYourFuture/js-exercises/pull/12#discussion_r171475682>
:


i stole this from internet and it works but I don't know why


function first(arr) {
return arr[0]; // complete this statement
}

function last(arr) {
return arr.slice(-1)[0]; // complete this statement
}

>  }

 function last(arr) {
+  return arr [2] ; // complete this statement

What if the parameter arr passed to this function has 4 items? Would this
still work?

Try to re-write this without hardcoding the last value.

----------------------------------------------------------------------------------------------------

Both those conditions should be true. What logical operator can you use for
that?

condition && condition

PR #13, opened 2018-02-24
dany's homework javacript-exercises

dany2kul:


week-1/E-strings/exercise.js
```diff
@@ -1,3 +1,5 @@
 // Start by creating a variable `message`
-
-console.log(message);
+var message="this is a string";
+var messageType=typeof message;
+console.log(messageType); 
+console.log(messageType);
```
panditita: Your code is good but the expected result for this exercise is to log the message and the type, you're consol logging the type twice

panditita: Your code is good but the expected result for this exercise is to log the message and the type, you're consol logging the type twice

week-1/F-strings-concatenation/exercise.js
```diff
@@ -1,3 +1,5 @@
 // Start by creating a variable `message`
-
-console.log(message);
+var greetingStart="hi,i'm called ";
+var name="daniella";
+var greeting=greetingStart + name;
+console.log(greeting);
```
panditita: It's a good practice to leave a space after the equal sign i.e. greetingStart = "hi, i'm called ";

dany2kul: notes taken.thanks

PR #15, opened 2018-02-24
my javascript home work

okolochuks:


week-2/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount > mentorCount;
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = roomMaxCapacity >= "studens and mentors";
```
jswiss: here you need to reuse the studentCount and mentorCount variables from above

week-2/E-conditionals/exercise.js
```diff
@@ -3,19 +3,12 @@
...
-
 var name2 = "Harun";
 var role2 = "student";
 
-
-console.log("Hi, My name is " + name);
-console.log("Hi, My name is " + name2);
-
-
+if (role2) {
```
jswiss: This is saying, 'if role2 exists, console.log the below string'. Instead, you want to ask whether role2 === 'student', right?

week-2/F-logical-operators/exercise.js
```diff
@@ -6,19 +6,20 @@
 */
 
 // Do not change these two statement
-var htmlLevel = 8;
+var htmlLevel = 10;
```
jswiss: I think the instructions say to not change these statements :)

week-2/F-logical-operators/exercise2.js
```diff
@@ -5,7 +5,14 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(value) {
+return typeof value === nunber;
```
jswiss: when using `typeof`, you need to set the value equal to "number", not nunber

week-2/G-conditionals-2/exercise-1.js
```diff
@@ -7,9 +7,17 @@
 */
 
 function negativeOrPositive(number) {
+  if (number >= 0) {
+    return "positive";
+  }
 
+  if (number < 0) {
+    return "negative";
```
jswiss: good work here

week-2/L-array-methods-2/exercise.js
```diff
@@ -15,8 +15,8 @@ var everyone = [
   "Swathi"
 ];
 
-var firstFive; // complete this statement
-var lastFive; // complete this statement
+var firstFive = everyone.slice(0, 5); // complete this statement
+var lastFive = everyone.slice(2, 7); // complete this statement
```
jswiss: Like we discussed in Mitchell, here you can just pass things that will make your code more reusable. Rather than counting the length of an array, try using everyone.length to know where to end. You can also use that to specify where to start to (for example, if you are trying to return the last five, you could do `everyone.slice(everyone.length-5, everyone.length);`

week-2/L-array-methods-2/exercise2.js
```diff
@@ -8,12 +8,13 @@
 */
 
 function capitalise(str) {}
+var name = ["D", "a", "n", "i", "e", "l"];
+var arr = name.join();
```
jswiss: remember, you have to pass WHAT you want it to be joined on. instead of `.join()`, try `.join('')`

week-2/L-array-methods-2/exercise3.js
```diff
@@ -6,8 +6,12 @@
 
 var ukNations = ["Scotland", "Wales", "England", "Northern Ireland"];
 
-function isInUK(country) {
-  return; // complete this statement
+function isInUk(country) {
+  if (ukNations) {
+    return ukNations.includes(country); // complete this statement
```
jswiss: Great work here!

PR #16, opened 2018-02-24
week 1 soheila azadi

azadi11:


week-2/B-boolean-literals/exercise.js
```diff
@@ -6,6 +6,7 @@
 */
 
 var codeYourFutureIsGreat = true;
+var mozafarIsCool = true;
```
tekul: I think you're missing a few here. Does your program give the expected result from the comment below?

week-2/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
...
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = roomMaxCapacity >= studentCount + mentorCount; // finish this
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = personA !== personB; // finish this statement
```
tekul: Great. This all looks very good :-).

week-2/I-array-properties/exercise.js
```diff
@@ -6,7 +6,13 @@
 */
 
 function isEmpty(arr) {
-  return; // complete this statement
+  if (arr.length == 0) {
+    return "empty";
+  } else {
+    return "notEmpty";
+  }
+
```
tekul: This is doing the right thing, but we want it to return a boolean value (`true` or `false`), not a string - see the comment showing the "Expected Result" from running the program.

week-2/F-logical-operators/exercise.js
```diff
@@ -27,10 +27,7 @@ var cssOrHtmlAbove5;
 console.log("Is Html knowledge above 5?", htmlLevelAbove5);
 console.log("Is CSS knowledge above 5?", cssLevelAbove5);
 console.log("Is Html And CSS knowledge above 5?", cssAndHtmlAbove5);
-console.log(
-  "Is either Html or CSS knowledge above 5?",
-  cssOrHtmlAbove5
-);
+console.log("Is either Html or CSS knowledge above 5?", cssOrHtmlAbove5);
```
tekul: Did you get stuck here? This exercise is for combining boolean values using "and" and "or". These are written `&&` and `||` in Javascript. e.g.

```
var imHungry = true;
var imThirsty = false;

var imHungryAndThirsty = imHungry && imThirsty;
```

week-2-practise/level-1/2-validation.js
```diff
@@ -2,7 +2,7 @@
 // TIP: You should write an expression that returns a boolean value
 
 var num = 10;
-var isBigEnough; // ONLY EDIT THIS LINE
+var isBigEnough = true; // ONLY EDIT THIS LINE
```
tekul: This will work but is cheating :slightly_smiling_face: . It says you should write "an expression" (using `num`). That means something like `num === 5` or `num > 5`.

week-2-practise/level-2/1-boolean.js
```diff
@@ -3,11 +3,12 @@
...
-  console.log("I am happy");
-} else {
-  console.log("I am not happy");
+{
+  if (isHappy) {
+    console.log("I am happy");
+  } else {
+    console.log("I am not happy");
+  }
 }
```
tekul: You don't need the outer `{}` here. This code isn't inside a function.

PR #17, opened 2018-02-24
week one homework 

Omar-Suliman:


week-2-practise/level-2/2-validation.js
```diff
@@ -4,7 +4,17 @@
...
+  if (num==10){
+    return (true);
+  }if (num ===10.5){
+return (true);
+  }if (num<=101){
+    return(false);
+  }else{
+    return (false);
+  }
+}
```
tekul: The idea is to write a function that checks that `num`:

- is a number
- is a positive number
- is less than or equal to 100

So you shouldn't need to check individual values like `10` or `10.5`. Also it says to write other small functions for each part. Then you would call those functions inside `validate`.

week-2-practise/level-2/3-sort.js
```diff
@@ -2,9 +2,8 @@
 // It should contain the values of `letters` and `letters2` in ascending order
 
 var letters = ["a", "n", "c", "e", "z", "f"];
-var letters2 = ["w", "b", "v", "g", "l", "o"];
-var sortedLetters; // ONLY EDIT THIS LINE
-
+var letters2 = [ "w", "b", "v", "g", "l", "o"];
+var sortedLetters=letters.concat(letters2);// ONLY EDIT THIS LINE
 console.log(sortedLetters);
```
tekul: Does this give the expected result when you run it? If not, what's missing?

week-2-practise/level-2/4-slice.js
```diff
@@ -5,7 +5,8 @@
 // - remove the item at the specified index
 
 function remove(arr, index) {
-  return; // complete this statement
+  if (newNames=names.remove())
```
tekul: This doesn't look like it will work. Hint: the name of the file should give you an idea what standard function to use.

week-2/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
...
+var moreStudentsThanMentors = studentCount > mentorCount;
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = roomMaxCapacity >= studentCount;
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = "Daniel" == "Irina";
```
tekul: Good stuff. Remember to always use `===` rather than `==`.

week-2/D-predicates/exercise.js
```diff
@@ -7,12 +7,14 @@
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
 function isNegative(number) {
-
+  return number < 0;
 }
 
 // Finish the predicate function to test if the passed number is between 0 and 10
 function isBetweenZeroAnd10(number) {
+  return 0<=10;
```
tekul: This will always return `true` (zero is always "less than or equal" to 10). You need to check two things. Do you remember the operator for "and"? It's `&&`. So `a && b` is true if `a` is true and `b` is true.

week-2/E-conditionals/exercise.js
```diff
@@ -10,10 +10,14 @@ var role = "mentor";
...
 var name2 = "Harun";
 var role2 = "student";
+if (name2){
+  return"hi , name ,is "+name2;
+}
 
 
 console.log("Hi, My name is " + name);
 console.log("Hi, My name is " + name2);
+console.log("hi, we are student "+ role2);
```
tekul: Does this only show the student name if you run it? Also `if(name2)` won't do anything.

week-2/F-logical-operators/exercise.js
```diff
@@ -8,11 +8,15 @@
...
+  }
+}
 
-// Finish the statement to check whether HTML, CSS knowledge are above 5
-// (hint: use the comparison operator from before)
-var htmlLevelAbove5;
-var cssLevelAbove5;
+var htmlLevelAbove5 = 8;
+var cssLevelAbove5 = 4;
+function satisfiesRequirements(cssAndHtmlAbove5)
```
tekul: What happens when you run this? You've defined `satisfiesRequirements` twice but it isn't used. I don't think you need to write a function here. Just set the values (check the "hint" in the comment). They should be boolean values (true or false), rather than numbers.

week-2/G-conditionals-2/exercise-1.js
```diff
@@ -7,9 +7,16 @@
 */
 
 function negativeOrPositive(number) {
-
+  if (number>=5){
```
tekul: Do you need to check the value `5` here? Remember that the definition of a function should work for all its inputs, rather than just the calls we are using below to check it. We are calling `negativeOrPositive(5)` but it could easily be any other number (e.g `negativeOrPositive(4)`).

week-2/G-conditionals-2/exercise-2.js
```diff
@@ -8,6 +8,14 @@
 */
 
 function studentPassed(grade) {
+  if (grade<=49){
+  return"failed";
+  }
+  if (grade>=50){
+return"passed";
+  }if (grade==100){
```
tekul: Will the condition `grade == 100` ever happen? Trace through the function assuming `grade` is 100 and work out what the function will do.

week-2/G-conditionals-2/exercise-2.js
```diff
@@ -8,6 +8,14 @@
 */
 
 function studentPassed(grade) {
+  if (grade<=49){
```
tekul: It's probably easier to write `grade < 50` here since that's what the requirements state. The effect is the same though (assuming only whole numbers are used for marks). Otherwise what would happen if `grade` was `49.5`?

week-2/I-array-properties/exercise.js
```diff
@@ -5,10 +5,7 @@
 
 */
 
-function isEmpty(arr) {
-  return; // complete this statement
-}
-
+function isEmpty(arr) {}
```
tekul: There's no function implementation here.

PR #23, opened 2018-02-24
Last modified Week2JS_exercises

shawki2:
Updated all week2JS and using two solutions for capitalise function.


week-1/E-strings/exercise.js
```diff
@@ -1,3 +1,11 @@
 // Start by creating a variable `message`
+var messageType = typeof message;
+var message = "This is a string",
```
djgrant: Good to see you experimenting here 👍

week-1/K-functions-parameters/exercise4.js
```diff
@@ -1,5 +1,7 @@
 // Declare your function first
-
+function add(a, b) {
+  return (sumab = a + b);
```
djgrant: `return a + b` is sufficient here.

In JavaScript, when you assign a value to a variable without using the `var` keyword, it actually creates a global variable, which are definitely best to avoid! In this line of code you are creating a global variable `sumab`.

week-1/K-functions-parameters/exercise5.js
```diff
@@ -1,5 +1,7 @@
 // Declare your function here
-
+function createLongGreeting(name, age) {
+  return "Hello, my name is " + name + " and I'm " + age + " years old.";
```
djgrant: Another way to do this is to use **template literals**.

```js
return `Hello, my name is ${name} and I'm ${age} years old`;
````

File: week-1/L-functions-nested/exercise.js [hunk 4]
```diff
@@ -1,5 +1,20 @@
...
+    mentorCount / (mentorCount + studentsCount) * 100
+  );
+  var studentsPercent = Math.round(
+    studentsCount / (mentorCount + studentsCount) * 100
+  );
+  /*console.log(
+    "mentorPercent = " + mentorPercent,
+    "studentsPercent = " + studentsPercent
+  );*/
+  printMessage(mentorPercent, studentsPercent);
```
djgrant: Whenever possible, try to ensure that your functions return a value and do not perform side effects. 

In this example it would be better if the `percentage` function returns a percentage value. You could then get the value, assign it to a variable, and use `console.log` to print it.

week-1/L-functions-nested/exercise2.js
```diff
@@ -0,0 +1,20 @@
...
+var mentor5 = "Yohannes";
+var mentor1 = "Daniel";
+var mentor2 = "Irina";
+var mentor3 = "Mimi";
+var mentor4 = "Rob";
+var mentor5 = "Yohannes";
+
+var mentors = ["Daniel", "Irina", "Mimi", "Rob", "Yohannes"];
+console.log(mentors);
+for (i = 0; i < mentors.length; i++) {
```
djgrant: Nice!

You might also like to try `mentors.forEach(printMessageFunction)` where `printMessageFunction` is a function you already defined.

Just remember, only use `forEach` if you want to perform a side effect, like printing something to the console. Where you intend to manipulate an array or produce a new value, use `map`, `filter`, `find` etc. (We'll cover them in more depth next week.

week-2/G-conditionals-2/exercise-2.js
```diff
@@ -8,7 +8,11 @@
 */
 
 function studentPassed(grade) {
-
+  if (grade >= 50) {
+    return (grade = "passed");
```
djgrant: Do not mutate (re-assign) the parameter passed to the function, as this will change it outside the function too.

`return "passed"` is sufficient here.

week-2/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return arr[0]; // complete this statement
 }
 
 function last(arr) {
-  return; // complete this statement
+  return arr[2]; // complete this statement
```
djgrant: What if the parameter `arr` passed to this function has 4 items? Would this still work?

Try to re-write this without hardcoding the last value.

week-2/L-array-methods-2/exercise2.js
```diff
@@ -7,8 +7,12 @@
   Tip: use the string method .split() and the array method .join()
 */
 
-function capitalise(str) {}
-
+function capitalise(str) {
+  var character1 = str.split("");
```
djgrant: `letters` would be a more descriptive name for this variable.

Otherwise, good solution!

week-2/L-array-methods-2/exercise2_anotherSolution.js
```diff
@@ -0,0 +1,29 @@
...
+  Complete the capitalise function 
+  It should return a string with the first letter in uppercase
+  For example, capitailise("hello") should return "Hello"
+  Tip: use the string method .split() and the array method .join()
+*/
+//This solution is using methods .splite() and .slice()
+function capitalise(str) {
+  var character1 = str.split("");
+  var tLength = str.length;
+  var str = character1[0].toUpperCase() + str.slice(1, tLength);
```
djgrant: I hadn't thought of writing the solution in this way. Nice work!

shawki2:
Could you please review it?

PR #26, opened 2018-02-28
Week-1-js

Matl2018:


week-1-practise/level-1/8-function-output-1.js
```diff
@@ -1,12 +1,12 @@
...
 var result = getNumber();
 console.log(result);
 
 // Answer these questions:
-// - What does this program do?
-// - How many parameters does the function take?
-// - What is the function name?
-// - Where is the function called? with what parameters?
\ No newline at end of file
+// - What does this program do? Returns a random number between 0 and i (excluded ) then multiply it to 10
```
tekul: What is "i" here ?

week-1-practise/level-1/8-function-output-1.js
```diff
@@ -1,12 +1,12 @@
...
+// - How many parameters does the function take? 1
```
tekul: Is this correct? How many parameters are being passed when you call `Math.random()` ?

week-1-practise/level-1/8-function-output-1.js
```diff
@@ -1,12 +1,12 @@
...
+// - What is the function name? getNumber
+// - Where is the function called? with what parameters? at five .
```
tekul: This is correct, but when we are talking about a position in a source file we usually refer to the line number. So we would say something like "The function is called at line 5".

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -1,14 +1,14 @@
...
-// - Where is the function called? with what parameters?
-// - What does the function return?
-// - How can this function be improved?
\ No newline at end of file
+// - What does this program do? Run it and see, or Google some of the methods used ot understand them. answer: Concatenation
+// - How many parameters does the function take? 2
+// - What is the function name? s
+// - Where is the function called? with what parameters?at 5 with parameter code , your
+// - What does the function return? codeyour
+// - How can this function be improved? lacks "future", a third parameter  :)
```
tekul: I'd also probably say that `s` isn't  a very good choice of name for a function. It should really describe what the function does.

week-1-practise/level-2/3-currency-conversion.js
```diff
@@ -5,7 +5,9 @@
   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
 */
 
-function convertToUSD() {}
+function convertToUSD(price) {
+  return "$".concat(price / 1.4);
```
tekul: This is supposed to convert from pounds to dollars, so for example £10 will become $14 (the dollar value should be larger). If a function doesn't provide examples, then it's a good idea to think some up for yourself and check if they are correct :-). The code is all good though.

week-1/I-floats/exercise.js
```diff
@@ -1,2 +1,9 @@
 var numberOfStudents = 15;
 var numberOfMentors = 8;
+var TotalStudentsAndMentors = numberOfMentors + numberOfStudents;
+var MentorOverTotatal = numberOfMentors / TotalStudentsAndMentors;
+var percentMentors = MentorOverTotatal * 100 + "%";
+var PrecisePecent = Math.round(MentorOverTotatal * 100);
+
+console.log(percentMentors);
+console.log(PrecisePecent);
```
tekul: All looks good. The only comment I'd make here is that we usually start variables with a lower case letter by convention. So it would be `totalStudentsAndMentors` etc.

week-1/J-functions/exercise2.js
```diff
@@ -1,7 +1,24 @@
...
 console.log(result);
+
+var result2 = triple(2.5);
+console.log(result2);
+
+function addition(number1, number2) {
+  var sum = number1 + number2;
+  return sum;
+
+  // complete function here
```
tekul: Perhaps delete the comment once you have completed the function.

week-1/L-functions-nested/exercise2.js
```diff
@@ -0,0 +1,18 @@
+/*HELLO DANIEL
+HELLO IRINA
+HELLO MIMI
+HELLO ROB
+HELLO YOHANNES */
+
+function shoutOuts(name) {
+  return "Hello" + name.toUpperCase();
```
tekul: It's probably better to put the space after "Hello" - i.e. `"Hello "` rather than having to pass it with every name.

week-2/B-boolean-literals/exercise.js
```diff
@@ -6,6 +6,9 @@
 */
 
 var codeYourFutureIsGreat = true;
+var mozafarIsCool = false;
+caluclationCorrect = true;
```
tekul: Make sure to use `var` when declaring your variables. Otherwise they are [global](https://www.w3schools.com/js/js_scope.asp), which isn't a good thing :-).

week-2/L-array-methods-2/exercise2.js
```diff
@@ -10,6 +10,11 @@
 function capitalise(str) {
   var var1 = str.charAt(0);
   var var2 = var1.toUpperCase();
+  var var4 = str.charAt(str.length - 1);
+  var var5 = var4.toUpperCase();
+
+  var6 = str;
```
tekul: What are var4, var5 and var6 for here? Do they affect the result?

week-1-practise/level-3/1-piping.js
```diff
@@ -10,8 +10,6 @@
   As a learning exercise (you wouldn't do this normally), do all of the above in 1 line of code 
 */
 
-var startingValue = 5;
```
tekul: You are given `startingValue` in the example here and all the functions you need. The exercise is just supposed to show how you can chain function calls together. So you can do something like `add(multiply(startingValue, 2), 10)`. Think about how you would continue from there.

week-2-practise/level-3/2-validation.js
```diff
@@ -6,7 +6,11 @@
 
 var excludedNums = [6, 14, 91, 111];
 
-function validate(num) {}
+function validate(num) {
+  return (
+    typeof num === "number" && num.isInteger() === true && num !== excludedNums
```
tekul: I don't think this will work - `excludedNums` is an array and `num` is a number, so you can't compare them directly. You need to check whether `num` is contained in the array (we saw a function to check whether something is in an array, or you can do a Google search for one).

week-2-practise/level-3/4-slice.js
```diff
@@ -5,7 +5,10 @@
 // - insert the replacement value at the provided index
 
 function replace(arr, index, value) {
-  return; // complete this statement
+  var left = arr.slice(0, index);
+  var right = arr.slice(index + 1);
+
+  return left.concat([value].concat(right));
```
tekul: Good. You can also use `push` to add a single element to the end of an array.

week-3/A-array-find/exercise.js
```diff
@@ -4,8 +4,18 @@
   Using .find() (and any other array methods necessary), find the lowest pass score
 */
 
+function findLowestPassScore(scores) {
```
tekul: `findLowestPassScore` isn't a good name for this function, given what it does - it takes a single score and returns `true` if it is a pass and `false` otherwise. What would be a better name. The parameter name is also confusing, since it is plural.

week-3/B-array-some/exercise.js
```diff
@@ -21,4 +21,8 @@ var pairs = pairsByIndex.map(function(indexes) {
   return [student, mentor];
 });
 
-console.log(pairs);
```
tekul: You've changed this line :slightly_smiling_face: . Re-read the instructions at the top of the file and make sure the program does what it's expected to do when you run it. I think there was a previous example on how to exit a program, but you can search for it too.

week-3/A-array-find/exercise.js
Matl2018: Yeah, i reckon  "isPass " would be a better name for the function. 
I have also changed the parameter name to singular , "score".
I will push the changes soon .

PR #31, opened 2018-03-03
Harun's homework

tiredsoldier:
Due to some git problems, we deleted the whole repo and started again. This is just one exercise to show the process.

### DIFF
--- a/week-1-practise/level-1/1-syntax-error-1.js
+++ b/week-1-practise/level-1/1-syntax-error-1.js
@@ -1,14 +1,13 @@
 // There is a syntax error in this code. Fix it.
 
-function addNumbers(a b c) {
-    return a + b + c;
+function addNumbers(a, b, c) {
+  return a + b + c;
 }
 
 var result = addNumbers(1, 3, 4);
 console.log(result); // Expected result 8 - Run the program and make sure it works
 
-
 // Answer these questions:
-// 1. How many parameters does the function take?
-// 2. What is the function name?
-// 3. Where is the function called? with what parameters?
+// 1. How many parameters does the function take?   3
+// 2. What is the function name? addNumbers
+// 3. Where is the function called? with what parameters? console.log(result); parametre is result.
--- a/week-1-practise/level-1/8-function-output-1.js
+++ b/week-1-practise/level-1/8-function-output-1.js
@@ -1,5 +1,5 @@
 function getNumber() {
-    return Math.random() * 10;
+  return Math.random() * 10;
 }
 
 var result = getNumber();
@@ -7,6 +7,10 @@ console.log(result);
 
 // Answer these questions:
 // - What does this program do?
+// - create a number
 // - How many parameters does the function take?
-// - What is the function name?
+// - there aren't parameters.
+// - What is the funcnodetion name?
+// - getNumber
 // - Where is the function called? with what parameters?
+// - console.log(result); // parameter is result
--- a/week-1/H-numbers/exercise.js
+++ b/week-1/H-numbers/exercise.js
@@ -1,1 +1,10 @@
 // Start by creating a variables `numberOfStudents` and `numberOfMentors`
+var numberOfStudents = 15,
+  numberOfMentors = 8;
+var sum = numberOfStudents + numberOfMentors;
+var messagestudent = "Number of students:",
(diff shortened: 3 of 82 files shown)


week-2-practise/level-3/5-indexOf.js
```diff
@@ -6,7 +6,9 @@
 // - return a new array with the replacement value inserted
 
 function replace(arr, valueToReplace, newValue) {
-  return; // complete this statement
+  var cashe = arr.slice();
```
tekul: Do you need to call `slice()` on `arr` here? What is the result?

week-2-practise/level-3/3-sort.js
```diff
@@ -4,7 +4,9 @@
 
 var nums = [10, 1, 5, 29, 100];
 var nums2 = [11, 6, 3, 29, 12];
-var sortedNums; // complete this statement
+var sortedNums = nums.concat(nums2).sort(function(a, b) {
```
tekul: Good stuff. This is correct (and useful to know if you want to use a different sort order), but you can just use `sort()` here without a custom sorting function, since the default behaviour is the same.

Omar-Suliman: thanks

week-1/H-numbers/exercise.js
```diff
@@ -1 +1,10 @@
 // Start by creating a variables `numberOfStudents` and `numberOfMentors`
+var numberOfStudents = 15,
+  numberOfMentors = 8;
```
tekul: You should use `var` for declaring variables. Otherwise Javascript treats them as global.

week-1-practise/level-1/8-function-output-1.js
```diff
@@ -1,12 +1,16 @@
...
 // - What does this program do?
+// - create a number
 // - How many parameters does the function take?
-// - What is the function name?
-// - Where is the function called? with what parameters?
\ No newline at end of file
+// - there aren't parameters.
+// - What is the funcnodetion name?
+// - getNumber
+// - Where is the function called? with what parameters?
```
tekul: I would just say "The function is called at line 5".

week-1-practise/level-1/1-syntax-error-1.js
```diff
@@ -1,14 +1,13 @@
...
 
-
 // Answer these questions:
-// 1. How many parameters does the function take?
-// 2. What is the function name?
-// 3. Where is the function called? with what parameters?
\ No newline at end of file
+// 1. How many parameters does the function take?   3
+// 2. What is the function name? addNumbers
+// 3. Where is the function called? with what parameters? console.log(result); parametre is result.
```
tekul: I'd say "It's called at line 7, with parameters 1, 3 and 4".

week-1/L-functions-nested/exercise.js
```diff
@@ -3,3 +3,20 @@ var mentor2 = "Irina";
...
+  var mentorpercent = Math.round(
+    mentorcount / (mentorcount + studentscount) * 100
+  );
+  var studentspercent = Math.round(
+    studentscount / (mentorcount + studentscount) * 100
+  );
+
+  console.log("Percentage mentors: " + mentorpercent + "%");
+  console.log("Percentage students: " + studentspercent + "%");
+}
```
tekul: Looking at the instructions, I think the idea is that you write a reusable function that calculates a percentage, given a number and a total, rather than calculating both percentages in a single function. It also says you should log the messages separately (you're logging the values inside the function which does the calculation).

week-1/H-numbers/exercise.js
tiredsoldier: I could use var but when I pressed the Ctrl + S VS Code changed to look like that. 
The other one thing You can use var on the one line for declaring more than one variables.

tekul: You're right. I didn't notice you were using commas `,`, instead of semicolons `;` :slightly_smiling_face: .

week-2-practise/level-3/3-sort.js
tekul: Actually I'm wrong. Javascript is confusing and sorts numbers as if they are strings by default, so you do need to pass a function to it.

Issue #21, opened 2018-02-24
I-array-properties (2)


LslyKng:
Apparently this is wrong - was (arr) and didn't work so mentors changed to (numbers). 
console.log(isEmpty(numbers));

Omar-Suliman:
ok thanks lsly

PR #33, opened 2018-03-08
pass 'numbers' not 'arr'

L3gomancer:
In reference to issue #21  "I-array-properties (2)" in original repo, the array name should be 'numbers' not 'arr'


L3gomancer:
Sorry I just realised I referenced an issue to a completely different exercise in this description. The Pr is for week-2/J-array-get-set/exercise.js but issue was for I. Closing PR. I'll discuss it at the session

PR #34, opened 2018-03-12
lesson outline for week 4

rarmatei:


week-4/B-more-complex-objects/README.md
```diff
@@ -0,0 +1,5 @@
+### Goals
+
+- to continue solidifying understanding of objects
+- solve more complex exercises with objects
+- show how even more complex types can be assigned to object properties: arrays and even other objects
```
djgrant: Real world examples could include:
array of objects: search results, blog posts, social feed
object with array: person.interests

week-4/C-methods-this-keyword/README.md
```diff
@@ -0,0 +1,19 @@
+### Goals
+- make objects even more useful by introducing object methods
+- introduce the `this` keyword **without going into any details about how it works** other than the fact that its a way to access the current object's properties/methods
```
djgrant: I'm not keen on introducing `this` because:
a) can achieve the same with simpler solutions
b) introducing it in this way creates a false understanding that `this` references the object literal 
c) `this` is overused anyway. IMO only use it when you there's not other solution

Example without using `this`:

```js
var athlete = {
    name: 'Usain Bolt',
    thirstLevel: 5,
    drink: function() {
        athlete.thirstLevel = athlete.thirstLevel + 2;
    }
};
```

This solution has the advantage of always working as expected whereas use of `this` can lead to unexpected results depending on context.

rarmatei: Yep - that looks like a good alternative. I'll use that, thanks.

With this topic I mainly wanted to make objects a bit more versatile by introducing methods to make them do stuff, give them some "life" and similarities to real-world objects, hence the "athlete" example.

I agree with points b) and c). However I don't think it would be that easy for them to get unexpected results with `this`. They'd have to either have other objects "borrow" the `drink` function, or `bind/call/apply` on it - both very unlikely scenarios (please correct me if I've missed anything).

kabaros: I think it's ok to have `this` without knowing full implications. They will also be using `this` often in React and the scope issues are less relevant with es6.

rarmatei:
Thanks! Yes, I'll split up the first part.

We can do `object.keys` as well. 

Do they need `Object.assign` though? Isn't that for when they start caring about immutability/redux?

week-4/D-arrays-of-objects/README.md
```diff
@@ -0,0 +1,29 @@
...
+var kitten2 = {
+    name: 'Megatron',
+    weeksOld: 1
+};
+var kitten3 = {
+    name: 'Billy',
+    weeksOld: 5
+};
+
+var kittens = [kitten1, kitten2, kitten3];
```
kabaros: could you also have examples where you use the objects inside the array, i.e.

```js
var kittens = [{name: 'something'}, {name: 'something2'}]
```

Students seem to take things literally and really don't understand that these variables are labels that reference other stuff. That way also `json` is less of a shock for them.

week-4/B-more-complex-objects/README.md
rarmatei: cool, good examples!

week-4/A-objects-intro/README.md
```diff
@@ -0,0 +1,9 @@
+### Goals
+
+- introduce objects as a concept
+- when can we use objects (to store what type of information)
+- the very specific object syntax (the commas after each property, the colon between property name and value, the curly brackets)
+- the language to use when talking about an objects: object literal, properties, property names, property values
+- reading object properties using the *dot notation*: `console.log(myObject.name)`
+- setting object properties: `myObject.name = "John"`
+- very simple exercises to get them used to declaring object literals, reading and writing from/to them
```
kabaros: Also rules for naming property names (not having spaces, what to do when there is space ...)

week-4/B-more-complex-objects/README.md
kabaros: Also I suggest a section where we explicitly explain when to use Arrays vs. Objects ... this is an area that was never clear for the students in the last round. This article might be a good anology https://medium.freecodecamp.org/javascript-arrays-and-objects-are-just-like-books-and-newspapers-6e1cbd8a1746

week-4/D-arrays-of-objects/README.md
rarmatei: yes, good point

rarmatei:
@kabaros I'll make a note to have a dedicated discussion towards the end when to use objects vs. arrays. 

I had that debugging exercise planned as well - but this lesson seems to be getting a bit too crowded already, so I might keep it out.

djgrant:
That is a good point about React. I'm all for setting the scene for future
modules

rarmatei:
@djgrant @kabaros I've made the changes - if you are ok with the sections split (A/B/C/D/E/F) then do you mind if I merge this in so people can start adding exercises to each section/homework?

If there are any other points we missed during our last class I can add them to my notes to make sure we remember to mention/discuss on Saturday.

kabaros:
@rarmatei .. no need to merge it, people can contribute/add PRs against this branch or contribute directly to it.

If you merge, then students would get it if they do pull upstream when getting homework (then maybe conflicts if we change structure or they do)

djgrant:
I had planned these topics for week 3 but ended up not including:
- array mutation methods vs pure methods (eg slice vs splice)
- reference vs primitive values (eg var arr2 = arr1 does not create a copy)
- arrow functions (comprehensive introduction rather than passing mention
in map topic)

Up to you if you want to include them in week 4

PR #36, opened 2018-03-14
Week 4 - Part B, Objects get/set

rarmatei:


week-4/B-objects-get-set/README.md
```diff
@@ -1,4 +1,52 @@
...
+    isTouchscreen: true
+};
+```
+
+> Try to `console.log(laptop)`. The output might depend on your environment!
+
+To find out the value of an object's property, you can use the dot notation..
+
+```js
+console.log(laptop.brand);
```
djgrant: This section might be better to follow along if the examples are in the node console (rather than console.log from a file)

rarmatei: That makes sense - when I show this in front of the class I'll make sure to use Node.

I thought I would still leave them in here as `console.log` as that gives a better indication of how it's being used.

Alternatively, I could use a variable:

```js
var brand = laptop.brand;
```

djgrant: For future reference, in previous exercises I've used this format :

```sh
$ node
> 1 + 2
3
> "hello"
'hello'
> var greeting = "hello"
undefined
> greeting
'hello'
> console.log(greeting)
hello
undefined
> .exit
$
```

Example: https://github.com/CodeYourFuture/js-exercises/tree/master/week-2/F-logical-operators

PR #41, opened 2018-03-16
Part D - Object methods

llh1:


week-4/D-methods/README.md
```diff
@@ -1,19 +1,69 @@
-### Goals
-- make objects even more useful and more closer to "real-life" by introducing object methods
-- explain how to get access to the current object reference (the `athlete` inside the the `drink` function in the example below)
+### Object methods
 
-The should be able to understand the following:
+Besides having specific properties, objects in the real world can also do things. For example, a plane can fly, a computer can display something on the screen, an athlete can run etc... To translate this into the Javascript world, let's introduce object's methods.
```
tekul: I think it's probably best to avoid examples like "flying", "running" etc, since objects are more anaemic in practice - methods just access and manipulate the data in the object. It'd be rare to implement more complicated functionality which uses the data (e.g displaying it on the screen) in a method on the object itself. Also it might be an idea to link the use of "methods" to the ones we've already seen for arrays.

week-4/D-methods/README.md
```diff
@@ -1,19 +1,69 @@
...
+    sayHi: function() {
+        return "Hi everybody!";
     }
 };
+```
 
-console.log(athlete.thirstLevel);
-athlete.drink();
-console.log(athlete.thirstLevel);
+How would call this method?
```
tekul: Missing we/you - "How would we call ..."

week-4/D-methods/exercise-4.js
```diff
@@ -0,0 +1,23 @@
+/* 
+Alice has a list of good friends. Define a method "makeFriend" to add a new friend called "Bob" to her list.
```
tekul: Maybe drop the 'called "Bob"' here, 'cos that's a specific example. A few students are still confused about the difference between adding functions with abstract parameters and specifically implementing them with the test data hard-coded :).

week-4/D-methods/exercise-1.js
```diff
@@ -0,0 +1,22 @@
+/* 
+A person named Alice is defined below. 
+Add a method "greet" so this person can say hello.
+*/
+
+var person = {
+    name: "Alice",
+    age: 25
+};
```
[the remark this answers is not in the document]
rarmatei: Yes, even though there is a previous exercise where they have to add a simpler property to an existing object without touching the initialization, I do agree something like the below might confuse them:

```js
person.printName = function() {
    console.log(this.name);
};
``` 

I guess we could let them directly edit the object in this exercise?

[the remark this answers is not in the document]
tekul: Yeah, I wouldn't have known you could do that :)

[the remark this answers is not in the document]
llh1: Good call, I thought they could implement it as person.printName = ... but it's probably less confusing to let them change the object directly.

week-4/D-methods/README.md
```diff
@@ -1,19 +1,70 @@
-### Goals
-- make objects even more useful and more closer to "real-life" by introducing object methods
-- explain how to get access to the current object reference (the `athlete` inside the the `drink` function in the example below)
+### Object methods
 
-The should be able to understand the following:
+Besides having specific properties, objects in the real world can also do things. For example, a computer can display something on the screen, a person can say their names etc... To translate this into the Javascript world, let's introduce object's methods. 
+A method is a function attached to a particular object.
```
tekul: "Let's introduce object's methods" sounds a bit awkward. Perhaps these two sentences (from "To translate ...") could be shortened to something like "In Javascript we do this using 'methods'. A method is <blah blah>. Then maybe say that we have already been using predefined methods on strings (like `toUppercase()`) and on arrays.

PR #43, opened 2018-03-16
Week 4 - Part F - Object.keys

rarmatei:


rarmatei:
@ShetlandJ I opened a PR from your branch so people can add comments. Hope that's okay!

week-4/F-object-keys/README.md
```diff
@@ -1,5 +1,18 @@
...
+
+```js
+var footballClubs = {
+  chelsea_fc: 'England',
+  fc_barcelona: 'Spain',
+  ac_milan: 'Italy'
+};
+
+console.log(Object.keys(footballClubs));
+// returns [ 'chelsea_fc', 'fc_barcelona', 'ac_milan' ]
```
tekul: I wouldn't say "returns" here to avoid confusion with returning values from functions - just say 'prints "....."'. Logging is a side-effect rather than a return value.

week-4/F-object-keys/README.md
```diff
@@ -1,5 +1,18 @@
 ### Goals
```
rarmatei: I think we can replace `Goals` with `Object.keys()` as the title of the lesson

week-4/F-object-keys/exercise-part-0.js
```diff
@@ -0,0 +1,32 @@
...
+var capitalCitiesKeys = ;
+var highScoresKeys;
+
+// ONLY EDIT ABOVE HERE
+
+console.log(capitalCitiesKeys);
+// Should return [ 'edinburgh', 'nairobi', 'canberra', 'ottowa' ]
+
+console.log(highScoresKeys)
+// Should return ['34, '55', '89', '100']
```
rarmatei: In the other lessons, I thought of adding the expected outputs in the console.log itself, so it stands out more to students if they made a mistake:

```js
console.log("Expected result: edinburgh, nairobi, canberra, ottowa. Actual result: " + capitalCitiesKeys);
```

week-4/F-object-keys/README.md
```diff
@@ -1,5 +1,18 @@
 ### Goals
 
-- explain `Object.keys()` 
-- good time to mention how `Object` is just another JS object provided by the runtime that we can access anytime (similar to `console`)
-- continue with more advanced exercises that use `Object.keys` but also build up on the previous topics
\ No newline at end of file
+`Object.keys()` allows your to iterate over an object's keys, in the same way that a `for..in` loop would, and return an array of strings of your object's keys.
+
+Object is a type that has built-in methods attached to it - such as `.keys()` or `.values()`. You're all very familiar with `console.log()` - `Object` is helpfully provided to us.
```
tekul: I dunno if `Object` is really best desribed as a type. I asked @rarmatei  and he said "`Object` is just another "object" available on the global `window` object", but that's probaby not much help here :). Also I don't think they've seen `for..in` or are familiar with the word "iterate" yet.

week-4/F-object-keys/exercise-part-2.js
```diff
@@ -0,0 +1,46 @@
+/*
+
+Objects very often contain other objects. You'll encounter nested objects all the time in the wild. This exercise will give you practice in getting the keys of objects inside other objects.
```
tekul: People might not know phrases like "in the wild" so that might just be extra noise when trying to understand the requirements - the first sentence already says that they are common. Maybe just reiterate that objects within objects are called "nested" so that people get that again when they come to read the exercise?

week-4/F-object-keys/exercise-part-1.js
```diff
@@ -0,0 +1,25 @@
...
+// ONLY EDIT BELOW THIS LINE
+
+var mentorsNames = ;
+
+var mentorsNamedCapitalised = ;
+
+// ONLY EDIT ABOVE THIS LINE
+
+console.log(mentorsNamedCapitalised);
+// Should return [ 'JAMES', 'JOSH', 'JAMIE', 'MOZAFAR' ]
```
tekul: Again, I'd avoid saying "return". I'd go with Loic's approach of logging both expected and actual values.

week-4/F-object-keys/exercise-part-0.js
```diff
@@ -0,0 +1,32 @@
...
+Return the keys of the following object
+
+*/
+
+var capitalCities = {
+  edinburgh: 'Scotland',
+  nairobi: 'Kenya',
+  canberra: 'Australia',
+  ottowa: 'Canada'
+};
```
rarmatei: I think this should be reversed. When I see an object "capitalCities", I'd expect to pass it a country, and get its capital city: `console.log(capitalCities['Scotland']);`

It also seems a bit more useful that way, as you usually know the country, but need to find out the capital city, rather than the other way around.

week-4/F-object-keys/exercise-part-0.js
```diff
@@ -0,0 +1,32 @@
...
+
+var highScores = {
+  55: 'Alistair',
+  100: 'David',
+  89: 'Hannah',
+  34: ['Sergi', 'Frank',]
+};
```
rarmatei: Hmm, is there any other example we could use here?

The way I would think to store a list of high scores would be in a sorted array, rather than an object of score, mapped to names:

```js
// I'd make sure to not allow the array to grow bigger than 10, or the max number of max scores I want to keep
var highScores = [
 {name: 'Alistair' , score: 100},
 {name: 'Hanna', score: 78}
//etc.
]
```

week-4/F-object-keys/exercise-part-1.js
```diff
@@ -0,0 +1,25 @@
...
+
+// ONLY EDIT BELOW THIS LINE
+
+var mentorsNames = ;
+
+var mentorsNamedCapitalised = ;
+
+// ONLY EDIT ABOVE THIS LINE
+
+console.log(mentorsNamedCapitalised);
```
rarmatei: Here as well, could we please add the expected output?

week-4/F-object-keys/exercise-part-2.js
```diff
@@ -0,0 +1,46 @@
...
+
+Use the provided console.log statements below and follow the instructions above them.
```
tekul: I think the students might need some extra guidance here. There aren't really instructions. You could perhaps say something like "Use the keys() method on the `storeBranches` object and complete the log statements to match the expected output in each case".

week-4/F-object-keys/exercise-part-2.js
```diff
@@ -0,0 +1,47 @@
...
+
+// # 2
+// Should return [ 'manager', 'assistant', 'interns' ]
+console.log()
+
+// # 3
+// Should return [ 'head_intern', 'intern' ]
+console.log()
+
+// ONLY EDIT ABOVE THIS LINE
```
rarmatei: Instead of them logging directly, should we create 3 variables for each part and ask them to assign values to them?

Then, similar to previous exercises, we can prepare at the bottom some `console.logs` with the **expected** and **actual** results.

tekul:
@rarmatei Are we good to merge this? I think it's preferable, since the current README on master for this section has no instructions.

ShetlandJ:
yes I guess so!

James

On 26 Mar 2018, at 15:59, ___ <notifications@github.com<mailto:notifications@github.com>> wrote:


@rarmatei<https://github.com/rarmatei> Are we good to merge this? I think it's preferable, since the current README on master for this section has no instructions.

rarmatei:
Thanks, I've merged it!

PR #49, opened 2018-03-22
Week5 - Ajax

llh1:


week-5/C-webchat/exercise.js
```diff
@@ -0,0 +1,56 @@
...
+API: https://codeyourfuture.herokuapp.com/api/messages
+Request Body: { "content": "some text" }
+*/
+
+// Write your code for task 1 here
+
+
+/*
+======
+Tast 2
```
kabaros: * Task

week-5/C-webchat/exercise.js
```diff
@@ -0,0 +1,56 @@
...
+
+HTTP Verb: GET
+API: https://codeyourfuture.herokuapp.com/api/messages
+*/
+
+// Write your code for task 2 here
+
+
+/*
+======
```
kabaros: is it possible to separate the exercises in separate files, and include them as three scripts? just to make it easier to follow (not sure . if it's a better idea but just to consider)

week-5/C-webchat/exercise.js
```diff
@@ -0,0 +1,56 @@
...
+Task 3
+======
+Reload the messages every 2 seconds, so you can see new incoming new messages without 
+having to reload the page. Hint: use the javascript function setInterval() to execute some 
+code every x seconds.
```
kabaros: great you're mentioning `setInterval` it will give some context to async stuff as well ... people find it easier to conceptualise this one as async than AJAX calls for some reason.

At this point of the class, it would be good if you just introduce `setTimeout` and `setInterval` as a branch off the main talk, and show them how the code doesn't execute linearly with those. (also throw in the word "callback" few times, even though they should know it from array functional methods)

week-5/README.md
```diff
@@ -0,0 +1,45 @@
+# AJAX
+
+## Client/Server architecture
+
+A **server** is device or program that provides functionality to other programs or devices. There are database servers, mail servers, game servers, etc. The vast majority of these servers are accessed over the internet.
+
+The server communicates with **clients**. Clients can be a web browser, a Slack app, your phone, etc.
+
+Client–server systems use the **request–response** model: a client sends a request to the server, which performs some action and sends a response back to the client, typically with a result or acknowledgement.
+
```
kabaros: This is an important concept to pin early. What is client/server, what is frontend/backend etc...? The students, in the last class, seemed confused by that for way longer than they should have.

A suggestion is to even incorporate this game we do in the taster classes: https://codeyourfuture.github.io/taster-class-master/intro/internet-intro.html .. explaining how internet works, and just emphasising what's a client and a server

week-5/README.md
```diff
@@ -0,0 +1,45 @@
...
+| Request type  | Description   |
+| ------------- |:-------------:|
+| GET          | Ask for a specified resource (e.g. show me that photo) |
+| POST          | Send content to the server (e.g. post a photo) |
+
+
+## What is AJAX?
+
+AJAX is a technique for implementing client-server communication in the browser. The word is a bit of a misnomer as generally we don't use XML any more, but it's still commonly used to describe this technique.
```
kabaros: `misnomer` is a difficult word :)

week-5/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+Typically, the server holds the data, and only sends it to the client (web page) when there's a request. Ajax requests are sent after the page has loaded, usually in response to an action by the user. For example when the user clicks a button, some JavaScript will trigger an Ajax request to fetch data.
+
+## Introduction to Fetch
+
+**fetch** is a way of creating HTTP requests in JavaScript. It uses the Promise API.
```
kabaros: I don't know whether you want to talk about it being `asynchronous`  here .. maybe if there is a code sample there is console.log before the call, after it, and inside the `thens` and we can ask them what order they think the console logs will be in.

maybe here also it will be good to show setInterval or setTimeout, as another example of code that wouldn't execute "in order"

week-5/A-fetch/index.html
```diff
@@ -0,0 +1,28 @@
+<!doctype html>
+
+<html lang="en">
+<head>
+  <meta charset="utf-8">
```
kabaros: @NateWr  suggested (in a different context) that we let them use a package like [serve](https://www.npmjs.com/package/serve) to serve the html file through a local server. I am just putting this here but it's up to you - a pro for that would be that they'd be introduced to npm and the idea of a server earlier, but a big drawback right now would be that they will confuse what's a server and what's a client even more)

week-5/README.md
```diff
@@ -0,0 +1,45 @@
+# AJAX
+
+## Client/Server architecture
+
+A **server** is device or program that provides functionality to other programs or devices. There are database servers, mail servers, game servers, etc. The vast majority of these servers are accessed over the internet.
```
tekul: s/is device/is a device/

week-5/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+The server communicates with **clients**. Clients can be a web browser, a Slack app, your phone, etc.
```
tekul: s/Clients/A client/

week-5/README.md
tekul: It might be better to skip the explanation of where the name comes from - I don't think many of them will know what XML is or that the X refers to that. Maybe just add "(it's called AJAX for historical reasons)" and skip the second sentence?

week-5/README.md
tekul: Do they need to know about the "Promise API" here? Reading that I don't think I would have any idea what it meant. Perhaps that can be left out and they can read more about it if they read the fetch docs?

rarmatei: Agree with @tekul - should we just skip mentioning promises?

Async, client/server, fetch, setTimeout/setInterval are already pretty difficult new concepts that we'll have to go over.

week-5/A-fetch/index.html
rarmatei: Hmm - as we're gonna have a lot of new concepts this lesson, can I suggest we leave `npm` for now? We can introduce at the `unit testing` lessons?

And yes, maybe it would be better if we keep the server at some external URL, and avoid introducing the concept of "local server" yet?

week-5/C-webchat/exercise.js
llh1: Yeah good point, I'm going to split what I can.

PR #50, opened 2018-03-22
Week 4 (Rukmini)

neelunsiri:


week-4/E-arrays-of-objects/exercise-2.js
```diff
@@ -7,50 +7,61 @@ Each destination has a name, a distance from Glasgow, and a list of transportati
...
+  destination3,
+  destination4
+];
 
 /* 
 DO NOT EDIT ANYTHING ABOVE THIS LINE
 WRITE YOUR CODE BELOW
 */
 
+var destinationNamesWithin500Kms = ["Edinburgh", "Dublin"]; // Complete here
```
ninakahr: Instead of hard-coding the answers you should try to use some of the array methods we've learned, such as `.filter()` and `.find()`. Let me know if you need some help with this!

PR #53, opened 2018-03-25
Islam complete Week5 exercises

islam95:


week-5/A-fetch/exercise.js
```diff
@@ -17,10 +17,33 @@ Open index.html in your browser. Every time you refresh the page,
...
+            paragraph.innerText = greeting;
+            count++;
+            if(count % 2 === 0){
+                paragraph.style.color = "blue";
+            } else {
+                paragraph.style.color = "red";
+            }
+            
+			box.appendChild(paragraph); 
+		});
```
tekul: Try to keep formatting and indentation consistent. This event listener code is hard to read because of the formatting. Also, why did you decide to change the `p` to a `div` and add the `p` element each time?

week-5/A-fetch/index.html
```diff
@@ -15,14 +15,31 @@
...
+    text-align: center;
+    text-decoration: none;
+    display: inline-block;
+    font-size: 16px;
+  }
+  #btn{
+    background-color: #4CAF50; /* Green */
+  }
+  #clear{
+    background-color: #ccc; /* Green */
```
tekul: I don't think `ccc` is green :slightly_smiling_face: .

week-5/C-webchat/exercise-1.js
```diff
@@ -34,4 +34,24 @@ When you open index.html in your browser, it should display the existing message
...
+			// iterate through an array of objects and store an array of html text
+			var messageText = messages.map(m => m.content + " <span style='float:right'>" + m.datetime + "</span>");
+			// iterate through an array of html text and put it inside paragraphs on the page
+			messageText.map(m => {
+				var p = document.createElement("p");
+				p.innerHTML = m;
+				messageList.appendChild(p);
+			});
+		});
+}, 2000);
```
tekul: Using an inline function with `setInterval` means that the user has to wait for 2 seconds before the information appears. It's a good idea to extract a named function, e.g. `loadMessages`, then you can call it when the page loads and also pass it to `setInterval` to refresh every 2 seconds:

    loadMessages();
    setInterval(loadMessages, 2000);

week-5/C-webchat/exercise-2.js
```diff
@@ -27,3 +27,12 @@ on the submit button. Then check the following:
 
 
 // Write your code here
+
+var btn = document.querySelector("#submit"); // select a button
+btn.addEventListener("click", function(){
+    var input = document.querySelector("#message-input"); // select an input field
+	var requestBody = { content: input.value }; // assign input value to the request variable as content of the object
```
tekul: Again, try to keep the indentation consistent.

islam95:
Thank you, Luke, for your suggestions, they are really helpful and I will try to follow them in the future.
Regarding the indentation, I do not know why it is so in GitHub, because in my editor the indentation is fine. Maybe it's because of Prettier plugin in the VS Code? I am not sure. But it definitely indents correctly in my editor.

week-5/A-fetch/exercise.js
islam95: I do not know where you see that. I never did change the `p` to a `div`. I just created a `p` element and put inside that `div` (box). So that all paragraphs will appear in the box.

week-5/A-fetch/index.html
islam95: Yeah, forgot that to change 👍

week-5/A-fetch/exercise.js
tekul: It looks like the [HTML was changed here](https://github.com/CodeYourFuture/js-exercises/pull/53/files#diff-01e4e272c3c492e60c22a4341b688359L24).

tekul:
We can take a look at the vscode settings sometime. It looks like the formatter may be using tabs, rather than converting them to spaces.

PR #59, opened 2018-04-17
completed Week 4 exercises(Awel)

abdulkadiret:


week-4-practise/level-1/exercise-1.js
```diff
@@ -36,9 +36,42 @@ var writers = [
...
+});
+
 /*
 If you want an extra challenge, only `console.log()` the writers that are alive.
-*/
\ No newline at end of file
+*/
+
+var aliveWriters = writers
+  .filter(function(writer) {
```
shesko: The use of filter is good. However you should also think of the performance consequence of using `filter` and `forEach one after the other. If you do `filter`, the program will iterate through the whole array and filter it. Then the call of `forEach` will iterate through the whole new array and `console.log`. If all writers are alive then you would need to iterate through the `writers` array 2 times. 

To avoid unnecessary iterations you could have used only the for each and have an if statement in it:

```
writers.forEach(function(writer) {
    if (writer.alive == true) {
        console.log(....)
    }
})
```

This would cause your code to iterate through the array only once. 

Not saying what you did is bad but it's good to think about performance

week-4-practise/level-2/exercise-1.js
```diff
@@ -9,17 +9,32 @@ We made a start on this for you here:
 */
 
 var bottle = {
```
shesko: I'm not sure you have fully understood what needs to be done. You are in the right path but some things are missing.

You assign an object called to a variable called `bottle`. This function has an internal variable, called `volume` (with an initial value of 0) and three functions: `fill`, `drink`, and `empty`. The function `fill` is meant to change the value of `volume` to `100`. In your implementation of bottle, the `fill` function never changes the the value of volume. You have the same issue for the function `drink`. It's meant to decrease the value of `volume` by 10, however it currently returns this subtracted value instead of setting is to `volume`

week-4-practise/level-2/exercise-1.js
```diff
@@ -35,7 +50,7 @@ bottle.fill();
...
 bottle.drink();
 bottle.drink();
-if(!bottle.empty()){
-	console.log('bottles volume = ' + bottle.volume);
+if (!bottle.empty()) {
+  console.log("bottles volume = " + bottle.volume);
 }
-console.log("Above volume should be: 70");
\ No newline at end of file
+console.log("Above volume should be: 70");
```
shesko: To test your program add these lines of code. The comment next to each `console.log` line represent what the output on the console should be:

```
console.log(bottle.volume) // 0

bottle.fill()
console.log(bottle.volume) //100

bottle.drink()
console.log(bottle.volume) //90

bottle.drink()
console.log(bottle.volume) // 80
console.log(bottle.empty()) // false


bottle.drink()
bottle.drink()
bottle.drink()
bottle.drink()
bottle.drink()
bottle.drink()
bottle.drink()
bottle.drink()
bottle.drink()
bottle.drink()
console.log(bottle.empty()) // true
console.log(bottle.volume) // 0
```

I purposely drank the bottle more than 10 times to make sure that you are not subtracting to negative values.

If you are unsure on how to achieve this, I would recommend to revise this page `https://github.com/CodeYourFuture/js-exercises/tree/master/week-4/D-methods`.

week-4/C-more-complex-objects/exercise-2.js
```diff
@@ -6,37 +6,42 @@
 */
```
shesko: Well done 👍

week-4/C-more-complex-objects/exercise-3.js
```diff
@@ -5,25 +5,27 @@
 */
```
shesko: Well done 👍

week-4/D-methods/exercise-1.js
```diff
@@ -3,16 +3,18 @@ A person named Alice is defined below.
 Add a method "greet" so this person can say hello.
```
shesko: 👍

week-4/D-methods/exercise-4.js
```diff
@@ -3,17 +3,20 @@ Alice has a list of good friends.
 Define a method "makeFriend" to add a new friend to her list.
```
shesko: 👍

week-4/D-methods/exercise-5.js
```diff
@@ -9,35 +9,64 @@ only if the inserted amount is greater or equal than the price of the coffee!
 */
 
 var coffeeMachine = {
```
shesko: Try and run this code, you will see that the output is not correct. Here is what I get when I run it:

```
Expected result: 'Please take your cappuccino'. Actual result: Please take your cappuccino
Expected result: 'Please take your blackCoffee'. Actual result: Please take your cappuccino
Expected result: 'Please take your flatWhite'. Actual result: Please take your cappuccino
Expected result: 'Sorry you don't have enough money for a flatWhite'. Actual result: Please take your cappuccino
```

The actual result doesn't always match the expected one. Could you have another go at this exercise please?

week-4/E-arrays-of-objects/exercise-1.js
```diff
@@ -5,42 +5,75 @@
 */
 
 var person1 = {
```
shesko: Great! I see you understand `map` and `filter` :)

week-4/E-arrays-of-objects/exercise-2.js
```diff
@@ -7,50 +7,71 @@ Each destination has a name, a distance from Glasgow, and a list of transportati
 3) Print in the console all the destination names more than 300 kms far away and reachable by train.
 */
 
-
 var destination1 = {
```
shesko: Well done! 👍

week-4/E-arrays-of-objects/exercise-3.js
```diff
@@ -15,36 +15,36 @@ and returns the number of restaurants in this area.
...
 var restaurant1 = {
-    name: "Paesano",
-    totalSeats: 10,
-    numberOfCustomers: 8,
-    address: {
-        city: "Glasgow",
-        area: "center"
-    },
-    menu: ["pizza", "calzone", "salad"]
+  name: "Paesano",
```
shesko: man... I love paesano!! Best pizza in the world!

week-4/E-arrays-of-objects/exercise-3.js
```diff
@@ -54,32 +54,52 @@ DO NOT EDIT ANYTHING ABOVE THIS LINE
...
-    },
-    countNumberOfRestaurantsInArea: function(area) {
-        // Complete here
-    }
+  applicationName: "Restaurant Finder",
+  applicationVersion: "1.0",
+  restaurants: restaurants,
+  findAvailableRestaurants: function(numberOfPeople) {
+    var availableRestaurants = restaurants.filter(function(restaurant) {
+      return;
```
shesko: I think you have a problem here, try to run the code and see :)

week-4/E-arrays-of-objects/exercise-3.js
```diff
@@ -54,32 +54,52 @@ DO NOT EDIT ANYTHING ABOVE THIS LINE
...
+      restaurant.totalSeats - restaurant.numberOfCustomers >= numberOfPeople;
+    });
+
+    return availableRestaurants.map(function(restaurant) {
+      return restaurant.name;
+    });
+  },
+  findRestaurantServingDish: function(dishName) {
```
shesko: The next two functions need an implementation. Could you let me know once you completed them?

PR #64, opened 2018-04-22
Francisca

franciscavargas:


week-1-practise/level-1/8-function-output-1.js
```diff
@@ -6,7 +6,7 @@ var result = getNumber();
...
 // Answer these questions:
-// - What does this program do?
-// - How many parameters does the function take?
-// - What is the function name?
-// - Where is the function called? with what parameters?
\ No newline at end of file
+// - What does this program do? It takes a random number and multiplies it by 10 and later it's printed the result
+// - How many parameters does the function take? 1
+// - What is the function name? getNumber
+// - Where is the function called? with what parameters? It's called in line 5 after the = symbol and it's called with an undefined parameter.
```
dubchoi: You could say that, but the more common way to describe this would be that the function is called with no parameters.

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,8 +7,19 @@
...
+  return priceUSD + fee(priceUSD);
+}
+
+function convertToBRL(price) {
+  const priceBRL = price * 4.8;
+  return priceBRL + fee(priceBRL);
+}
+
+function fee(base) {
+  return base / 100 * 1;
```
dubchoi: The `* 1` portion is unnecessary

week-1-practise/level-3/1-piping.js
```diff
@@ -11,6 +11,7 @@
 */
 
 var startingValue = 5;
+const result = (startingValue * 2 + 10).toFixed(2);
```
dubchoi: we want you to use the functions provided below (add, multiply, format) to achieve the same thing.

week-1/J-functions/exercise.js
```diff
@@ -1,7 +1,9 @@
 function halve(number) {
-  // complete the function here
+  return number / 2;
 }
 
 var result = halve(12);
+console.log(result);
 
+var result = halve(24);
```
dubchoi: You have already "declared" the variable called result in line 5. If you want to re-use this, you shouldn't *declare* it again, but just *assign*. So, line 8 should read `result = halve(24);`

week-1-practise/level-2/4-currency-conversion-2.js
franciscavargas: That's true. hahah Got it :)

week-1-practise/level-3/1-piping.js
franciscavargas: let result = format(add((multiply(startingValue, 2)), 10));

PR #65, opened 2018-04-23
Answer to the exerxise for the first pull request

rajgthub:
This is my solutions for the week1 exercises. Please have look at them! thanks for your valuable time spending on reviewing my work!

### DIFF
--- a/week-1-practise/level-1/8-function-output-1.js
+++ b/week-1-practise/level-1/8-function-output-1.js
@@ -7,6 +7,10 @@ console.log(result);
 
 // Answer these questions:
 // - What does this program do?
+    /*it generates random numbers between 0 and 9; includes 0 but not 10*/ 
 // - How many parameters does the function take?
+    /*none*/ 
 // - What is the function name?
+    /*getNumber*/ 
 // - Where is the function called? with what parameters?
+    /*in line 7 and with no arguments*/ 
--- a/week-1-practise/level-2/1-tax.js
+++ b/week-1-practise/level-2/1-tax.js
@@ -5,7 +5,9 @@
   Sales tax is 20% of the price of the product
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  return .2*price;
+}
 
 /* DO NOT EDIT BELOW THIS LINE */
 /* --------------------------- */
--- a/week-1-practise/level-3/1-piping.js
+++ b/week-1-practise/level-3/1-piping.js
@@ -11,7 +11,9 @@
 */
 
 var startingValue = 5;
-
+startingValue = multiply(startingValue, 2);
+startingValue = add(startingValue,10);
+result = format(startingValue);
 /* DO NOT EDIT BELOW THIS LINE */
 /* --------------------------- */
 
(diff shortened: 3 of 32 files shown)


week-1-practise/level-1/8-function-output-1.js
```diff
@@ -7,6 +7,10 @@ console.log(result);
 
 // Answer these questions:
 // - What does this program do?
+    /*it generates random numbers between 0 and 9; includes 0 but not 10*/
```
dubchoi: Actually this will return any floating-point number between 0 and 10, not including 10. So for example, it could generate 9.99

week-1-practise/level-2/1-tax.js
```diff
@@ -5,7 +5,9 @@
   Sales tax is 20% of the price of the product
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  return .2*price;
```
dubchoi: Stylistically, it is preferred to have empty spaces around mathematical operators. Makes it more readable. `.2 * price`

week-1-practise/level-3/1-piping.js
```diff
@@ -11,7 +11,9 @@
 */
 
 var startingValue = 5;
-
+startingValue = multiply(startingValue, 2);
+startingValue = add(startingValue,10);
+result = format(startingValue);
```
dubchoi: Can you think of a way to do all this in one line? And also not modify `startingValue`

week-1/J-functions/exercise.js
```diff
@@ -1,7 +1,12 @@
 function halve(number) {
   // complete the function here
+  return number/2;
 }
 
 var result = halve(12);
 
 console.log(result);
+// calling with different value
+var result = halve(122);
```
dubchoi: You have already "declared" the variable called result in line 5. If you want to re-use this, you shouldn't *declare* it again, but just *assign*. So, line 8 should read `result = halve(24);`

week-1/L-functions-nested/exercise2.js
```diff
@@ -0,0 +1,20 @@
+var mentor1 = "Daniel";
+var mentor2 = "Irina";
+var mentor3 = "Mimi";
+var mentor4 = "Rob";
+var mentor5 = "Yohannes";
+//
+function stringToUpperCase(myName){
```
dubchoi: Think about better naming for your functions. What does this function do ultimately? When someone looks at the code and reads "stringToUpperCase", is it obvious it will do that?

week-1-practise/level-2/1-tax.js
rajgthub: Ok, thanks!

week-1-practise/level-3/1-piping.js
rajgthub: result = format( multiply ( startingValue , 2) + add( startingValue , 10 ) );

week-1/J-functions/exercise.js
rajgthub: Yes, thanks for spotting! I did not realise.

week-1/L-functions-nested/exercise2.js
rajgthub: Yes, I will.

week-1-practise/level-1/8-function-output-1.js
rajgthub: I was thinking of Math.floor((Math.random() * 10) );

PR #70, opened 2018-04-24
week-1

KarwanM:


week-1-practise/level-1/9-function-output-2.js
```diff
@@ -11,4 +11,11 @@ console.log(result);
...
-// - How can this function be improved?
\ No newline at end of file
+// - How can this function be improved?
+
+// Merge two strings w1, w2
+// Two parameters
+// Function name is s
+// Line 5 with two parameters 'code' and 'your'
+// codeyour, w1 merge w2
+//
```
dubchoi: Can you think of a way to improve this function?

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,8 +7,12 @@
   Find a way to add 1% to all currency conversions (think about the DRY principle)
 */
 
-function convertToUSD(price) {}
-function convertToBRL(price) {}
+function convertToUSD(price) {
+  return "$" + ((price * 1.4) + (price * 1.4 * 0.01));
+}
+function convertToBRL(price) {
+ return "R$" + ((price * 4.8) + (price * 4.8 * 0.01));
```
dubchoi: What if the 1% fee changes to 2% tomorrow? You have to modify it in two places. Can you think of a way to make sure both functions use the same fee always?

week-1/I-floats/exercise.js
```diff
@@ -1,2 +1,9 @@
 var numberOfStudents = 15;
 var numberOfMentors = 8;
+
+var preceStudents = numberOfStudents * 100 / 23;
```
dubchoi: Where did 23 come from? Use variables and let the computer do the addition!

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -6,12 +6,12 @@
   They have also decided that they should add a 1% fee to all foreign transactions
   Find a way to add 1% to all currency conversions (think about the DRY principle)
 */
-
+var fee = 0.01;
```
dubchoi: This is better. Even better would be to use the const keyword here for values you don't think will change.

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -6,12 +6,12 @@
...
 function convertToUSD(price) {
-  return "$" + ((price * 1.4) + (price * 1.4 * 0.01));
+  return "$" + ((price * 1.4) + (price * 1.4 * fee));
```
dubchoi: Perhaps simpler to say `price * 1.4 * (1 + fee)`.

PR #71, opened 2018-04-24
complete class exercise

Behnazz:


week-1/K-functions-parameters/exercise5.js
```diff
@@ -1,5 +1,7 @@
 // Declare your function here
-
+function createLongGreeting(name, number) {
```
martynamaron: Just a tiny comment - it's good practice to name variables and parameters what they represent - in this case "number" could be more meaningful named "age" :)

week-1/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,42 @@
+function percentage(Students, Mentors) {
+  var studentPercentage = Math.round(100 * Students / (Students + Mentors));
+  return studentPercentage;
+}
+
+function text(Students, Mentors) {
+  var findPercentage = percentage(Students, Mentors);
```
martynamaron: Usually, variable names are nouns, and function names contain verbs - I think it's nice to represent them like this, given variables represent _things_ and functions _do_ things :)

week-1/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,42 @@
...
+
+//...........Exercise2......
+
 var mentor1 = "Daniel";
 var mentor2 = "Irina";
 var mentor3 = "Mimi";
 var mentor4 = "Rob";
 var mentor5 = "Yohannes";
+
+function spell(name) {
```
martynamaron: I think the exercise is a bit unclear about what they're asking you to do - they say to create a function that returns the name all in caps, but then the whole output is in capitals. I think in your case you don't need this function - would you agree?

week-1-practise/level-1/8-function-output-1.js
```diff
@@ -1,12 +1,12 @@
...
 console.log(result);
 
 // Answer these questions:
-// - What does this program do?
-// - How many parameters does the function take?
-// - What is the function name?
-// - Where is the function called? with what parameters?
\ No newline at end of file
+// - What does this program do? choose a random number and times it by 10.
+// - How many parameters does the function take? one
```
martynamaron: are you sure the `getNumber()` function has one parameter?

week-1-practise/level-1/8-function-output-1.js
```diff
@@ -1,12 +1,12 @@
...
 var result = getNumber();
 console.log(result);
 
 // Answer these questions:
-// - What does this program do?
-// - How many parameters does the function take?
-// - What is the function name?
-// - Where is the function called? with what parameters?
\ No newline at end of file
+// - What does this program do? choose a random number and times it by 10.
```
martynamaron: What is the specific range that number in? Is it between 1 and 1000? -1000 and 1000?

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -1,14 +1,14 @@
...
-// - Where is the function called? with what parameters?
-// - What does the function return?
-// - How can this function be improved?
\ No newline at end of file
+// - What does this program do? Run it and see, or Google some of the methods used ot understand them.   The concat() function concatenates the string arguments to the calling string and returns a new string.
+// - How many parameters does the function take? two
+// - What is the function name? s
+// - Where is the function called? with what parameters? it is called by var result and it takes 'code' and 'your' as parameters.
+// - What does the function return? A new string containing the combined text of the strings provided.
+// - How can this function be improved? in .concat we can add an empty ' ' to have some space between the w1 and w2
```
martynamaron: Good idea!
How would you improve the function and parameter names? :)

week-1-practise/level-3/1-piping.js
```diff
@@ -11,6 +11,7 @@
 */
 
 var startingValue = 5;
+var result = (startingValue * 2 + 10).toFixed(2);
```
martynamaron: Please try and use the provided functions to `add`, `multiply`, and `format`

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,8 +7,16 @@
   Find a way to add 1% to all currency conversions (think about the DRY principle)
 */
 
-function convertToUSD(price) {}
-function convertToBRL(price) {}
+function convertToUSD(price) {
+  var conversion = price * 1.4;
+
+  return conversion + conversion * 0.01;
```
martynamaron: If the conversion fee changes from 0.01 to 0.05 for example, you would need to change that in two places. Can you think of a way to write your program so that `0.01` is only used once?

week-1/L-functions-nested/exercise.js
Behnazz: Yes, you are right because of the examples at the end of the instruction, I thought that greeting should also be in the capital.

week-1-practise/level-1/8-function-output-1.js
Behnazz: I think no parameter is defined in parenthesis. So, zero?

week-1-practise/level-1/9-function-output-2.js
Behnazz: The function can be: joinWords
parameters can be: word1, word2

week-1-practise/level-3/1-piping.js
Behnazz: I really don't understand this exercise. I read some stuff about piping functions but I still don't get it :(

week-1-practise/level-2/4-currency-conversion-2.js
Behnazz: var fee = 0.01;
function convertToUSD(price) {
  var conversion = price * 1.4;

  return conversion + conversion * fee;
}
function convertToBRL(price) {
  var conversion = price * 4.82;

  return conversion + conversion * fee;
}

Behnazz: I defined a variable for 0.01 as a fee

PR #73, opened 2018-04-24
Js exercises/seraphine young

Seraphineyoung:
Hi, Please review and give feedback on Js-exercise. week 1.

### DIFF
--- a/week-1-practise/level-3/1-piping.js
+++ b/week-1-practise/level-3/1-piping.js
@@ -12,6 +12,11 @@
 
 var startingValue = 5;
 
+function piping(x) {
+  return startingValue * 2 + 10;
+}
+var result = piping(startingValue);
+
 /* DO NOT EDIT BELOW THIS LINE */
 /* --------------------------- */
 
--- a/week-1-practise/level-3/1-piping.js
+++ b/week-1-practise/level-3/1-piping.js
@@ -18,6 +18,7 @@ function piping(x) {
 var result = piping(startingValue);
 
 /* DO NOT EDIT BELOW THIS LINE */
+var result = format(add(multiply(2, startingValue), 10));
 /* --------------------------- */
 
 function add(a, b) {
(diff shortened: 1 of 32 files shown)


week-1/K-functions-parameters/exercise5.js
```diff
@@ -1,3 +1,6 @@
+function createLongGreeting(string, num) {
```
martynamaron: Just a tiny comment, that it's more readable if you give your parameters meaningful names, in this case `name` and `age` would be more readable :)

week-1/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,57 @@
+// Exercise 1//
```
martynamaron: Meaningful names for parameters and variables - I know it's just an exercise, but it's god to remember! :)

week-1/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,57 @@
...
+  return a.toUpperCase();
+}
+var a = shoutoutNames(mentor1);
+var b = shoutoutNames(mentor2);
+var c = shoutoutNames(mentor3);
+var d = shoutoutNames(mentor4);
+var e = shoutoutNames(mentor5);
+
+function shoutyGreeting(a, b, c, d, e) {
+  return (
```
martynamaron: Is this how Visual studio formatted the code? It'll be more readable if you put the `HELLO` and the `name` on one line each - if that's what you want the final out put to look like. The more you can understand from just glancing at a function the better! :)

week-1-practise/level-3/1-piping.js
```diff
@@ -12,6 +12,11 @@
 
 var startingValue = 5;
 
+function piping(x) {
+  return startingValue * 2 + 10;
```
martynamaron: How can you modify this, so instead of `*` and `+` you use the functions provided below? Also, remember about the `format` method :)

PR #75, opened 2018-04-24
Steve/practise

MHDGhanemBalhawan:
Hi Martyna,
Please find my practise for your review.
Thank you!

### DIFF
--- a/week-1-practise/level-1/10-function-output-3.js
+++ b/week-1-practise/level-1/10-function-output-3.js
@@ -1,6 +1,7 @@
 function concatenate(firstWord, secondWord, thirdWord) {
-    // Write the body of this function to concatenate three words together
+  // Write the body of this function to concatenate three words together
+  return firstWord.concat(secondWord).concat(thirdWord);
 }
 
-var result = concatenate('code', 'your', 'future');
+var result = concatenate("code", "your", "future");
 console.log(result); // expected result = "codeyourfuture"
--- a/week-1-practise/level-1/8-function-output-1.js
+++ b/week-1-practise/level-1/8-function-output-1.js
@@ -1,5 +1,5 @@
 function getNumber() {
-    return Math.random() * 10;
+  return Math.random() * 10;
 }
 
 var result = getNumber();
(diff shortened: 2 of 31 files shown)


week-1-practise/level-1/10-function-output-3.js
```diff
@@ -1,6 +1,7 @@
 function concatenate(firstWord, secondWord, thirdWord) {
-    // Write the body of this function to concatenate three words together
+  // Write the body of this function to concatenate three words together
+  return firstWord.concat(secondWord).concat(thirdWord);
```
martynamaron: You can get rid of the last `.concat()` method and have the same output. Can you think of a way how? :)

week-1-practise/level-1/8-function-output-1.js
```diff
@@ -1,12 +1,16 @@
...
-    return Math.random() * 10;
+  return Math.random() * 10;
 }
 
 var result = getNumber();
 console.log(result);
 
 // Answer these questions:
 // - What does this program do?
+// This program returns a random number
```
martynamaron: What is the range that the returned number is in? Between -1000 and 1000? Or 0 and 1000?

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -1,14 +1,18 @@
...
 // - How many parameters does the function take?
+// The function takes two parameters.
 // - What is the function name?
+// The function name is "s".
 // - Where is the function called? with what parameters?
+// The function was called in line 5 with the parameters 'code' and 'your'.
 // - What does the function return?
-// - How can this function be improved?
\ No newline at end of file
+// - How can this function be improved?
```
martynamaron: How? :) What about the function and parameter names?

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,8 +7,16 @@
   Find a way to add 1% to all currency conversions (think about the DRY principle)
 */
 
-function convertToUSD(price) {}
-function convertToBRL(price) {}
+function convertToUSD(price) {
+  var usdPrice = "$" + price * 1.4 * 1.1;
```
martynamaron: That's a cool idea :)
But when you add 1% to something, it becomes 101% not 110%. How can you fix your function?

week-1-practise/level-3/1-piping.js
```diff
@@ -11,7 +11,7 @@
 */
 
 var startingValue = 5;
-
+var result = (startingValue * 2 + 10).toFixed(2);
```
martynamaron: Please try and use the provided methods, instead of `*`, `+` and toFixed()`

PR #78, opened 2018-04-25
week-1 exercises

Bisratta:


week-1/F-strings-concatenation/exercise.js
```diff
@@ -1,3 +1,9 @@
 // Start by creating a variable `message`
 
-console.log(message);
+//console.log(message);
+var greetingStart = "Hello, my name is ";
+var name = "bisrat";
+
+var greeting = greetingStart + name;
+
+console.log(greeting);
```
foaly-nr1: Can you think of a way to do this without the variable `greeting`?

week-1/G-strings-methods/exercise2.js
```diff
@@ -1,3 +1,7 @@
-var name = " Daniel  ";
+var name = " Daniel    ";
+var nameLowerCase = name.trim().toLowerCase();
+var nameLength = nameLowerCase.length;
+
+var message = "my name is "+ name.trim() + " and my name is "+ nameLength +" characters long";
```
foaly-nr1: Good find – this is a great use case for`trim()`.

week-1/G-strings-methods/exercise2.js
```diff
@@ -1,3 +1,7 @@
-var name = " Daniel  ";
+var name = " Daniel    ";
+var nameLowerCase = name.trim().toLowerCase();
```
foaly-nr1: If you declare and assign a variable `nameLowerCase` don't forget to use it :)

week-1/J-functions/exercise2.js
```diff
@@ -1,5 +1,5 @@
 function triple(number) {
-  // complete function here
+  return number  *2
 }
 
 var result = triple(12);
```
foaly-nr1: Your method is called `triple()` – did you get the expected result?

week-1/K-functions-parameters/exercise2.js
```diff
@@ -1,5 +1,6 @@
-// Declare your function first
-
+function divide(num1,num2) {
+  return num1 / num2;
+}
 var result = divide(3, 4);
```
foaly-nr1: Do you want to try the same with some other parameters?

week-1/K-functions-parameters/exercise5.js
```diff
@@ -1,5 +1,6 @@
-// Declare your function here
-
-const greeting = createLongGreeting("Daniel", 30);
-
+function getnameage(name,age) {
+  return ("Hello,my name is"  +  name  + "and I'm" +  age  + "years old");
+}
+const greeting = getnameage("Daniel",30) ;
```
foaly-nr1: Can you think of a way to avoid using the extra variable?

week-1/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,15 @@
...
+  return (student+ mentors);
+}
+var numberOfStudents = 15;
+var numberOfMentors = 8;
+var sum = numberOfStudents +numberOfMentors;
+var percentagestudents;
+var percentagementors;
+
+percentagestudents = numberOfStudents/sum*100;
+percentagementors = numberOfMentors/sum*100;
```
foaly-nr1: You're calculating percentages in two places. Can you solve this by using a function instead?

PR #79, opened 2018-04-26
Hasanein Amori/ All exercises in week 1 folder have been solved based…

hasaninqh:
… on the given instruction

### DIFF
--- a/week-1-practise/level-1/1-syntax-error-1.js
+++ b/week-1-practise/level-1/1-syntax-error-1.js
@@ -1,6 +1,6 @@
 // There is a syntax error in this code. Fix it.
 
-function addNumbers(a b c) {
+function addNumbers(a, b, c) {
     return a + b + c;
 }
 
@@ -10,5 +10,10 @@ console.log(result); // Expected result 8 - Run the program and make sure it wor
 
 // Answer these questions:
 // 1. How many parameters does the function take?
+//taking three parameters
+
 // 2. What is the function name?
+//addNumbers()
+
 // 3. Where is the function called? with what parameters?
+// The function assigned to result variable and it is been called with three numbers which represent the three parameters
--- a/week-1-practise/level-1/10-function-output-3.js
+++ b/week-1-practise/level-1/10-function-output-3.js
@@ -1,5 +1,8 @@
 function concatenate(firstWord, secondWord, thirdWord) {
     // Write the body of this function to concatenate three words together
+    let firstConcat = firstWord.concat(secondWord);
+    let finalConcat = firstConcat.concat(thirdWord);
+    return finalConcat;
 }
 
 var result = concatenate('code', 'your', 'future');
--- a/week-1-practise/level-1/4-logic-error-1.js
+++ b/week-1-practise/level-1/4-logic-error-1.js
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word;
 }
 
 var result = trimWord("  CodeYourFuture ");
--- a/week-1-practise/level-1/5-logic-error-2.js
+++ b/week-1-practise/level-1/5-logic-error-2.js
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return word.trim;
+  return word;
 }
 
 var result = trimWord("  CodeYourFuture ");
--- a/week-1-practise/level-1/9-function-output-2.js
+++ b/week-1-practise/level-1/9-function-output-2.js
@@ -7,8 +7,19 @@ console.log(result);
 
 // Answer these questions:
 // - What does this program do? Run it and see, or Google some of the methods used ot understand them.
+// the function concatenates two strings
+
 // - How many parameters does the function take?
+//   it takes two parameters
+
 // - What is the function name?
+// it is called s
+
 // - Where is the function called? with what parameters?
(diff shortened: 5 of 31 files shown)


week-1-practise/level-1/1-syntax-error-1.js
```diff
@@ -10,5 +10,10 @@ console.log(result); // Expected result 8 - Run the program and make sure it wor
...
 // 1. How many parameters does the function take?
+//taking three parameters
+
 // 2. What is the function name?
-// 3. Where is the function called? with what parameters?
\ No newline at end of file
+//addNumbers()
+
+// 3. Where is the function called? with what parameters?
+// The function assigned to result variable and it is been called with three numbers which represent the three parameters
```
dubchoi: But where is it called? :-) Usually answer with the line number

week-1-practise/level-1/10-function-output-3.js
```diff
@@ -1,5 +1,8 @@
 function concatenate(firstWord, secondWord, thirdWord) {
     // Write the body of this function to concatenate three words together
+    let firstConcat = firstWord.concat(secondWord);
+    let finalConcat = firstConcat.concat(thirdWord);
+    return finalConcat;
```
dubchoi: This works well. In this case though, maybe we can do this without having two different variables declared. Can you think of a way?

week-1-practise/level-1/4-logic-error-1.js
```diff
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word;
```
dubchoi: Hmm not quite. We want to use the trim() function that is available for strings.

week-1-practise/level-1/5-logic-error-2.js
```diff
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return word.trim;
+  return word;
```
dubchoi: Same here

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -7,8 +7,19 @@ console.log(result);
...
 // - Where is the function called? with what parameters?
+// it is assigned to a variable called result and called with two strings ("code" and "your")
+
 // - What does the function return?
-// - How can this function be improved?
\ No newline at end of file
+//it returns string
+
+// - How can this function be improved?
+// we could use trime function and also we could print out the final stirng with upper case letters using toUppercase()
```
dubchoi: There's a more obvious problem with this function as defined. Can you think of one? How do we know what the function does from its name `s`?

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,8 +7,14 @@
...
 */
 
-function convertToUSD(price) {}
-function convertToBRL(price) {}
+function convertToUSD(price) {
+  let currency = (price * 1.39)+ 1/100;
+  return Math.round(currency);
+}
+function convertToBRL(price) {
+  let currency = (price * 4.86) + 1/100;
```
dubchoi: Can you think of a way to unify the logic for adding 1% fee? So that next time the fee changes (say, to 2%), we only need to modify it in one place?

week-1-practise/level-3/1-piping.js
```diff
@@ -26,6 +25,14 @@ function multiply(a, b) {
...
 }
+var startingValue = 5;
+
+
+
+
+let result = format(add(multiply(2,5),10));
+               
 
+ multiply(5,2) | add(5,10) | format;
```
dubchoi: What's this? :-)

week-1-practise/level-3/1-piping.js
```diff
@@ -26,6 +25,14 @@ function multiply(a, b) {
 function format(num) {
   return "£" + num.toFixed(2);
 }
+var startingValue = 5;
+
+
+
+
+let result = format(add(multiply(2,5),10));
```
dubchoi: Hmm, almost there, but let's use startingValue variable!

week-1/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,33 @@
...
+function greetingMentors(name){
+    var  greeting = "hello";
+    message = greeting.toUpperCase()+" "+name.toUpperCase();
+    return message;
+}
+console.log(greetingMentors(mentor1));
+console.log(greetingMentors(mentor2));
+console.log(greetingMentors(mentor3));
+console.log(greetingMentors(mentor4));
+console.log(greetingMentors(mentor5));
```
dubchoi: Nice. Since console.log() is also repeated across every line, maybe it can also go into the function

week-1-practise/level-3/1-piping.js
hasaninqh: I was trying to use the piping sign ( |>) by it looks  this sign is not defined in vcode. Consequently, I used the brackets ways instead

week-1-practise/level-3/1-piping.js
hasaninqh: You mean instead of using 5 , assign 5 to a variable such as startingValue and then use that variable within the formula as fellow

let result = format(add(multiply(2,startingValue),10));

is that what you mean?

week-1/L-functions-nested/exercise.js
hasaninqh: Do I need to create another function to call console.log??

PR #83, opened 2018-05-02
week-2 and week-2-practises

KarwanM:


week-2-practise/level-1/5-indexOf.js
```diff
@@ -3,7 +3,7 @@
 
 var itemToFind = 7;
 var arr = [3, 5, 61, 7, 123];
-var index; // ONLY EDIT THIS LINE
+var index = arr.indexOf(7); // ONLY EDIT THIS LINE
```
dubchoi: Always use the variable that already holds the value you are checking! So in this case, using `itemToFind` helps your code be more readable

week-2-practise/level-1/4-slice.js
```diff
@@ -5,7 +5,7 @@
   DO NOT EDIT BELOW THIS LINE
   --------------------------- */
 var numbers = [1, 2, 3, 4, 5, , 6, 7, 8];
-var first5Numbers = first5(numbers);
+var first5Numbers = numbers.slice(0,5);
```
dubchoi: You could do it this way, but the instructions specifically wanted you to create a function, and it's more readable that way too :-)

week-2-practise/level-2/1-boolean.js
```diff
@@ -2,14 +2,22 @@
 // Why doesn't it work?
 // Fix it.
 
-var isHappy = "false";
+var isHappy = true;
+//var isHappy = false;
 
-if (isHappy) {
+
+if (!isHappy) {
```
dubchoi: Hmm really? :-)

week-2-practise/level-2/2-validation.js
```diff
@@ -4,7 +4,17 @@
...
+function validate(num) {
+  if(typeof(num) == 'number'){
+    return positive(num);
+  }else{
+    return false;
+  }
+}
+
+function positive(num){
+  return 101 > num && num > 0;
```
dubchoi: Think about the name of this method. Is your code doing what is expected? If the function is called "positive", I'd expect it to only check if the number is positive.

dubchoi: Also, what happens when num is 100.5?

week-2-practise/level-2/4-slice.js
```diff
@@ -4,8 +4,9 @@
 // - return a new array with the item removed
 // - remove the item at the specified index
 
-function remove(arr, index) {
-  return; // complete this statement
+function remove(arr, index) { 
+  arr.splice(index,index);
```
dubchoi: Not quite correct here. For example, if you change numbers to [1,2,3,4], this does something unexpected. Can you see why? Read the documentation for the splice method carefully

week-2-practise/level-2/5-indexOf.js
```diff
@@ -5,13 +5,14 @@
 // TIP: Use the .indexOf() method
 
 function remove(arr, valueToRemove) {
-  return; // complete this statement
+  arr.splice(arr.indexOf(valueToRemove), arr.indexOf(valueToRemove));
```
dubchoi: Same issue as above. What are the parameters to the splice function according to documentation?

week-2-practise/level-3/2-validation.js
```diff
@@ -6,7 +6,31 @@
...
-function validate(num) {}
+  function validate(num) {
+    if(typeof(num) == 'number'){
+      return isFloat(num);
+    } else{
+      return false;
+    }
+  }
+
+  function isFloat(number){
```
dubchoi: Same comment as above. Look at how the method names and the logic of that method don't match here. Can you think of a way to fix that?

week-2-practise/level-3/5-indexOf.js
```diff
@@ -6,7 +6,7 @@
 // - return a new array with the replacement value inserted
 
 function replace(arr, valueToReplace, newValue) {
-  return; // complete this statement
+  return arr.slice(0,arr.indexOf(valueToReplace)).concat(newValue, arr.slice(arr.indexOf(valueToReplace) + 1));// complete this statement
```
dubchoi: What happens when valueToReplace is not inside the array to begin with?

week-2-practise/level-3/6-map.js
```diff
@@ -4,18 +4,19 @@
 //    - every resulting item run through the `captialise` function
 
 function tidyUpString(str) {
-  return str
+    var mentorsTidy = str
     .trim()
     .toLowerCase()
     .replace("/", "");
+    return captialise(mentorsTidy);
```
dubchoi: What if we don't want to modify tidyUpString function, as specified in the instructions?

week-2/C-comparison-operators/exercise.js
```diff
@@ -7,23 +7,20 @@
...
+var moreStudentsThanMentors = mentorCount < studentCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = roomMaxCapacity <= (mentorCount + studentCount); // finish this statement
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = personA == personB; // finish this statement
```
dubchoi: Always use ===

week-2/D-predicates/exercise.js
```diff
@@ -7,12 +7,12 @@
...
 // Finish the predicate function to test if the passed number is negative (less than zero)
 function isNegative(number) {
-
+  return number < 0;
 }
 
 // Finish the predicate function to test if the passed number is between 0 and 10
 function isBetweenZeroAnd10(number) {
-
+  return 10 > number > 0;
```
dubchoi: This won't work. One difference between javascript and common math is that in javascript, you can only compare two things at a time. So, if the number was, say, -1, the computer is actually doing:
`(10 > -1) > 0`, which then gets evaluated to `true > 0`. Since boolean value true is being compared as if it's a number, it gets converted into `1 > 0` which results in true. This is not what you expected.

How can you phrase this such that you are doing comparisons between two numbers at a time?

week-2/E-conditionals/exercise.js
```diff
@@ -11,10 +11,11 @@ var role = "mentor";
 var name2 = "Harun";
 var role2 = "student";
 
-
-console.log("Hi, My name is " + name);
-console.log("Hi, My name is " + name2);
-
+if(role2 == "student"){
```
dubchoi: Always use ===

week-2/E-conditionals/exercise.js
```diff
@@ -11,10 +11,11 @@ var role = "mentor";
...
+  console.log("Hi, My name is " + name2);
+}else{
+  console.log("Hi, My name is " + name);
```
dubchoi: Not quite doing what the instructions requested

week-2/G-conditionals-2/exercise-3.js
```diff
@@ -9,7 +9,15 @@
 */
 
 function calculateGrade(mark) {
-
+  if(mark >= 80){
+    return 'A';
+  }else if(60 <= mark && mark < 80){
```
dubchoi: You don't actually need `mark < 80` here because we have already evaluated that in line 12, and got to the "else" clause of it.

week-2/G-conditionals-2/exercise-3.js
```diff
@@ -9,7 +9,15 @@
...
+    return 'B';
+  }else if(50 <= mark && mark < 60){
```
dubchoi: Same here

week-2/L-array-methods-2/exercise3.js
```diff
@@ -7,15 +7,20 @@
 var ukNations = ["Scotland", "Wales", "England", "Northern Ireland"];
 
 function isInUK(country) {
-  return; // complete this statement
+  for(var i = 0; i < ukNations.length; i++){
+    if(country == ukNations[i]){
```
dubchoi: There's a function already that you can use on arrays to find an element. Probably better to use that instead of writing your own code!

PR #84, opened 2018-05-02
Week 2

hasaninqh:
All exercised for week2 folder have been solved .

### DIFF
--- a/week-2/C-comparison-operators/exercise.js
+++ b/week-2/C-comparison-operators/exercise.js
@@ -7,14 +7,15 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var people = studentCount + mentorCount;
+var enoughSpaceInRoom = roomMaxCapacity == people; // finish this statement
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = personA == personB ; // finish this statement
 
 /* 
   DO NOT EDIT BELOW THIS LINE
--- a/week-2/D-predicates/exercise.js
+++ b/week-2/D-predicates/exercise.js
@@ -7,12 +7,12 @@
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
 function isNegative(number) {
-
+ return number < 0 ;
 }
 
 // Finish the predicate function to test if the passed number is between 0 and 10
 function isBetweenZeroAnd10(number) {
-
+ return 0 < number < 10;
 }
 
 /* 
--- a/week-2/E-conditionals/exercise.js
+++ b/week-2/E-conditionals/exercise.js
@@ -10,8 +10,12 @@ var role = "mentor";
 
 var name2 = "Harun";
 var role2 = "student";
-
-
+ if (role == "student"){
+   return name;
+ } 
+if (role2 == "student"){
+  return name2;
+}
 console.log("Hi, My name is " + name);
 console.log("Hi, My name is " + name2);
 
--- a/week-2/F-logical-operators/exercise2.js
+++ b/week-2/F-logical-operators/exercise2.js
@@ -5,17 +5,32 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(num) {
+ var res = num < 0;
+ return res; 
+}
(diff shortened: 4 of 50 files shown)


week-2/D-predicates/exercise.js
```diff
@@ -7,12 +7,12 @@
...
 // Finish the predicate function to test if the passed number is negative (less than zero)
 function isNegative(number) {
-
+ return number < 0 ;
 }
 
 // Finish the predicate function to test if the passed number is between 0 and 10
 function isBetweenZeroAnd10(number) {
-
+ return 0 < number < 10;
```
dubchoi: This won't work. One difference between javascript and common math is that in javascript, you can only compare two things at a time. So, if the number was, say, 11, the computer is actually doing:
`(0 < 11) < 10`, which then gets evaluated to `true < 10` Since boolean value true is being compared as if it's a number, it gets converted into `1 < 10` which results in true. This is not what you expected.

How can you phrase this such that you are doing comparisons between two numbers at a time?

week-2/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,15 @@
...
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var people = studentCount + mentorCount;
+var enoughSpaceInRoom = roomMaxCapacity == people; // finish this statement
 
 var personA = "Daniel";
 var personB = "Irina";
-var sameName; // finish this statement
+var sameName = personA == personB ; // finish this statement
```
dubchoi: Always use === when comparing for equality

week-2/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,15 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var people = studentCount + mentorCount;
```
dubchoi: Is `people` the best name?

week-2/E-conditionals/exercise.js
```diff
@@ -10,8 +10,12 @@ var role = "mentor";
 
 var name2 = "Harun";
 var role2 = "student";
-
-
+ if (role == "student"){
+   return name;
```
dubchoi: Hmm...you are not inside a function. What does this return do? Does your code run and do what is expected?

week-2/F-logical-operators/exercise2.js
```diff
@@ -5,17 +5,32 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(num) {
+ var res = num < 0;
+ return res; 
+}
+function isBetween5and10(num){
+var res = 5 < num < 10;
```
dubchoi: Same comment as above

week-2/F-logical-operators/exercise2.js
```diff
@@ -5,17 +5,32 @@
...
+return res
+}
+function isLongName(name){
+ var res = name.length  < 5;
```
dubchoi: Interesting. So if the name is longer than 5 characters, isLongName returns false? That seems counter-intuitive :-)

week-2/G-conditionals-2/exercise-3.js
```diff
@@ -9,20 +9,32 @@
 */
 
 function calculateGrade(mark) {
-
+ if (mark > 80){
```
dubchoi: Not quite correct.
What if mark is 80? What does your function output?

week-2/G-conditionals-2/exercise-3.js
```diff
@@ -9,20 +9,32 @@
...
+   return "A";
+  }
+  else if (mark < 80 && mark > 60){
```
dubchoi: Once you hit the else statement, the pairing if (e.g. from line 12) has already been evaluated as false. So you don't actually need to repeat the inverse condition here like `mark < 80`

week-2/G-conditionals-2/exercise-4.js
```diff
@@ -9,7 +9,13 @@
 */
 
 function containsCode(sentence) {
-
+ if (sentence.indexOf("code") >= 0){
+   res = true;
```
dubchoi: res is a variable without declaration. That's not good :-)
Best way though is to do this all without any variables. Remember conditional expressions will result in a boolean value already. So in this case you can directly return:
`return sentence.indexOf("code") >= 0;`

PR #87, opened 2018-05-03
week-2 answers

Tebesi:


week-2/D-predicates/exercise.js
```diff
@@ -7,12 +7,20 @@
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
 function isNegative(number) {
-
+if(number < 0) {
```
dubchoi: This is correct but could be a bit more concise. Remember that conditional expressions like `number < 0` always results in a boolean value already. That means you could just return it also, e.g. `return number < 0;`

week-2/E-conditionals/exercise.js
```diff
@@ -10,10 +10,15 @@ var role = "mentor";
 
 var name2 = "Harun";
 var role2 = "student";
+if(role2) {
```
dubchoi: Hmm, what does `if (role2)` do? Does this log as intended?

week-2/F-logical-operators/exercise.js
```diff
@@ -12,13 +12,40 @@ var cssLevel = 4;
 // Finish the statement to check whether HTML, CSS knowledge are above 5
 // (hint: use the comparison operator from before)
 var htmlLevelAbove5;
+   if(htmlLevel > 5) {
+    htmlLevelAbove5 =  true;
```
dubchoi: Can you apply my above comment about being concise here as well?

week-2/F-logical-operators/exercise.js
```diff
@@ -12,13 +12,40 @@ var cssLevel = 4;
...
+        cssLevelAbove5 = false;
+     }
+   
 
 // Finish the next two statement
 // Use the previous variables and logical operators
 // Do not "hardcode" the answers
 var cssAndHtmlAbove5;
-var cssOrHtmlAbove5;
+  if( htmlLevel > 5 && cssLevel > 5){
```
dubchoi: Can you use variables you have from above so you don't repeat the same checks?

week-2/G-conditionals-2/exercise-1.js
```diff
@@ -7,9 +7,14 @@
 */
 
 function negativeOrPositive(number) {
-
+if(number< 0) {
+  return "negative";
+}else if(number >= 0) {
```
dubchoi: Once you are in the "else" block, that means the if statement in line 10 already evaluated to false. So you don't need to check again here. Just say `} else {`

week-2/G-conditionals-2/exercise-3.js
```diff
@@ -9,7 +9,17 @@
 */
 
 function calculateGrade(mark) {
-
+    if(mark >= 80) {
+       return "A";
+    }else if(mark > 60 && mark < 80) {
```
dubchoi: See my previous comment. Can we simplify this by removing redundant conditions?

week-2/G-conditionals-2/exercise-4.js
```diff
@@ -9,7 +9,11 @@
 */
 
 function containsCode(sentence) {
-
+if(sentence =sentence("code")) {
```
dubchoi: What is `sentence("code")` I don't think this would work

week-2/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return arr[0]; // complete this statement
 }
 
 function last(arr) {
-  return; // complete this statement
+  return arr.pop();// complete this statement
```
dubchoi: This does return the last element, but read the documentation closely! It will actually end up modifying arr by removing the last element from it. That's not what I would expect to happen from calling `last(arr)`. How can we return the last element without modifying arr?

week-2/J-array-get-set/exercises2.js
```diff
@@ -7,7 +7,9 @@
 */
 
 var numbers = [2, 2, 3]; // Don't change this array literal declaration
-
+     numbers.push(4);
+     numbers.shift(2);
```
dubchoi: This is interesting. How did you learn about shift function? It's good to experiment, but always pay attention to the documentation. shift does not actually take any parameters.

Is there a simpler way to just access and modify the value of an array at a specific index? (in this case, at index 0).

week-2/L-array-methods-2/exercise.js
```diff
@@ -15,8 +15,8 @@ var everyone = [
   "Swathi"
 ];
 
-var firstFive; // complete this statement
-var lastFive; // complete this statement
+var firstFive = everyone.slice(0,5); // complete this statement
+var lastFive =everyone.slice(2,7); // complete this statement
```
dubchoi: Hmm, this assumes the array has 7 elements. Can we make it work even if the array has 14, 20, or any number of elements?

week-2/L-array-methods-2/exercise2.js
```diff
@@ -7,7 +7,9 @@
   Tip: use the string method .split() and the array method .join()
 */
 
-function capitalise(str) {}
+function capitalise(str) {
+  return str.split
```
dubchoi: This doesn't seem like it would work :-) What are you thinking about here? How can we help?

week-2/L-array-methods-2/exercise3.js
```diff
@@ -6,9 +6,7 @@
 
 var ukNations = ["Scotland", "Wales", "England", "Northern Ireland"];
 
-function isInUK(country) {
-  return; // complete this statement
-}
+function isInUK(country) {}// complete this statement
```
dubchoi: Same here

PR #89, opened 2018-05-05
Week 2 practise

hasaninqh:
week-2-practise solutions

### DIFF
--- a/week-2-practise/level-1/2-validation.js
+++ b/week-2-practise/level-1/2-validation.js
@@ -2,7 +2,7 @@
 // TIP: You should write an expression that returns a boolean value
 
 var num = 10;
-var isBigEnough; // ONLY EDIT THIS LINE
+var isBigEnough = true; // ONLY EDIT THIS LINE
 
 if (isBigEnough) {
   console.log("num is bigger than or equal to 10");
--- a/week-2-practise/level-1/5-indexOf.js
+++ b/week-2-practise/level-1/5-indexOf.js
@@ -3,7 +3,7 @@
 
 var itemToFind = 7;
 var arr = [3, 5, 61, 7, 123];
-var index; // ONLY EDIT THIS LINE
+var index = arr.indexOf(7); // ONLY EDIT THIS LINE
 
 console.log(index);
 
--- a/week-2-practise/level-2/2-validation.js
+++ b/week-2-practise/level-2/2-validation.js
@@ -4,8 +4,48 @@
 // - is less than or equal to 100
 // Tip: write other small functions for each requirement
 
-function validate(num) {}
-
+function validate(num) {
+  var res;
+  if (isNum(num) && isPositive(num) && isLessOrEqual100(num)){
+    return res = true;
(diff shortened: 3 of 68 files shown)


week-2-practise/level-1/2-validation.js
```diff
@@ -2,7 +2,7 @@
 // TIP: You should write an expression that returns a boolean value
 
 var num = 10;
-var isBigEnough; // ONLY EDIT THIS LINE
+var isBigEnough = true; // ONLY EDIT THIS LINE
```
dubchoi: We wanted a boolean condition here, so that the log line will execute only if num is bigger than or equal to 10.

week-2-practise/level-1/5-indexOf.js
```diff
@@ -3,7 +3,7 @@
 
 var itemToFind = 7;
 var arr = [3, 5, 61, 7, 123];
-var index; // ONLY EDIT THIS LINE
+var index = arr.indexOf(7); // ONLY EDIT THIS LINE
```
dubchoi: You already have a variable called `itemToFind`. Use it instead!

week-2-practise/level-2/2-validation.js
```diff
@@ -4,8 +4,48 @@
...
+    else{
+      res = false;
+    }
+  
+  
+  return res;
+}
+function isLessOrEqual100(num){
+  var res;
+  if (num <= 100){
```
dubchoi: Logic here is perfect. One way to make things more concise is to remember that conditional expressions evaluate to a boolean value already. So here, you can just do `return num <= 100` and it will be equivalent to your code. This applies to the other functions here

week-2-practise/level-2/4-slice.js
```diff
@@ -5,7 +5,8 @@
 // - remove the item at the specified index
 
 function remove(arr, index) {
-  return; // complete this statement
+  arr.splice(2, 1);
```
dubchoi: Hmm, the function should work for any array and index. How can we use the parameters instead?

week-2-practise/level-2/5-indexOf.js
```diff
@@ -5,7 +5,10 @@
 // TIP: Use the .indexOf() method
 
 function remove(arr, valueToRemove) {
-  return; // complete this statement
+  var arrIndex = arr.indexOf(valueToRemove);
+  arr.splice(arrIndex , 1);
```
dubchoi: What happens when valueToRemove was not found in the array?

week-2-practise/level-3/5-indexOf.js
```diff
@@ -6,21 +6,24 @@
 // - return a new array with the replacement value inserted
 
 function replace(arr, valueToReplace, newValue) {
-  return; // complete this statement
+  var x = arr.indexOf(valueToReplace);
+   arr[x] = newValue;
```
dubchoi: What happens when valueToReplace is not found?

PR #95, opened 2018-05-13
Week 3-Behnaz

Behnazz:


week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,15 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+function Check(names) {
```
dubchoi: This is not quite right. For example, if group was `["Austine", "Dany"]`, `group.every(Check)` should return true.
What are the inputs and outputs to the Check function? Can you explain in words?

week-3/D-array-filter/exercise.js
```diff
@@ -7,8 +7,15 @@
 */
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
+var filterarr = [];
 
-var pairsByIndex; // Complete this statement
+pairsByIndexRaw.filter(function(item) {
+  if (item !== null && item !== false && item !== "whoops" && item.length > 1) {
```
dubchoi: But what if pairsByIndexRaw contains other arbitrary strings or values like, "hello" or the number 3? Remember the "typeof" keyword might come in handy here for making your check more robust.

PR #97, opened 2018-05-14
Week3 than

than2042:
Hi Johnny,

I am not quite sure some exercises as I confused some instructions. others working on it! thank you

### DIFF
--- a/week-3/B-array-some/exercise.js
+++ b/week-3/B-array-some/exercise.js
@@ -7,10 +7,20 @@
 */
 
 var pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
+var pairsByIndex = [[0, 3], [1, 2], [2, 1], [3, 0]];
 
 // If there is a null value in the array exit the program with the error code
 // https://nodejs.org/api/process.html#process_process_exit_code
 // process.exit(1);
+
+
+var nullValue = function(element) {
+
+  return element !== 0;
+};
+
+console.log(pairsByIndex.some(nullValue));
+
 
 var students = ["Islam", "Lesley", "Harun", "Rukmini"];
 var mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
@@ -22,3 +32,5 @@ var pairs = pairsByIndex.map(function(indexes) {
 });
 
 console.log(pairs);
+
+
--- a/week-3/B-array-some/exercise.js
+++ b/week-3/B-array-some/exercise.js
@@ -34,3 +34,5 @@ var pairs = pairsByIndex.map(function(indexes) {
 console.log(pairs);
 
 
+ 
+ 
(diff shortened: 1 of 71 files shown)


week-3/B-array-some/exercise.js
```diff
@@ -7,11 +7,21 @@
 */
 
 var pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
+var pairsByIndex = [[0, 3], [1, 2], [2, 1], [3, 0]];
```
dubchoi: You weren't supposed to change this input :-)  Make sure the nullValue function below actually checks for nulls!

week-3/C-array-every/exercise.js
```diff
@@ -5,9 +5,14 @@
...
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = students.every(function(element) {
+  for (i = 0; i < students.length; i++) {
+    return element;
+  }
+}) // complete this statement
 
-if (groupIsOnlyStudents) {
+
+if (groupIsOnlyStudents == students) {
```
dubchoi: Hmm this doesn't make sense. groupIsOnlyStudents is a result of the `every()` function, so will be a boolean value (true or false). That shouldn't be compared to students, which is an array!

week-3/C-array-every/exercise.js
```diff
@@ -5,9 +5,14 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = students.every(function(element) {
+  for (i = 0; i < students.length; i++) {
```
dubchoi: What is this function doing?

week-3/D-array-filter/exercise.js
```diff
@@ -6,12 +6,20 @@
   - Do not edit any of the existing code
 */
 
-var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
+ var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
+ var pairsByID = [];
+ 
+ pairsByIndexRaw.filter(function(element) {
+   if (element !== null && element !==false && element !== "whoops" && element.length > 1) {
```
dubchoi: This works, but what if the input contains other arbitrary values like "hello" or "oops". Remember the "typeof" keyword can be handy in this situation!

week-3/F-array-forEach/exercise.js
```diff
@@ -9,6 +9,25 @@
 
 var arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
 
+
+function multiplesFizzBuzz(element) {
+  for (var i = 1; i < arr.length; i++) {
+    if (arr[i] % 3 == 0 && arr[i] % 5 == 0) {
+      arr[i] = "FizzsBuzz";
+    }
+    if (arr[i] % 3 == 0) {
```
dubchoi: Unfortunately this won't work well since you don't have "else" instead of "else if" in L18 and L21.
For example, for element 15, L16, L19, and L22 will be executed, which means it will be transformed to "Buzz" instead of the correct "FizzBuzz"

PR #99, opened 2018-05-15
Steve/week3

MHDGhanemBalhawan:
Hello,
Please check my homework.
Thank you!


B-array-some/exercise.js
```diff
@@ -8,17 +8,26 @@
...
-// https://nodejs.org/api/process.html#process_process_exit_code
-// process.exit(1);
+// https://nodejs.org/api/process.html#process_process_exit_codeprocess.exit(1);
 
 var students = ["Islam", "Lesley", "Harun", "Rukmini"];
 var mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
 
+var filterarr = [];
+
+pairsByIndex.some(function(item) {
```
dubchoi: This technically works but is dangerous. Remember, functions like `some` or `every` are about checking a condition so you wouldn't necessarily expect it to change variables inside as in `forEach`.

The prompt here also wants you to merely check for nulls and then exit out of the program rather than filtering it out as you've done here.

C-array-every/exercise.js
```diff
@@ -5,7 +5,11 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = group.every(function(students) {
```
dubchoi: Not quite there. What should the function passed to every do here? What are its input (parameter) and output (return value)?

D-array-filter/exercise.js
```diff
@@ -8,7 +8,15 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-var pairsByIndex; // Complete this statement
+var filterarr = [];
+
+pairsByIndexRaw.filter(function(item) {
+  if (item !== null && item !== false && item !== "whoops" && item.length > 1) {
```
dubchoi: This works, but what if the input contains other arbitrary values like "hello" or "oops". Remember the "typeof" keyword can be handy in this situation!

F-array-forEach/exercise.js
```diff
@@ -9,6 +9,21 @@
...
+    }
+    if (arr[i] % 3 == 0) {
+      arr[i] = "Fizz";
+    }
+    if (arr[i] % 5 == 0) {
+      arr[i] = "Buzz";
+    }
+  }
+});
+console.log(arr);
```
dubchoi: Unfortunately this won't work well since you don't have "else" instead of "else if" in L17 and L20.
For example, for element 15, L15, L18, and L21 will be executed, which means it will be transformed to "Buzz" instead of the correct "FizzBuzz"

PR #100, opened 2018-05-16
Week3 by hasanein

hasaninqh:


week-3/B-array-some/exercise.js
```diff
@@ -11,7 +11,20 @@ var pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 // If there is a null value in the array exit the program with the error code
 // https://nodejs.org/api/process.html#process_process_exit_code
 // process.exit(1);
+function checkNull(pair) {
+  if (pair == null) {
+    console.log(pair);
+    //process.exit(1);
```
dubchoi: process.exit(1) is what you want to do here. why was it commented out? :-)

week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,7 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = group.every((name)=>students.includes(name)); // complete this statement
```
dubchoi: great use of includes!

week-3/D-array-filter/exercise.js
```diff
@@ -8,8 +8,13 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-var pairsByIndex; // Complete this statement
-
+var pairsByIndex = pairsByIndexRaw.filter(arrFiltering); // Complete this statement
+function arrFiltering(pair){
+  if (pair !== null && pair && pair !=="flase" && pair !== "whoops" && pair.length == 2){
```
dubchoi: This works, but what if the input contains other arbitrary values like "hello" or "oops". Remember the "typeof" keyword can be handy in this situation!

PR #107, opened 2018-05-20
Week 3

KarwanM:


week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,17 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+function isStudent(student){
```
dubchoi: Actually the parameter here is coming from the group array, so you don't know if it's a student yet. A more appropriate parameter name might just be `member`

week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,17 @@
...
+  for(var i = 0; i < student.length; i++){
```
dubchoi: Array method `includes` could be useful here rather than looping yourself

week-3/D-array-filter/exercise.js
```diff
@@ -8,7 +8,11 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-var pairsByIndex; // Complete this statement
+function isValid(validContent){
+    return validContent;
```
dubchoi: Hmm not actually doing any validation here. How can you return true only for arrays of length 2

week-3/F-array-forEach/exercise.js
```diff
@@ -3,11 +3,25 @@
   - For multiples of 3 print “Fizz” instead of the number 
   - For the multiples of 5 print “Buzz”. 
   - For numbers which are multiples of both 3 and 5 print “FizzBuzz”
-
   An array with numbers 0-15 has been provided.
 */
 
-var arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
+function combine(array){
```
dubchoi: Parameter name is misleading since this is not an "array" here

PR #119, opened 2018-06-03
Francisca - Week5

franciscavargas:


week-5-practise/js/main.js
```diff
@@ -0,0 +1,66 @@
+var btnBlue = document.querySelector("#blueBtn");
+var btnOrange = document.querySelector("#orangeBtn");
+var btnGreen = document.querySelector("#greenBtn");
+
+var bgJumbotron = document.querySelector(".primary-content .jumbotron");
+var bikeDonate = document.querySelector(".primary-content .btn-primary");
+var volunteerButton = document.querySelector(".primary-content .btn-secondary");
```
40thieves: I like the organisation 😄

week-5-practise/js/main.js
```diff
@@ -0,0 +1,66 @@
...
+});
+
+var btnSubmit = document.querySelector("form button");
+var formName = document.querySelector("#example-text-input");
+var formComment = document.querySelector("#exampleTextarea");
+var formMail = document.querySelector("#exampleInputEmail1");
+
+var allFields = document.querySelectorAll("form .form-control");
+
+btnSubmit.addEventListener("click", function() {
```
40thieves: You might like to try listening for a different event, because sometimes you don't actually need to click to get a form to submit - you can hit the Enter key for example. Try looking up the `submit` event (hint: this event is triggered by the form, not the button).

week-5-practise/js/main.js
```diff
@@ -0,0 +1,66 @@
...
+  event.preventDefault();
```
40thieves: I don't think `event` is defined here. But I actually think the default behaviour (which you're trying to prevent) is what we want. Or at least it is if you're listening for click events on the button.

week-5-practise/js/main.js
```diff
@@ -0,0 +1,66 @@
...
+  if (!formMail.value.includes("@")) {
+    formMail.style.backgroundColor = "red";
+    formError = true;
+  }
+
+  if (formError == false) {
+    alert("Thank you for filling out the form");
+    allFields.forEach(x => {
+      x.value = "";
+      x.style.backgroundColor = "white";
```
40thieves: This is a nice catch on a potential bug 🍰

PR #123, opened 2018-06-10
Add callbacks and promises exercises

40thieves:
Adds exercises for callbacks and promises.

I mostly just adapted from them old syllabus, which had them in this [CodePen](https://codepen.io/fortythieves/pen/QOYgWb?editors=1010).

@BexB1 @kkarimi @kabaros


week-5/B-promises/exercise.js
```diff
@@ -0,0 +1,156 @@
...
+
+function exercise1() {
+  // Write your implementation here
+}
+
+/*
+  EXERCISE 2
+  =======
+
+  Write a callback function that inserts the result into the #exercise2
```
djgrant: I think some documentation is needed to explain how this callback gets called.

djgrant: Wasn't clear to me until I scrolled down and looked at the function definition for functionThatCallsBack

week-5/B-promises/exercise.js
```diff
@@ -0,0 +1,156 @@
...
+*/
+functionThatCallsBack(function() {
+  // Write your implementation here
+})
+
+/*
+  EXERCISE 3
+  =======
+
+  Write a function that calls the callback with "Hello from the callback"
```
djgrant: ....with _the parameter_ "Hello from the callback"

40thieves:
Thanks for the review 😄

> I think it would be good for them to be tackled in separate files.

Yeah, I was thinking that week5 has a lot of stuff now. But as I mentioned on slack I think we might want to split up this content a bit more (which we should do separately). I've moved the callbacks exercises to their own directory.

I've also addressed your other feedback.

djgrant:
Looks good!

40thieves:
Cheers! Merging

PR #124, opened 2018-06-13
Week6/raj

rajgthub:
Please have a look at my code! Thanks for your time.


week-5/B-callbacks/exercise.js
```diff
@@ -27,10 +28,11 @@ function exercise1() {
   Finish writing the callback function to insert the result into the #exercise2
   element
 */
-functionThatCallsBack(exercise2)
+functionThatCallsBack(exercise2);
```
40thieves: It's totally fine to add a semicolon here (and some people would argue that it's good practice). However you don't actually _need_ to add them. There are special rules where Javascript will add them for you. If you want to learn more about this try googling "automatic semicolon insertion"

week-5/B-callbacks/exercise.js
```diff
@@ -46,6 +48,7 @@ function exercise2(result) {
 
 function exercise3(callback) {
   // Write your implementation here
+  functionThatCallsBack(callback);
```
40thieves: This isn't quite right. This code will set the inner text for the `#exercise3` element to "Hello from the function caller". It should say "Hello from the callback". How could you fix this? Hint: the `callback` parameter that you are provided with is a function that you can call yourself inside `exercise3()`.

I can see that the exercise was maybe confusing, so I'll try to improve the wording.

week-5/C-promises/exercise.js
```diff
@@ -45,6 +48,9 @@ function exercise3() {
 */
 function exercise4() {
   var promise4 = concatPromise()
+  promise4
+  .then(result => result + 'RAJ')
+  .then( conResult => document.querySelector('#exercise4').textContent = conResult);
```
40thieves: This is nice & clear what you're doing, I like it 👍 

There is another way of doing it (which isn't necessarily better, just letting you know), where you concatenate the string and update the DOM in 1 step:

```js
.then((result) => {
  document.querySelector('#exercise4').textContent = result + 'RAJ';
})
```

week-5/C-promises/exercise.js
```diff
@@ -57,11 +63,12 @@ function exercise4() {
   Take the result of your promise and insert the it into the #exercise5
   element
 */
-
-function exercise5() {
+const exe5Promise = async () => "Hello Promises!";
```
40thieves: 😄 I actually didn't think about making a Promise this way! Nice work 👍 Just be a little careful with the `async`/`await` keywords as they aren't quite fully supported yet.

There is another way of creating a Promise (the way I was thinking of!). Do you know what that would look like?

week-5/C-promises/exercise.js
```diff
@@ -72,8 +79,13 @@ function exercise5() {
   Take the result of your promise and insert the it into the #exercise6
   element
 */
+const exe6Promise = async () => { 
+  throw new Error ("Something went wrong!"); // this will return a promise!
```
40thieves: This works too! Do you know the other way of creating a rejected Promise?

week-5/B-callbacks/exercise.js
40thieves: Hmm, looking at some of the other PRs, I think it's possible you have something like [Prettier](https://prettier.io/) installed. In which case feel free to ignore my comment :)

rajgthub: I am using VS code and think it has the plugin installed, not sure!

40thieves: Yep I think you're right

week-5/B-callbacks/exercise.js
rajgthub: Oh! I did it in a hard way! Yes, I have not read clearly.

week-5/C-promises/exercise.js
rajgthub: I thought we were expected to familiar with promise chaining. 
 step 1: Take the result of promise4 and concatenate your name onto the end of it.
  step 2: Then return your concatenated string as a promise. (!!! )
  step 3: Finally insert the result  of your promise and insert it into the #exercise4 element

week-5/C-promises/exercise.js
rajgthub: Thanks for letting me know. In that case, we need to run our code via babel and then deploy for production build. Is that correct? I was reading what you asked us from the JS chapter 11 and tried to use async/ await.

week-5/C-promises/exercise.js
rajgthub: Yes, I do. It is similar to what you did in the previous exercises.

rajgthub:
I have done the corrections, but, unfortunately I am unable to push it. As I changed password, I am getting an error stating fatal: unable to access 'https://github.com.. I will try to setup ssh key again.Should I need to do that or any ideas?

week-5/C-promises/exercise.js
rajgthub: const exe5Promise = () => 
  new Promise((resolve, reject) => resolve("Hello Promises!"));

week-5/C-promises/exercise.js
rajgthub: const exe6Promise = () =>
  new Promise((resolve, reject) => reject("Something went wrong!"));

rajgthub:
Oh! last time I pushed directly to the upstream after changing password. That was my mistake!

week-5/C-promises/exercise.js
40thieves: This wasn't a criticism! Just pointing out that there's several ways of doing it. In general I would expect most people would use method I used because it's shorter, however I prefer your way :)

week-5/C-promises/exercise.js
40thieves: Yes, you're right - Babel will allow you to convert async/await so that you can run it in production (there is a small little issue, but it's not really important).

I was mostly just making sure that you knew the other way. And yes that is indeed correct 🎉

week-5/B-callbacks/exercise.js
```diff
@@ -42,13 +42,14 @@ function exercise2(result) {
...
-  What happens to the #exercise3 element? Why does this happen? (Hint: look at
+  What happens to the #exercise3 element? 
+  Why does this happen? (Hint: look at
   line 63 below)
 */
 
 function exercise3(callback) {
   // Write your implementation here
-  functionThatCallsBack(callback);
+  callback("Hello from the  callback");
```
40thieves: Nice work!

40thieves:
Sorry it's taken me so long to get back to you :/

I also meant to say that most of my feedback was trying to get you to think about different ways we could solve the problems, and your solutions weren't wrong. Sometimes in programming there isn't a "correct" or "wrong" answer, and you have to work out what the best solution for that problem. That's what I was attempting to do

PR #126, opened 2018-06-16
All JS exercise

Behnazz:


week-5/B-callbacks/exercise.js
```diff
@@ -14,6 +14,10 @@
 document.querySelector('#button1').addEventListener('click', exercise1)
 
 function exercise1() {
+  var name = "Behnaz";
+  var paragraph = document.querySelector("#exercise1");
+  paragraph.innerText = name
```
40thieves: Very clear 👍

week-5/C-promises/exercise.js
```diff
@@ -10,18 +10,28 @@
 
   Insert the result of promise1 into the #exercise1 element. Hint: .then()
 */
+
+var pEx1 = document.querySelector("#exercise1")
```
40thieves: This is not wrong, but personally I would have defined the `pEx1` variable inside the `.then` handler function. This is because we only really need to use the variable inside the `.then` handler function and nowhere else. If it's defined outside of this scope it's a bit unnecessary. We sometimes call this "polluting the scope".

Again, it's not wrong, just a question of style and that comes with experience.

week-5/C-promises/exercise.js
```diff
@@ -58,9 +81,23 @@ function exercise4() {
...
 */
 
+
+
 function exercise5() {
-  // Write your implementation here
-}
+  var pEx5 = document.querySelector("#exercise5");
+  var promise = new Promise(function (resolve) {
+    resolve("Hello Promise")
```
40thieves: Nice work!

There's also a shorter way of writing a resolved Promise. Do you know what that is? It is shorter & easier to write, however it is less flexible that the method you used here.

PR #127, opened 2018-06-16
Week5 than

than2042:
Hi Ali, 
I pushed a week before homework but it had gone under master. I have no idea how that happened. However, I have complete callback and promise. I have a few questions to raise tomorrow. looking forward to have a feedback. 

Thanks for giving us your time.


week-5/C-promises/exercise.js
```diff
@@ -11,8 +11,15 @@
   Insert the result of promise1 into the #exercise1 element. Hint: .then()
 */
 function exercise1() {
-  var promise1 = resolvedPromise()
-}
+  var promise1 = resolvedPromise();
+  promise1.then(function (){
+    var promise = document.querySelector('#exercise1');
+    promise.innerText = "promise1";
```
40thieves: I don't think this is quite right. `promise1` will resolve back with a result that is a string. You need to take that string and insert it into the DOM. Here you are making your own string and inserting that into the DOM.

Hint: if you add a parameter to the `.then` handler function can then `console.log` it, that is what we need to insert into the DOM:

```js
function exercise1() {
  var promise1 = resolvedPromise();
  promise1.then(function(result) {
    console.log(result)
    // Then we need to insert the result variable into the DOM
  })
}
```

Does that make sense?

week-5/C-promises/exercise.js
```diff
@@ -21,7 +28,11 @@ function exercise1() {
   Insert the result of promise2 into the #exercise2 element. Hint: .catch()
 */
 function exercise2() {
-  var promise2 = rejectedPromise()
+  var promise2 = rejectedPromise();
+  promise2.catch(function () {
+    var insertPromise2 = document.querySelector('#exercise2');
+    insertPromise2.innerText = 'promise2';
```
40thieves: This is the same problem as exercise1. The solution is similar too 😄

week-5/C-promises/exercise.js
```diff
@@ -32,7 +43,12 @@ function exercise2() {
   takes longer than the others to finish
 */
 function exercise3() {
-  var promise3 = delayedPromise()
+  var promise3 = delayedPromise();
+
+  promise3.then(function() {
+    var resultPromise3 = document.querySelector('#exercise3');
+    resultPromise3.innerText = 'promise3'
```
40thieves: This one too!

week-5/C-promises/exercise.js
```diff
@@ -44,7 +60,15 @@ function exercise3() {
   of your promise and insert it into the #exercise4 element
 */
 function exercise4() {
-  var promise4 = concatPromise()
+  var promise4 = concatPromise();
+    promise4.then(result => {
+      let newResult = result + 'Than';
+      document.querySelector('#exercise4').textContent = newResult;
```
40thieves: This exercise is correct! Maybe you can look at how this one works to solve exercises 1, 2 and 3

than2042: Thank you very much for your time. I will need help. However, I am on holiday for almost 3 weeks from Tuesday. I will try my best to improve it!

40thieves: No problem :) I'm also around on slack if you need some more help

PR #128, opened 2018-06-17
Improve wording of callbacks and promises exercises

40thieves:
Does what it says on the 🛢 In response to students' feedback

cc @djgrant

### DIFF
--- a/week-5/B-callbacks/exercise.js
+++ b/week-5/B-callbacks/exercise.js
@@ -24,8 +24,9 @@ function exercise1() {
   EXERCISE 2
   =======
 
-  A function is provided for you (functionCallsBack) that will call the first
-  parameter as a callback. The callback receives a result as the first parameter
+  A function is provided for you (functionThatCallsBack) that will call the
+  first parameter as a callback. The callback receives a result as the first
+  parameter
 
   Finish writing the callback function to insert the result into the #exercise2
   element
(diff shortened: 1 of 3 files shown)


week-5/B-callbacks/exercise.js
```diff
@@ -21,11 +24,14 @@ function exercise1() {
   EXERCISE 2
   =======
 
-  The function provided for you (functionCallsBack) will call the callback
-  with a result
+  A function is provided for you (functionCallsBack) that will call the first
```
djgrant: Actual function is called `functionThatCallsBack` not functionCallsBack

week-5/B-callbacks/exercise.js
```diff
@@ -37,15 +43,22 @@ function exercise2(result) {
   EXERCISE 3
   =======
 
-  Write a function that calls the callback with the parameter "Hello from the
+  The exercise3 function will be called for you automatically. It is called
+  with a callback function as the first parameter. Call this callback function
+  (hint: use () parentheses) with the parameter "Hello from the callback" and
```
djgrant: I like to put code syntax inside back ticks to attempt to distinguish from English syntax.

week-5/B-callbacks/exercise.js
```diff
@@ -37,15 +43,22 @@ function exercise2(result) {
...
+  it will insert it into the #exercise3 element.
+
+  EXPECTED RESULT: The #exercise3 element has textContent = "Hello from the
   callback"
 
-  What happens to the #exercise3 element? Why does this happen? (Hint: look at
-  line 63 below)
+  Can you explain why we get this result? Write down your explanation in a
+  comment below (Hint: look at line 63 below)
```
djgrant: I think line 63 is already out of date! Maybe something like "Look at the comment inside the function"

week-5/B-callbacks/exercise.js
40thieves: Good catch. Thanks

week-5/B-callbacks/exercise.js
40thieves: You're right that it was out of date, but actually I was referencing the `exercise3` callsite (now on line 77). Updated to not use a line number.

40thieves:
Thanks @djgrant! Pushed a commit to address the feedback

40thieves:
Hopefully you're happy with those fixes @djgrant. Merging

PR #133, opened 2018-07-07
Week 5 and practise first commit

MHDGhanemBalhawan:
Hi there,
Please check my week 5.
Thank you and regards,
Steve

### DIFF
--- a/week-5-practise/index.html
+++ b/week-5-practise/index.html
@@ -1,157 +1,171 @@
 <!DOCTYPE>
 <html lang="en_US">
-	<head>
-		<title>Bikes for Refugees | Scotland</title>
-		<meta charset="utf-8">
-		<link rel="dns-prefetch" href="//fonts.googleapis.com">
-		<link rel="dns-prefetch" href="//maxcdn.bootstrapcdn.com">
-		<link rel="dns-prefetch" href="//code.jquery.com">
-		<link rel="dns-prefetch" href="//cdnjs.cloudflare.com">
-		<link href="https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,300i,400,400i,600,600i,700,700i" rel="stylesheet">
-		<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-alpha.6/css/bootstrap.min.css" integrity="sha384-rwoIResjU2yc3z8GV/NPeZWAv56rSmLldC3R/AZzGRnGxQQKnKkoFVhFQhNUwEyJ" crossorigin="anonymous">
-		<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css">
-		<link rel="stylesheet" href="styles/style.css">
-	</head>
-	<body>
-		<a href="#content" class="sr-only sr-only-focusable">Skip to main content</a>
-		<div class="site-header container">
-			<div class="navbar navbar-toggleable-md navbar-light">
-				<button class="navbar-toggler navbar-toggler-right" type="button" data-toggle="collapse" data-target="#navbarSupportedContent" aria-controls="navbarSupportedContent" aria-expanded="false" aria-label="Toggle navigation">
-					<span class="navbar-toggler-icon"></span>
-				</button>
-				<a href="/" class="navbar-brand">
-					<h1 class="sr-only">Bikes for Refugees Scotland - Providing donated bikes and accessories to refugees and asylum seekers.</h1>
-				</a>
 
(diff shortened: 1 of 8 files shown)


week-5/A-dom-manipulation/exercise.js
```diff
@@ -23,11 +29,34 @@ Task 3
...
+var selectButtonsDiv = document.querySelector(".buttons");
+addParagText.addEventListener("click", function() {
+  addSomeText();
+});
 
+function addSomeText() {
+  console.log("heyy");
+  var newParag = document.createElement("p");
+  selectButtonsDiv.appendChild(newParag);
+  newParag.innerText = "Read More!";
```
40thieves: Normally we would want to add the `innerText` before we append the element into the DOM

week-5-practise/index.html
```diff
@@ -1,157 +1,171 @@
...
+					</div>
+					<div class="form-group">
+						<label for="example-text-input">Your name</label>
+						<input class="form-control" type="text" value="" id="example-text-input">
+					</div>
+					<div class="form-group">
+						<label for="exampleTextarea">Describe yourself</label>
+						<textarea class="form-control" id="exampleTextarea" rows="3"></textarea>
+					</div>
+					<button type="submit" class="btn btn-primary" onclick="ValidateRegisterForm()">Submit</button>
```
40thieves: It's mostly considered bad practice to use `onclick` DOM attributes now. It makes it more difficult to change the code and harder to read if the function calls are split between JS and HTML. Generally you should use `.addEventListener` instead.

week-5/A-dom-manipulation/exercise.js
```diff
@@ -23,11 +29,34 @@ Task 3
...
+*/
+var largerButtons = document.querySelector("#largerLinksBtn");
+
+var selectBtn = document.getElementsByClassName("btn");
+
+largerButtons.addEventListener("click", function() {
+ 
+  var i;
+  for (i = 0; i < selectBtn.length; i++) {
+    selectBtn[i].style.width = "150px";
```
40thieves: I think the question is asking you to change the `font-size` not the width of the links

week-5/A-dom-manipulation/exercise.js
```diff
@@ -23,11 +29,34 @@ Task 3
...
 Task 4
 ======
 
 When the 'Larger links!' button is clicked, the text of all links on the page should increase.
-*/
\ No newline at end of file
+*/
+var largerButtons = document.querySelector("#largerLinksBtn");
+
+var selectBtn = document.getElementsByClassName("btn");
```
40thieves: You're selecting anything with the class `btn` here. However the question is asking you to change the all links. A "link" is just another word for the `<a>` element. How could you change this selector to get all `<a>` elements instead?

week-5/D-webchat/exercise-3.js
```diff
@@ -42,4 +44,40 @@ Task 4
 Use the setInterval function to reload automatically the messages of your webchat every 2 seconds.
 The code responsible to show the messages in the page is in exercise-1.js, so you will need to write your code there :-)
 */
-
+let ticks = 0;
+var requestURL = "https://codeyourfuture.herokuapp.com/api/messages";
+
+var xhttp = new XMLHttpRequest();
+xhttp.onreadystatechange = function() {
```
40thieves: You don't need to use `XMLHttpRequest` here if you are using `fetch`. This code will actually send 2 requests to the same URL because you are using both. I would stick to `fetch` as not many developers use `XMLHttpRequest` any more.

week-5/A-dom-manipulation/exercise.js
MHDGhanemBalhawan: OK. I'll fix it.

week-5/A-dom-manipulation/exercise.js
MHDGhanemBalhawan: I'll change it.Thank you for the advice.

week-5/A-dom-manipulation/exercise.js
MHDGhanemBalhawan: ByTagName. I'll change it, too.

week-5/D-webchat/exercise-3.js
MHDGhanemBalhawan: I'll remove it.

MHDGhanemBalhawan:
Thank you for your review.

MHDGhanemBalhawan:
Hi Ali,
I have updated the files and they are here:
https://github.com/MHDGhanemBalhawan/js-exercises/commit/b5aec30c451d1f77c4334fcf17ce4d0491c23141
I'm having trouble with git  all the time.
Thank you and regards,
Steve

PR #134, opened 2018-07-08
Remove readmes to reduce duplication with syllabus

40thieves:
Removes the `README.md` files in each of the individual directories so that they aren't duplicated on the syllabus, which could have got out of sync.

There's some directories that don't match up with the syllabus already:

- `week-1/G-strings-methods` isn't in the syllabus at all AFAIK. Should we remove it?
- `week-2/B-boolean-literals` is called "Boolean Filters" in the syllabus. Should we update the syllabus? (IMO "Boolean literals" describes the content better)

@kabaros @djgrant @kkarimi


djgrant:
Before we go ahead with this I'd like to take a step back and consider whether we made the right call with this structure (readme in `syllabus` and exercises in `js-exercises`). 

I had a look through the syllabus today and I'm feeling unconvinced that the new structure improves the experience of a student working through the exercises. To progress through the exercises a student now needs to be going constantly back and forth, either opening new tabs or trying to find the correct instructions for a given task. This was easy with the instructions living in the same directory as the file they were working on. 

A secondary concern is that it is now harder to update the exercises. The exercise and their readmes are very tightly coupled, including information like expected result. It adds a big overhead, and an increased likelihood of things getting out of sync, if contributors need to update two repos for each change.

I see that the aim is to have all our teaching material in `syllabus`. Perhaps it's also an aim to have students start from syllabus and work outwards from there. If that's the case, the content in syllabus needs to include links to `js-exercises`. But let's consider that `js-exercises` is installed and run locally - we move from a browser context to an IDE context. 

We should also consder that part of the design of `js-exercises` is to move away from the long pages of content we have in syllabus that students seemed to avoid reading. `js-exercises` breaks the reading material up and forces students to practise what they've learnt at regular intervals. I fear the new structure risks students a) skimming through all the lessons in one go (as they appear in a long syllabus pages), and b) working their way through the exercises without going back and referencing the syllabus. 

If we are to move all this content to syllabus I think we need to do a couple of things:
1. Make the exercises available in a hosted (in browser) IDE, ideally embeded in syllabus pages
2. Ensure that content that is tightly coupled to the exercises themselves (like "expected result") stay with the exercise (both in version control and from the perspective of the student).

40thieves:
Firstly sorry for the extremely long delay in replying.

Having thought about it I think I agree with you. I do like the workflow of keeping the students in the IDE. And you make good points about keeping the exercises in sync with the syllabus.

My main concerns are:
- For London at least, we sometimes struggle with having consistency from the mentors that are teaching. That means I quite like the ability to point at the syllabus document and say "teach this". It nicely scopes the week's teaching IMO. On the other hand, I understand that this is still possible with the exercises structure
- I also quite like the ability to put the syllabus up on the projector and easily walk through all of the content without too much context switching.

So... I don't really know what to do. I'm happy in the meantime the drop this PR - I don't have particularly strong feelings about it, other than the goal of reducing the number of places to update if we want to change the content. Perhaps we should have a discussion with the larger group about solving this problem?

jslopezgithub:
I recently did the first 4 week of exercises as a student and I found myself going back and forth between the syllabus, the markdown file, and a markdown reader loaded with the instructions. 

I prefer how the instructions look on the syllabus. However, keeping the instructions in one place and the code in other tends to cause divergence.

My suggestion would be to keep the instructions on the .js file, remove the .md file and the instructions from the syllabus. That will avoid any possible back and forth or divergence. This PR would have alleviate my experience since I would have allowed me to solely focused on the syllabus. I will ask the students to see what is their experience.

PR #142, opened 2018-10-17
Daniel's homework week-1

DanGhanbari:


week-1-practise/level-1/10-function-output-3.js
```diff
@@ -1,5 +1,6 @@
 function concatenate(firstWord, secondWord, thirdWord) {
     // Write the body of this function to concatenate three words together
+    return firstWord.concat(secondWord).concat(thirdWord)
```
amaalali: Good stuff. You can also make things easier by passing multiple arguments to `concat`. Take a look a the docs here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat#Syntax.

As an interesting note, there are multiple ways to do everything, and you also use the `+` to achieve the same result and there are slight differences in the way they work https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat#Performance.

week-1-practise/level-1/7-logic-error-4.js
```diff
@@ -1,11 +1,11 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+  
+  return a * b * c;
```
amaalali: :+1:

week-1-practise/level-1/7-logic-error-4.js
```diff
@@ -1,11 +1,11 @@
...
 }
 
-var result = multiply(1, 3, 4);
+const result = multiply(1, 3, 4);
```
amaalali: Nice one to recognize that this can be a `const`

week-1-practise/level-3/1-piping.js
```diff
@@ -11,6 +11,7 @@
 */
 
 var startingValue = 5;
+ result = (format( add (multiply (5,2),10)));
```
amaalali: We don't normally leave spaces between the function name and the parameter list

```suggestion
 result = format( add( multiply(5,2), 10));
```

week-1-practise/level-1/3-syntax-error-3.js
```diff
@@ -1,6 +1,6 @@
 // There is a syntax error with this code. Fix it.
 
-function(a, b, c) {
+function addNumbers(a, b, c) {
```
amaalali: Interestingly there is another way to solve this, search for Javascript and anonymous functions

week-1-practise/level-1/5-logic-error-2.js
```diff
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return word.trim;
+  return word.trim();
```
amaalali: Correct! Did you take note of what `word.trim` returned?

DanGhanbari: yes. it removed unnecessary spaces

week-1-practise/level-1/7-logic-error-4.js
DanGhanbari: 😊

week-1-practise/level-1/7-logic-error-4.js
DanGhanbari: 👍

week-1-practise/level-1/10-function-output-3.js
DanGhanbari: Do you mean in this way:
return firstWord + secondWord + thirdWord;
}

PR #143, opened 2018-10-19
Mahshid's homework

MahshidAzami:


week-1-practise/level-1/10-function-output-3.js
```diff
@@ -1,6 +1,6 @@
 function concatenate(firstWord, secondWord, thirdWord) {
-    // Write the body of this function to concatenate three words together
+  return firstWord.concat(secondWord, thirdWord);
```
amaalali: Good to use the fact that concat takes multiple parameters!

week-1-practise/level-1/4-logic-error-1.js
```diff
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
```
amaalali: Great. Quick one to check, ‘word.trim’ returns a different type of JS object. Check it and see what it returns (you can use typeof to help)

week-1-practise/level-2/2-format-currency.js
```diff
@@ -6,7 +6,9 @@
   Write a function that transforms numbers into the format £0.00
 */
 
-function formatCurrency() {}
+function formatCurrency(price) {
+  return "£" + price.toFixed(2);
```
amaalali: Good stuff! What happens if someone used this function with a string? Try it out :)

week-1-practise/level-1/10-function-output-3.js
MahshidAzami: Thanks and sorry for delay in responding. Actually I thought we will be in touch just through the slack. I didn't notice your comment on my GitHub.

week-1-practise/level-1/4-logic-error-1.js
MahshidAzami: I've checked, word.trim return string as I expected. so I think it won't change the type of object.

week-1-practise/level-2/2-format-currency.js
MahshidAzami: I'm not sure I understand your question.

PR #147, opened 2018-11-10
Ahmad HomeWork

Ashaghel:
Finished level1-3 in the homework

### DIFF
--- a/week-1-practise/level-1/1-syntax-error-1.js
+++ b/week-1-practise/level-1/1-syntax-error-1.js
@@ -1,6 +1,6 @@
 // There is a syntax error in this code. Fix it.
 
-function addNumbers(a b c) {
+function addNumbers(a ,b ,c) {
     return a + b + c;
 }
 
@@ -10,5 +10,8 @@ console.log(result); // Expected result 8 - Run the program and make sure it wor
 
 // Answer these questions:
 // 1. How many parameters does the function take?
+//3
 // 2. What is the function name?
+//addNumbers
 // 3. Where is the function called? with what parameters?
+//console.log (results)
--- a/week-1-practise/level-1/1-syntax-error-1.js
+++ b/week-1-practise/level-1/1-syntax-error-1.js
@@ -1,6 +1,15 @@
 // There is a syntax error in this code. Fix it.
+/* ways to write this function
+1-function addNumbers(a, b, c){}
+2- (a, b, c) => {}. and more consice later
+However I read here 
+https://stackoverflow.com/questions/1925976/declaring-functions-in-javascript
+that the first method is the good practice. 
 
-function addNumbers(a ,b ,c) {
+
+*/
+
+let addNumbers = function (a, b, c) {
     return a + b + c;
 }
 
@@ -8,10 +17,11 @@ var result = addNumbers(1, 3, 4);
 console.log(result); // Expected result 8 - Run the program and make sure it works
 
 
-// Answer these questions:
-// 1. How many parameters does the function take?
-//3
-// 2. What is the function name?
-//addNumbers
-// 3. Where is the function called? with what parameters?
-//console.log (results)
+/*Answer these questions:
+ 1. How many parameters does the function take?
+3 parameters a, b, and c.
+2. What is the function name?
+addNumbers()
+3. Where is the function called? with what parameters?
+in line 7, with parameters 1, 3, and 4.
+*/
--- a/week-1-practise/level-1/10-function-output-3.js
+++ b/week-1-practise/level-1/10-function-output-3.js
@@ -1,5 +1,7 @@
 function concatenate(firstWord, secondWord, thirdWord) {
     // Write the body of this function to concatenate three words together
+    return firstWord.concat(secondWord,thirdWord)
+    
 }
 
 var result = concatenate('code', 'your', 'future');
--- a/week-1-practise/level-1/10-function-output-3.js
+++ b/week-1-practise/level-1/10-function-output-3.js
@@ -1,7 +1,7 @@
 function concatenate(firstWord, secondWord, thirdWord) {
     // Write the body of this function to concatenate three words together
-    return firstWord.concat(secondWord,thirdWord)
-    
+    return firstWord.concat(secondWord, thirdWord)
+
 }
 
 var result = concatenate('code', 'your', 'future');
--- a/week-1-practise/level-1/2-syntax-error-2.js
+++ b/week-1-practise/level-1/2-syntax-error-2.js
@@ -1,8 +1,10 @@
 // There is a syntax error in this code. Fix it.
 
-function addNumbers(a, b, c)
-return a + b + c;
+function addNumbers(a, b, c){
 
+
+return a + b + c;
+}
 
 var result = addNumbers(1, 3, 4);
 console.log(result); // Expected result 8 - Run the program and make sure it works
--- a/week-1-practise/level-1/2-syntax-error-2.js
+++ b/week-1-practise/level-1/2-syntax-error-2.js
@@ -1,9 +1,9 @@
 // There is a syntax error in this code. Fix it.
 
-function addNumbers(a, b, c){
+function addNumbers(a, b, c) {
 
 
-return a + b + c;
+    return a + b + c;
 }
 
 var result = addNumbers(1, 3, 4);
--- a/week-1-practise/level-1/3-syntax-error-3.js
+++ b/week-1-practise/level-1/3-syntax-error-3.js
@@ -1,6 +1,6 @@
 // There is a syntax error with this code. Fix it.
 
-function(a, b, c) {
+let addNumbers= function(a, b, c) {
     return a + b + c;
 }
 
--- a/week-1-practise/level-1/3-syntax-error-3.js
+++ b/week-1-practise/level-1/3-syntax-error-3.js
@@ -1,6 +1,6 @@
 // There is a syntax error with this code. Fix it.
 
-let addNumbers= function(a, b, c) {
+let addNumbers = function (a, b, c) {
     return a + b + c;
 }
 
--- a/week-1-practise/level-1/4-logic-error-1.js
+++ b/week-1-practise/level-1/4-logic-error-1.js
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
 
 var result = trimWord("  CodeYourFuture ");
--- a/week-1-practise/level-1/5-logic-error-2.js
+++ b/week-1-practise/level-1/5-logic-error-2.js
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return word.trim;
+  return word.trim();
 }
 
 var result = trimWord("  CodeYourFuture ");
--- a/week-1-practise/level-1/6-logic-error-3.js
+++ b/week-1-practise/level-1/6-logic-error-3.js
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trim(word) {
-  return "word".trim();
+  return word.trim();
 }
 
 var result = trim("CodeYourFuture  ");
--- a/week-1-practise/level-1/7-logic-error-4.js
+++ b/week-1-practise/level-1/7-logic-error-4.js
@@ -1,8 +1,8 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+  
+  return a * b * c;
 }
 
 var result = multiply(1, 3, 4);
--- a/week-1-practise/level-1/7-logic-error-4.js
+++ b/week-1-practise/level-1/7-logic-error-4.js
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function multiply(a, b, c) {
-  
+
   return a * b * c;
 }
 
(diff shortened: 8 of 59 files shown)


week-1-practise/level-1/1-syntax-error-1.js
```diff
@@ -1,6 +1,6 @@
 // There is a syntax error in this code. Fix it.
 
-function addNumbers(a b c) {
+function addNumbers(a ,b ,c) {
```
edwardbrowncross: Good. Normally we put the space after the comma like this:
```suggestion
function addNumbers(a, b, c) {
```
(but that is just for neatness)

week-1-practise/level-1/1-syntax-error-1.js
```diff
@@ -10,5 +10,8 @@ console.log(result); // Expected result 8 - Run the program and make sure it wor
 
 // Answer these questions:
 // 1. How many parameters does the function take?
+//3
 // 2. What is the function name?
-// 3. Where is the function called? with what parameters?
\ No newline at end of file
+//addNumbers
+// 3. Where is the function called? with what parameters?
+//console.log (results)
```
edwardbrowncross: When the question asks where the function is _called_, it means where was it executed (like this: `theFunction()`). It also asks what parameters it was given (what were the values of a, b, and c when it was called).
You have given the place where its return value is subsequently used, which is different.

week-1-practise/level-1/2-syntax-error-2.js
```diff
@@ -1,8 +1,10 @@
 // There is a syntax error in this code. Fix it.
 
-function addNumbers(a, b, c)
-return a + b + c;
+function addNumbers(a, b, c){
+
 
+return a + b + c;
```
edwardbrowncross: It is good to make sure that code inside the function is indented like this:
```suggestion
    return a + b + c;
```
When the code you are writing gets more complex, it will makes it much easier to read!

week-1-practise/level-1/3-syntax-error-3.js
```diff
@@ -1,6 +1,6 @@
 // There is a syntax error with this code. Fix it.
 
-function(a, b, c) {
+let addNumbers= function(a, b, c) {
```
edwardbrowncross: This absolutely works but it is not the most common way! There is another way to write this line that doesn't use an equals sign or let/var/const. Can you comment with the other solution?

Ashaghel:
You do have it! any comments?

week-1-practise/level-1/8-function-output-1.js
```diff
@@ -7,6 +7,10 @@ console.log(result);
 
 // Answer these questions:
 // - What does this program do?
+//generate random numbers everytime it's called, from 0 to one (not included)
```
edwardbrowncross: Are you sure it generates numbers in the range 0 to 1?

Hint: `Math.random` does but `getNumber` does not.

week-1-practise/level-1/8-function-output-1.js
```diff
@@ -7,6 +7,10 @@ console.log(result);
...
 // - How many parameters does the function take?
+//none needed as random() will generate the number alone
 // - What is the function name?
-// - Where is the function called? with what parameters?
\ No newline at end of file
+//getNumber
+// - Where is the function called? with what parameters?
+//line 6 with console log witout any parameters
```
edwardbrowncross: Close! The _result_ is _used_ on line 6, but the function is not _called_ on that line (like in exercise 1).

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -6,9 +6,22 @@ var result = s('code', 'your');
 console.log(result);
 
 // Answer these questions:
-// - What does this program do? Run it and see, or Google some of the methods used ot understand them.
+// - What does this program do? Run it and see, or Google some of the methods used ot understand them. 
+//it will join strings and give back a new string with the sum result. In arrays, it will join arrays by order
+//and create new array with the elements of each parameter 
 // - How many parameters does the function take?
+//as many as wanted
```
edwardbrowncross: Really? Try it with three parameters and see what happens.

What happens if you pass no parameters?

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -6,9 +6,22 @@ var result = s('code', 'your');
...
 // - What is the function name?
+//s
 // - Where is the function called? with what parameters?
+//line 6, code, your
```
edwardbrowncross: The parameters are correct! See previous comments about what is meant by _called_, however.

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -6,9 +6,22 @@ var result = s('code', 'your');
...
 // - What does the function return?
-// - How can this function be improved?
\ No newline at end of file
+//a new string with a joined strings codeyour
+// - How can this function be improved?
+//by giving a more clear names for the functions and the paramters as well as adding a space between both strings
```
edwardbrowncross: I definitely agree on the name! What would you call it?

Without knowing what the function is going to be used for, it's hard to say whether the space is better. It does seem to make more sense for the example given doesn't it.

week-1-practise/level-2/1-tax.js
```diff
@@ -5,7 +5,9 @@
   Sales tax is 20% of the price of the product
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(num) {
+return num * 0.2
```
edwardbrowncross: Very good! Remember that indentation though.
```suggestion
    return num * 0.2
```

week-1-practise/level-3/1-piping.js
```diff
@@ -28,4 +30,4 @@ function format(num) {
 }
 
 console.log("Expected result: 20.00");
-console.log("Actual result: " + result);
+console.log("Actual result: " + result());
```
edwardbrowncross: It looks like you're trying to call result like it's a function. It should be a string though, right?

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,10 +7,14 @@
   Find a way to add 1% to all currency conversions (think about the DRY principle)
 */
 
-function convertToUSD(price) {}
-function convertToBRL(price) {}
+function convertToUSD(price) {
+  return price *1.3*1.01;
+}
+function convertToBRL(price) {
+  return convertToUSD(price)*3.73;
```

The question I would like to ask you is this: if another developer came to this code, would they understand why `convertToBRL` must call `convertToUSD` and not do the conversion straight away? Is it clear from reading the code that all foreign currency conversions should have an extra 1% applied?

week-1-practise/level-1/1-syntax-error-1.js
Ashaghel: so it's just like writing in English. Sure I will do that

week-1-practise/level-1/8-function-output-1.js
Ashaghel: I have updated them all ;D

week-1-practise/level-1/3-syntax-error-3.js
Ashaghel: ways to write this function
1-function addNumbers(a, b, c){}
2- (a, b, c) => {}. and more concise later
However, I read here 
https://stackoverflow.com/questions/1925976/declaring-functions-in-javascript
that the first method is a good practice.

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,10 +7,17 @@
...
 
-function convertToUSD(price) {}
-function convertToBRL(price) {}
+//As all currency conversions raised by 1%
+//function extra(price){
+  //return 1.01* price}
 
-/* DO NOT EDIT BELOW THIS LINE */
+function convertToUSD(price) {
+  return (1.3* extra(price)).toFixed(2);
```
edwardbrowncross: Much clearer 👌 

(It looks like you commented out your `extra` function though)

week-1-practise/level-1/10-function-output-3.js
```diff
@@ -1,5 +1,7 @@
 function concatenate(firstWord, secondWord, thirdWord) {
     // Write the body of this function to concatenate three words together
+    return firstWord.concat(secondWord, thirdWord)
+
 }
 
 var result = concatenate('code', 'your', 'future');
```
christopherallanperry: Good

week-1-practise/level-1/2-syntax-error-2.js
```diff
@@ -1,8 +1,10 @@
...
-function addNumbers(a, b, c)
-return a + b + c;
+function addNumbers(a, b, c) {
 
 
+    return a + b + c;
+}
+
 var result = addNumbers(1, 3, 4);
 console.log(result); // Expected result 8 - Run the program and make sure it works
```
christopherallanperry: Good

week-1-practise/level-1/3-syntax-error-3.js
```diff
@@ -1,6 +1,6 @@
 // There is a syntax error with this code. Fix it.
 
-function(a, b, c) {
+let addNumbers = function (a, b, c) {
     return a + b + c;
 }
```
christopherallanperry: With this one, you've created an anonymous function and have assigned it to a variable. You could also create it as a named function which works equally well ass follows:

```
function addNumbers(a, b, c) {
    return a + b + c;
}

```

week-1-practise/level-1/4-logic-error-1.js
```diff
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
 
 var result = trimWord("  CodeYourFuture ");
```
christopherallanperry: Good. No problems there.

week-1-practise/level-1/5-logic-error-2.js
```diff
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return word.trim;
+  return word.trim();
 }
 
 var result = trimWord("  CodeYourFuture ");
```
christopherallanperry: Again, no problems.

week-1-practise/level-1/6-logic-error-3.js
```diff
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trim(word) {
-  return "word".trim();
+  return word.trim();
 }
 
 var result = trim("CodeYourFuture  ");
```
christopherallanperry: That's fine.

week-1-practise/level-1/7-logic-error-4.js
```diff
@@ -1,8 +1,8 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+
+  return a * b * c;
 }
 
 var result = multiply(1, 3, 4);
```

week-1-practise/level-1/8-function-output-1.js
```diff
@@ -7,6 +7,10 @@ console.log(result);
...
 // - What does this program do?
+//generate random numbers everytime it's called, from 0 to 10 ( 10 not included)
 // - How many parameters does the function take?
+//none needed, as random() will generate the numbers alone
 // - What is the function name?
-// - Where is the function called? with what parameters?
\ No newline at end of file
+//getNumber()
+// - Where is the function called? with what parameters?
+//line 5
```
christopherallanperry: No problems here. Good answers

week-1-practise/level-1/9-function-output-2.js
christopherallanperry: Just adding to Edward's comment, it's important that you run the programs as you're working on them to see that they do as you expect. As you move forward into more complex program's, you'll start to see the need for more in depth testing, where you'll feed, not only the expected values into functions, but also incorrect values to find out whether they'll fail gracefully.

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -6,9 +6,24 @@ var result = s('code', 'your');
...
+/* - How can this function be improved?
+by giving a more clear names for the functions and the paramters as well as adding a space between both strings
+stringJoiner(str1, str2)
+*/
+function s(w1, w2) {
+    return w1.concat(" ", w2);
+}
+
+var result = s('code', 'your');
+console.log(result);
```
christopherallanperry: Just to round things off, it would have been good to see you use your suggested 'clear names' in the the improved function.

week-1-practise/level-2/1-tax.js
```diff
@@ -18,3 +20,4 @@ console.log("Product: " + product);
 console.log("Price: £" + price);
 console.log("Sales tax: £" + salesTax);
 console.log("Total: £" + (price + salesTax));
+//I noticed to get use of the prettier function you need to press alt+shift+f. it is not automatical
```
christopherallanperry: No problems there.

week-1-practise/level-2/2-format-currency.js
```diff
@@ -6,7 +6,9 @@
   Write a function that transforms numbers into the format £0.00
 */
 
-function formatCurrency() {}
+function formatCurrency(num) {
+  return "£" + num.toFixed(2)
+}
 
 /* DO NOT EDIT BELOW THIS LINE */
 /* --------------------------- */
```

week-1-practise/level-2/3-currency-conversion.js
```diff
@@ -5,7 +5,9 @@
   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
 */
 
-function convertToUSD() {}
+function convertToUSD(pound) {
+  return pound*1.4;
+}
 
 /* DO NOT EDIT BELOW THIS LINE */
 /* --------------------------- */
```
christopherallanperry: As per the comment on the previous question, what happens if this function is being used to convert currencies other than "pound" to $? Would the way the function is written still make sense, particularly when read by another programmer at a later date?

week-1-practise/level-2/4-currency-conversion-2.js
christopherallanperry: Also see previous comments

PR #150, opened 2018-11-13
All exercises Done

Abdoulrazack95:
This my pull request

### DIFF
--- a/week-1-practise/level-1/1-syntax-error-1.js
+++ b/week-1-practise/level-1/1-syntax-error-1.js
@@ -1,12 +1,11 @@
 // There is a syntax error in this code. Fix it.
 
-function addNumbers(a b c) {
-    return a + b + c;
+function addNumbers(a, b, c) {
+  return a + b + c;
 }
 
 var result = addNumbers(1, 3, 4);
 console.log(result); // Expected result 8 - Run the program and make sure it works
-
 
 // Answer these questions:
 // 1. How many parameters does the function take?
--- a/week-1-practise/level-1/1-syntax-error-1.js
+++ b/week-1-practise/level-1/1-syntax-error-1.js
@@ -9,5 +9,8 @@ console.log(result); // Expected result 8 - Run the program and make sure it wor
 
 // Answer these questions:
 // 1. How many parameters does the function take?
+// The fonction takes three parameters.
 // 2. What is the function name?
+//The function's name is addNumbers.
 // 3. Where is the function called? with what parameters?
+// The function is called outside of the function e.g. : var result - addNumbers();
--- a/week-1-practise/level-1/1-syntax-error-1.js
+++ b/week-1-practise/level-1/1-syntax-error-1.js
@@ -13,4 +13,4 @@ console.log(result); // Expected result 8 - Run the program and make sure it wor
 // 2. What is the function name?
 //The function's name is addNumbers.
 // 3. Where is the function called? with what parameters?
-// The function is called outside of the function e.g. : var result = addNumbers();. The parameters are called with these three agreements : 1, 2, 3;
+// The function is called outside of the function e.g. : var result = addNumbers();. The parameters are called with these three agreements : 1, 3, 4;
--- a/week-1-practise/level-1/1-syntax-error-1.js
+++ b/week-1-practise/level-1/1-syntax-error-1.js
@@ -9,7 +9,7 @@ console.log(result); // Expected result 8 - Run the program and make sure it wor
 
 // Answer these questions:
 // 1. How many parameters does the function take?
-// The fonction takes three parameters.
+// The fonction takes three parameters which are a, b, c.
 // 2. What is the function name?
 //The function's name is addNumbers.
 // 3. Where is the function called? with what parameters?
--- a/week-1-practise/level-1/1-syntax-error-1.js
+++ b/week-1-practise/level-1/1-syntax-error-1.js
@@ -13,4 +13,4 @@ console.log(result); // Expected result 8 - Run the program and make sure it wor
 // 2. What is the function name?
 //The function's name is addNumbers.
 // 3. Where is the function called? with what parameters?
-// The function is called outside of the function e.g. : var result - addNumbers();
+// The function is called outside of the function e.g. : var result = addNumbers();. The parameters are called with these three agreements : 1, 2, 3;
--- a/week-1-practise/level-1/8-function-output-1.js
+++ b/week-1-practise/level-1/8-function-output-1.js
@@ -7,7 +7,7 @@ console.log(result);
 
 // Answer these questions:
 // - What does this program do?
-// Math.random() return a number in the range 0 and 1.
+// Math.random() returns a number in the range 0 and 1.
 // - How many parameters does the function take?
 // There is no parameters;
 // - What is the function name?
--- a/week-1-practise/level-1/8-function-output-1.js
+++ b/week-1-practise/level-1/8-function-output-1.js
@@ -1,5 +1,5 @@
 function getNumber() {
-    return Math.random() * 10;
+  return Math.random() * 10;
 }
 
 var result = getNumber();
@@ -7,6 +7,10 @@ console.log(result);
 
 // Answer these questions:
 // - What does this program do?
+// this program generate a random number whenever it is called between 0 to 1.
 // - How many parameters does the function take?
+// There is no parameters;
 // - What is the function name?
+// The function's name is getNumber.
 // - Where is the function called? with what parameters?
+// the function is called at line 5. with getNumber.
--- a/week-1-practise/level-1/8-function-output-1.js
+++ b/week-1-practise/level-1/8-function-output-1.js
@@ -7,7 +7,7 @@ console.log(result);
 
 // Answer these questions:
 // - What does this program do?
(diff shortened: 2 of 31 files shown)


week-1-practise/level-1/1-syntax-error-1.js
```diff
@@ -1,14 +1,16 @@
...
-
 // Answer these questions:
 // 1. How many parameters does the function take?
+// The fonction takes three parameters which are a, b, c.
 // 2. What is the function name?
-// 3. Where is the function called? with what parameters?
\ No newline at end of file
+//The function's name is addNumbers.
+// 3. Where is the function called? with what parameters?
+// The function is called outside of the function e.g. : var result - addNumbers();
```
edwardbrowncross: What about the parameters it is called with? Do you understand the question?

week-1-practise/level-1/8-function-output-1.js
```diff
@@ -1,12 +1,16 @@
...
-    return Math.random() * 10;
+  return Math.random() * 10;
 }
 
 var result = getNumber();
 console.log(result);
 
 // Answer these questions:
 // - What does this program do?
+// this program generates a random number whenever it is called between 0 to 1.
```
edwardbrowncross: It it between 0 and 1?

Hint: `Math.random()` produces a number between 0 and 1 but `getNumber` does not

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -1,14 +1,20 @@
 function s(w1, w2) {
-    return w1.concat(w2);
+  return w1.concat(w2);
 }
 
-var result = s('code', 'your');
+var result = s("code", "your");
```
edwardbrowncross: It looks like you changed the single quotes to double quotes. Single quotes are also fine. In JS you can use either of them to define a string.

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -1,14 +1,20 @@
...
 console.log(result);
 
 // Answer these questions:
 // - What does this program do? Run it and see, or Google some of the methods used ot understand them.
+// This program will join together two or more strings.
```
edwardbrowncross: Are you certain it can join more than two? How?

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -1,14 +1,20 @@
...
 // - How many parameters does the function take?
+// The funtion takes 2 parameters.
 // - What is the function name?
+// the function name is s.
 // - Where is the function called? with what parameters?
+// The function is called at line 5.
```
edwardbrowncross: That's right. What about the parameters it is called with?

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -1,14 +1,20 @@
...
 // - What does the function return?
-// - How can this function be improved?
\ No newline at end of file
+// the function return "codeyour".
+// - How can this function be improved?
+// this function can be improved strings with space inside the parenthese of concat.
```
edwardbrowncross: It could be better to add a space, but it's hard to say without knowing what the function will be used for. There's one other thing I would definitely want to improve. Remember that it is not only important that the code works well but also that it is easy for people reading it to understand. Do you know what I am talking about?

week-1-practise/level-2/1-tax.js
```diff
@@ -5,7 +5,9 @@
   Sales tax is 20% of the price of the product
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  return price * (20 / 100);
```
edwardbrowncross: Excellent! I like that it is very clear that it is adding 20%.

There is a way to write this without both a multiplication and a division. It can be done with only 1 operation. Can you comment to show me how?

week-1-practise/level-2/3-currency-conversion.js
```diff
@@ -5,7 +5,9 @@
   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
 */
 
-function convertToUSD() {}
+function convertToUSD(price) {
+  return "$" + Math.round(price * 1.29631);
```
edwardbrowncross: Well done for putting in the effort to look up the current exchange rate and not just using the value in the question! I also like that you've tried to format it with a currency symbol, even though the question didn't ask for that.

I'm not sure that rounding is necessarily a good thing to do. We could be overcharging or undercharging our customers by removing the cents! How could you improve this?

As an extra (difficult) question, can you think of a time when adding the dollar sign would not be a good thing to do and might actually cause problems?

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,8 +7,12 @@
   Find a way to add 1% to all currency conversions (think about the DRY principle)
 */
 
-function convertToUSD(price) {}
-function convertToBRL(price) {}
+function convertToUSD(price) {
+  return price * 1.3 * (1 + 1 / 100);
```
edwardbrowncross: Correct! Can you simplify `1 + 1 / 100` at all to reduce the number of calculations being done on this line?

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,8 +7,12 @@
...
+}
+function convertToBRL(price) {
+  return convertToUSD(price) * 3.78328;
```

The question I would like to ask you is this: if another developer came to this code, would they understand why convertToBRL must call convertToUSD and not do the conversion straight away? Is it clear from reading the code that all foreign currency conversions should have an extra 1% applied?

week-1-practise/level-3/1-piping.js
```diff
@@ -11,10 +11,12 @@
 */
 
 var startingValue = 5;
+var result = function() {
+  return "£" + (startingValue * 2 + 10).toFixed(2);
```
edwardbrowncross: I think the question would like to test that you can make use of the `add`, `multiply`, and `format` functions on lines 20, 24 and 28 to produce the correct result.

Can you push a new solution that does this?

Also: you don't need to wrap it all in a function. This is perfectly valid:
```js
var result = "£" + (startingValue * 2 + 10).toFixed(2);
```

week-1-practise/level-2/3-currency-conversion.js
```diff
@@ -5,7 +5,9 @@
   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
 */
 
-function convertToUSD() {}
+function convertToUSD(price) {
```
edwardbrowncross: One small thing to think about: having good function parameter names is important. When other developers on your team see your functions, it can help them to understand how to use them. The names should make it clear what input is expected.

In this case, the input must be in GBP (pounds), otherwise the conversion does not work. How could you change the function parameter name (which you called price) to make this clearer?

week-1-practise/level-1/1-syntax-error-1.js
Abdoulrazack95: Yes i understand but i forgot to answer it i will do it now.

week-1-practise/level-1/9-function-output-2.js
Abdoulrazack95: Yes, you are right. I use double quotes but i will use both of them.

week-1-practise/level-1/9-function-output-2.js
Abdoulrazack95: I meant concatenates. It does not join.

week-1-practise/level-1/9-function-output-2.js
Abdoulrazack95: Sorry, i forgot to mention the arguments.

week-1-practise/level-3/1-piping.js
```diff
@@ -11,9 +11,7 @@
 */
 
 var startingValue = 5;
-var result = function() {
-  return "£" + (startingValue * 2 + 10).toFixed(2);
-};
+var result = "£" + (startingValue * 2 + 10).toFixed(2);
```
edwardbrowncross: It get's the right answer, but not in quite the way the question is looking for. You have copied what the functions do, but the question would like your to _call_ the functions.

For example, if you only wanted result to be doubled you could have:
```js
var startingValue = 5;
var result = multiply(startingValue, 2);
```

The question would like you to double, add 10 and format in a way like this (and all on the same line).

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,11 +7,11 @@
...
 
-function convertToUSD(price) {
-  return "$" + price * 1.3 * (1 + 1 / 100);
+function convertToUSD(priceInGBP) {
+  return price * 1.3 * 1.01;
 }
-function convertToBRL(price) {
-  return "R$" + convertToUSD(price) * 3.78328;
+function convertToBRL(priceInGBP) {
+  return price * 3.78 * 1.01;
```
edwardbrowncross: This is quite easy to understand and I like that. The question suggests you should think about DRY (don't repeat yourself) principals.

Imagine if tomorrow we wanted to change it to 2% charge on all currency conversions. We would need to to change it in two places. Maybe if we handle hundreds of currencies we would need to change it in hundreds of places.

As an extra improvement, can you find a way to use functions so that you'd only need to change the number in one place?

PR #151, opened 2018-11-13
Maaly  home work 

MaalyALkaraki:


week-1-practise/level-1/9-function-output-2.js
```diff
@@ -1,14 +1,22 @@
...
-// - How can this function be improved?
\ No newline at end of file
+// it join two strings togather
+// - How many parameters does the function take?It took two
+// - What is the function name? s
+// - Where is the function called? with what parameters? In line no.5 with two parameters 'code' and 'your'
+// - What does the function return? codeyour
+// - How can this function be improved? we could change the name of the function and paramerers so it shows
+// it will be easy to understand the function
+/*function string(str1, str2) {
```
edwardbrowncross: Very good answer! The name is certainly the biggest problem with this function at the moment and your reason you gave is exactly right.

In coding, the word 'string' has a special meaning of text data, so maybe `string` is still not the best name it could have.

week-1-practise/level-2/1-tax.js
```diff
@@ -5,8 +5,9 @@
   Sales tax is 20% of the price of the product
 */
 
-function calculateSalesTax() {}
-
+function calculateSalesTax(value) {
+  return value * (20 / 100);
```
edwardbrowncross: Excellent! I like that it is very clear that it is adding 20%.

There is a way to write this without both a multiplication and a division. It can be done with only 1 operation. Can you comment to show me how?

week-1-practise/level-2/3-currency-conversion.js
```diff
@@ -5,7 +5,9 @@
   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
 */
 
-function convertToUSD() {}
+function convertToUSD(priceInPound) {
```
edwardbrowncross: Very good! I really like that you have used a function parameter name that makes it really clear how this function should be used. 🏆

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,9 +7,12 @@
...
 */
 
-function convertToUSD(price) {}
-function convertToBRL(price) {}
-
+function convertToUSD(price) {
+  return price * 1.4 * 1.01;
+}
+function convertToBRL(price) {
+  return convertToUSD(price) * 3.80363;
```

The question I would like to ask you is this: if another developer came to this code, would they understand why convertToBRL must call convertToUSD and not do the conversion straight away? Is it clear from reading the code that all foreign currency conversions should have an extra 1% applied?

week-1-practise/level-3/1-piping.js
```diff
@@ -11,7 +11,8 @@
 */
 
 var startingValue = 5;
-
+//var result = format(add(10, multiply(startingValue, 2)));
+var result = format(add(10, multiply(startingValue, 2))).match(/\d+/g);
```
edwardbrowncross: Excellent! I'm intrigued by the regular expression matcher at the end. What's that for? The solution on line 14 looks fine!

week-1-practise/level-1/10-function-output-3.js
```diff
@@ -1,6 +1,7 @@
...
-    // Write the body of this function to concatenate three words together
+  // Write the body of this function to concatenate three words together
+  return '"' + firstWord.concat(secondWord, thirdWord) + '"';
 }
 
-var result = concatenate('code', 'your', 'future');
-console.log(result); // expected result = "codeyourfuture"
\ No newline at end of file
+var result = concatenate("code", "your", "future");
+console.log(result); // expected result = "codeyourfuture"
```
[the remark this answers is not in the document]
maaly11: Thank you I was trying to mach it exactly with the expected result

week-1-practise/level-1/8-function-output-1.js
```diff
@@ -1,12 +1,13 @@
...
 var result = getNumber();
 console.log(result);
 
 // Answer these questions:
 // - What does this program do?
-// - How many parameters does the function take?
-// - What is the function name?
-// - Where is the function called? with what parameters?
\ No newline at end of file
+//-returns a floating-point, pseudo-random number between range [0,1) , 0 (inclusive) and 1 (exclusive).
```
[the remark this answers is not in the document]
maaly11: It means that  Math.random() function returns a number that is greater or equal to 0 and less than 1.
And yes it is correct to say that  Math.random() produces numbers in the range 0 to 1, but getNumber() does not. 
because  :
Math.random()  will produce a number  greater or equal to 0 and less than 1
getNumber() will produce a number  greater or equal to 0 and less than 10

week-1-practise/level-1/9-function-output-2.js
maaly11: yes you are right may be using  
connect 
as a name  will  be better

week-1-practise/level-2/1-tax.js
maaly11: It will be :
return value * 0.2  ;

week-1-practise/level-2/4-currency-conversion-2.js
maaly11: yes may be i should add some comments to explain the functions

week-1-practise/level-3/1-piping.js
maaly11: 1.  The matcher  I  used it to remove all non numeric characters from the result so the out put will be just the Expected result: 20,00  without the pound symbol from the the function format 
2. The solution on line 14 is fine put the result will be £ 20.00 not just 20.00 
3. I was trying to find a matcher that will return  20.00 put i could not . so if you know a mathcer that will return 20.00 I will be more than happy to use .

week-1-practise/level-2/1-tax.js
maaly11: return value * 0.2 ;

maaly11:
```suggestion
  return value * 0.2;
```

week-1-practise/level-3/1-piping.js
edwardbrowncross: I see! I think maybe the expected result is wrong in this case 🤔

week-1-practise/level-2/1-tax.js
christopherallanperry: Perfect

week-1-practise/level-2/2-format-currency.js
```diff
@@ -6,7 +6,9 @@
   Write a function that transforms numbers into the format £0.00
 */
 
-function formatCurrency() {}
+function formatCurrency(value) {
+  return "£" + value.toFixed(2);
+}
```
christopherallanperry: In the context of the question asked, your answer is right. However, it's worth bearing in mind that the function is now very firmly tied to being of use with GBP (£) only. Can you think of a way in which you might make the function more generic?

week-1-practise/level-2/3-currency-conversion.js
```diff
@@ -5,7 +5,9 @@
...
+  return "$" + priceInPound * 1.4;
+}
```
christopherallanperry: As with my last comment, the function is now tied not only to specific currencies, but also to a particular exchange rates, as both have been hard-coded into the function. One to think about as your skills develop.

week-1-practise/level-2/4-currency-conversion-2.js
christopherallanperry: Do you think it would be of use to break out the 1% into its own function?

week-2-practise/level-1/6-map.js
```diff
@@ -11,7 +11,7 @@ function tidyUpString(str) {
 }
 
 var mentors = ["/Daniel ", "irina ", " Gordon", "ashleigh "];
-var mentorsTidy; // ONLY EDIT THIS LINE
+var mentorsTidy = mentors.map(tidyUpString); // ONLY EDIT THIS LINE
 
 console.log(mentorsTidy);
```
christopherallanperry: That's good

week-2-practise/level-1/5-indexOf.js
```diff
@@ -3,7 +3,7 @@
 
 var itemToFind = 7;
 var arr = [3, 5, 61, 7, 123];
-var index; // ONLY EDIT THIS LINE
+var index = arr.indexOf(itemToFind); // ONLY EDIT THIS LINE
 
 console.log(index);
```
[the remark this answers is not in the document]
maaly11: yes sure because the array index starts from 0

week-2-practise/level-2/6-map.js
```diff
@@ -4,7 +4,10 @@
 //    - each item in `percentages` formatted by your function
 
 var percentages = [1, 23, 92, 18];
-
+function strWithPercentage(number) {
+  return number + "%";
```
christopherallanperry: It's always handy to remember that using the `+` operator on any type plus a string, will give a string output e.g.
```
var x = true + "test";
x; // "truetest"
```

week-2-practise/level-2/2-validation.js
```diff
@@ -4,7 +4,9 @@
 // - is less than or equal to 100
 // Tip: write other small functions for each requirement
 
-function validate(num) {}
+function validate(num) {
+  return typeof num === "number" && num > 0 && num <= 100;
+}
```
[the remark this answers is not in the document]
maaly11: i found that if we use the Math.sign() function to returns the sign of a number it will return 0 for o  because of this i didnt  include it in the function but if i will concide it positive than i will add num>=0
```suggestion

```
```suggestion
console.log(Math.sign(0));
// expected output: 0
```

week-2-practise/level-3/1-boolean.js
```diff
@@ -16,3 +15,18 @@ if (isHappy == true) {
...
+2. we better not use the == unless we know all the possible coercion and we need it in the function
+so 
+3.because of this i think we have to remove the    == true    from the if function and remove 
+the double quote from  false in      isHappy = false;
+4.it really depends upon the expected values for boolean and how you want the code to work.
+if you know in advance that it's only ever going to have a true or false value, then comparing it
+ explicitly with
+if (booleanValue)
+is more compact and arguably cleaner/better.
+
```
christopherallanperry: Good explanation.

week-2-practise/level-2/2-validation.js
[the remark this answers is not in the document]
maaly11:
```suggestion
function validate(num) {
  return typeof num === "number" && num >= 0 && num <= 100;
}
```

week-2-practise/level-3/2-validation.js
```diff
@@ -6,7 +6,21 @@
...
+    return Number.isInteger(number);
+  }
+  function isNotInclude() {
+    return !excludedNums.includes(number);
+  }
+  return isNumber() && isInteger() && isNotInclude();
+
+  // If we don't want to use small functions we can write it this way
+  // return (typeof number === "number" && Number.isInteger(number) &&!excludedNums.includes(number));
+}
```
christopherallanperry: I'd argue that in this instance, it would be cleaner to not use the functions. Functions are intended to be reusable, but in this instance, they would not be accessible to any code outside the `validate()` function, so not reusable. Also, your code would be more concise.

week-2-practise/level-3/3-sort.js
```diff
@@ -4,12 +4,29 @@
...
 
+function isNumber(element) {
+  return typeof element === "number";
+}
+console.log(
+  " Did we change the variable type in the arrays nums and nums2 ? " +
+    !sortedNums.every(isNumber)
+);
+/* I tried the following function but it didn't work ,it gave me an error that the function Array.observe is not A function 
+can you tell me why ?
```
christopherallanperry: Check the warning message at the top of this page - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/observe - also the browser compatibility table lower down. Where did you find that `Array.observe()` was an option for this task?

week-2-practise/level-3/4-slice.js
```diff
@@ -5,7 +5,10 @@
 // - insert the replacement value at the provided index
 
 function replace(arr, index, value) {
-  return; // complete this statement
+  return arr
+    .slice(0, index)
+    .concat(value)
+    .concat(arr.slice(index + 1)); // complete this statement
 }
```
christopherallanperry: Did you look at using `splice()` at all?

week-2/E-conditionals/exercise.js
```diff
@@ -4,18 +4,19 @@
...
+  console.log("Hi, My name is " + name);
+}
 
-console.log("Hi, My name is " + name);
-console.log("Hi, My name is " + name2);
-
-
+if (role2 === "student") {
+  console.log("Hi, My name is " + name2);
+}
```
christopherallanperry: Could you write a function to handle the task, rather than repeating the same `if` statement with different names/roles?

week-2/F-logical-operators/exercise2.js
```diff
@@ -5,17 +5,28 @@
...
   --------------------------- */
 
 console.log("Is -10 is a negative number?", isNegative(-10));
 console.log("Is 5 a negative number?", isNegative(5));
-console.log("Is 10 in the range 5-10?" isBetween5and10(10))
-console.log("Is Daniel a short name?" isLongName("Daniel"));
-console.log("Does Daniel start with 'D'?" startsWithD("Daniel"));
+console.log("Is 10 in the range 5-10?", isBetween5and10(10));
+console.log("Is Daniel a long name?", isLongName("Daniel"));
+console.log("Does Daniel start with 'D'?", startsWithD("Daniel"));
```
christopherallanperry: All good here

week-2/G-conditionals-2/exercise-3.js
```diff
@@ -9,7 +9,17 @@
...
+  }
+  if (mark >= 60 && mark < 80) {
+    return "B";
+  }
+  if (mark < 60 && mark >= 50) {
+    return "C";
+  } else {
+    return "F";
+  }
 }
```
christopherallanperry: You could also make use of `elseif` here

week-2/L-array-methods-2/exercise2.js
```diff
@@ -7,8 +7,20 @@
...
+  word.shift();
+  word.unshift(letterUp);
+  return word.join(" ");
+}
+/*function capitalise(str) {
+  word = str.split(" ");
+  for (var i = 0, x = word.length; i < x; i++) {
+    word[i] = word[i][0].toUpperCase() + word[i].substr(1);
+  }
+  return word.join(" ");
```
christopherallanperry: It would be worth you checking whether this line gives you the expected result.

week-3/F-array-forEach/exercise.js
```diff
@@ -3,11 +3,31 @@
...
+    if (num % 5 == 0) {
+      return "Buzz";
+    } else {
+      return num;
+    }
+  })
+  .forEach(function(name, index) {
+    console.log(name);
+  });
+console.log(print);
```
christopherallanperry: Have another look at the expected output vs what your code produces - does the expected answer include 0?

week-2-practise/level-3/4-slice.js
maaly11: i know that there is more than way to  solve the problem but because the problem title was slice so i used  slice  , I will try to use splice and post it here for your review .

week-3/F-array-forEach/exercise.js
maaly11: var print = arr.map(function(num) {
  if (num == 0) {
    return num;
  }
  if (num % 3 == 0 && num % 5 == 0) {
    return "FizzBuzz";
  }
  if (num % 3 == 0) {
    return "Fizz";
  }
  if (num % 5 == 0) {
    return "Buzz";
  } else {
    return num;
  }
});
print.forEach(function(name) {
  console.log(name);
});

maaly11: yes i noticed that and i fixed it but i didn't commit it yet but this is how  i did it

week-2/L-array-methods-2/exercise2.js
maaly11: i did but because i was trying to try the other function i forget to un comment it

week-2-practise/level-3/3-sort.js
maaly11: No  but he was asking  if there was any change happened to the array and array observe  one of the method which will .

PR #154, opened 2018-11-16
Homework done_week 1.js_Paulina

PaulinaGrabania:


PaulinaGrabania:
Homework done_week1.js_Paulina

week-1-practise/level-1/10-function-output-3.js
```diff
@@ -1,6 +1,6 @@
 function concatenate(firstWord, secondWord, thirdWord) {
-    // Write the body of this function to concatenate three words together
+	return firstWord + secondWord + thirdWord; // Write the body of this function to concatenate three words together
```
edwardbrowncross: Yep. This will work. Everyone else used this function for their solution:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat
You might like to look at their pull requests to see what they did. Actually I like yours because it is easy to understand.

week-1-practise/level-1/8-function-output-1.js
```diff
@@ -1,12 +1,12 @@
...
 var result = getNumber();
 console.log(result);
 
 // Answer these questions:
-// - What does this program do?
-// - How many parameters does the function take?
-// - What is the function name?
-// - Where is the function called? with what parameters?
\ No newline at end of file
+// - What does this program do? return a random number from between 0-9(inclusive)
```
edwardbrowncross: Very almost right (and closer than most people got)!

It's correct that it will never return 10 exactly, but because it does not only return whole (integer) numbers, it can also return `9.1` or `9.99999` or any other decimal between 9 and 10.

week-1-practise/level-2/3-currency-conversion.js
```diff
@@ -5,14 +5,16 @@
   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
 */
 
-function convertToUSD() {}
+function convertToUSD(price) {
```
edwardbrowncross: One small thing to think about: having good function parameter names is important. When other developers on your team see your functions, it can help them to understand how to use them. The names should make it clear what input is expected.

In this case, the input must be in GBP (pounds), otherwise the conversion does not work. How could you change the function parameter name (which you called price) to make this clearer?

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -6,18 +6,23 @@
...
 */
-
-function convertToUSD(price) {}
-function convertToBRL(price) {}
-
+function transactionFee(price) {
+	return price * 0.01;
+}
+function convertToUSD(price) {
+	return price * 1.4 + transactionFee(price);
```
edwardbrowncross: So close! You have definitely understood how to use functions to abstract important logic (DRY/Don't Repeat Yourself principle) and that is good to see.

There is a slight error here, because the result of `price * 1.4` is a number in USD but the `transactionFee(price)` is still in GBP. You need to also convert the transaction fee to the correct currency or it makes no sense to add them.

Do you want to try and re-arrange the maths to correct this?

week-1-practise/level-3/1-piping.js
```diff
@@ -11,21 +11,21 @@
 */
 
 var startingValue = 5;
-
+var result = format(startingValue * 2 + 10);
```
edwardbrowncross: It get's the right answer, but not in quite the way the question is looking for. You have copied what the `add` and `multiply` functions do, but the question would like your to _call_ the functions directly, like how you have called the `format` function.

PaulinaGrabania:
Thank you Edward very much for checking...I will read everything and check again tomorrow.

PR #156, opened 2018-11-16
Boubacar exercise.js

boubacardllo:
This is my JavaScript exercise

### DIFF
--- a/week-1-practise/level-1/1-syntax-error-1.js
+++ b/week-1-practise/level-1/1-syntax-error-1.js
@@ -1,12 +1,11 @@
 // There is a syntax error in this code. Fix it.
 
-function addNumbers(a b c) {
-    return a + b + c;
+function addNumbers(a, b, c) {
+  return a + b + c;
 }
 
 var result = addNumbers(1, 3, 4);
 console.log(result); // Expected result 8 - Run the program and make sure it works
-
 
 // Answer these questions:
 // 1. How many parameters does the function take?
--- a/week-1-practise/level-1/10-function-output-3.js
+++ b/week-1-practise/level-1/10-function-output-3.js
@@ -1,6 +1,7 @@
 function concatenate(firstWord, secondWord, thirdWord) {
-    // Write the body of this function to concatenate three words together
+  // Write the body of this function to concatenate three words together
+  return firstWord.concat(secondWord, thirdWord);
 }
 
-var result = concatenate('code', 'your', 'future');
+var result = concatenate("code", "your", "future");
 console.log(result); // expected result = "codeyourfuture"
--- a/week-1-practise/level-1/2-syntax-error-2.js
+++ b/week-1-practise/level-1/2-syntax-error-2.js
@@ -1,8 +1,8 @@
 // There is a syntax error in this code. Fix it.
 
-function addNumbers(a, b, c)
-return a + b + c;
-
+function addNumbers(a, b, c) {
+  return a + b + c;
+}
 
 var result = addNumbers(1, 3, 4);
 console.log(result); // Expected result 8 - Run the program and make sure it works
--- a/week-1-practise/level-1/3-syntax-error-3.js
+++ b/week-1-practise/level-1/3-syntax-error-3.js
@@ -1,7 +1,7 @@
 // There is a syntax error with this code. Fix it.
 
-function(a, b, c) {
-    return a + b + c;
+function addNumbers(a, b, c) {
+  return a + b + c;
 }
 
 var result = addNumbers(1, 3, 4);
--- a/week-1-practise/level-1/4-logic-error-1.js
+++ b/week-1-practise/level-1/4-logic-error-1.js
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return "CodeYourFuture";
 }
 
 var result = trimWord("  CodeYourFuture ");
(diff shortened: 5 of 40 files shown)


week-1-practise/level-1/1-syntax-error-1.js
```diff
@@ -1,14 +1,13 @@
...
 var result = addNumbers(1, 3, 4);
 console.log(result); // Expected result 8 - Run the program and make sure it works
 
-
 // Answer these questions:
 // 1. How many parameters does the function take?
 // 2. What is the function name?
-// 3. Where is the function called? with what parameters?
\ No newline at end of file
+// 3. Where is the function called? with what parameters?
```
boubacardllo: // There is a syntax error in this code. Fix it.

function addNumbers(a, b, c) {
  return a + b + c;
}

// Answer these questions:
// 3. Where is the function called? with what parameters?
/*Respose */

/* -The function take three parameters
-The function name is: addNumbers
- The is called in line 7 and (1,2,3) as parameters */

christopherallanperry: No problems with it for the most part, but you might like to check the parameters in your answer to Q3

week-1-practise/level-1/10-function-output-3.js
```diff
@@ -1,6 +1,7 @@
...
-    // Write the body of this function to concatenate three words together
+  // Write the body of this function to concatenate three words together
+  return firstWord.concat(secondWord, thirdWord);
 }
 
-var result = concatenate('code', 'your', 'future');
-console.log(result); // expected result = "codeyourfuture"
\ No newline at end of file
+var result = concatenate("code", "your", "future");
+console.log(result); // expected result = "codeyourfuture"
```
christopherallanperry: No problms again with this one

week-1-practise/level-1/2-syntax-error-2.js
```diff
@@ -1,8 +1,8 @@
...
 
-function addNumbers(a, b, c)
-return a + b + c;
-
+function addNumbers(a, b, c) {
+  return a + b + c;
+}
 
 var result = addNumbers(1, 3, 4);
 console.log(result); // Expected result 8 - Run the program and make sure it works
```
christopherallanperry: Same hre

week-1-practise/level-1/3-syntax-error-3.js
```diff
@@ -1,7 +1,7 @@
 // There is a syntax error with this code. Fix it.
 
-function(a, b, c) {
-    return a + b + c;
+function addNumbers(a, b, c) {
+  return a + b + c;
 }
 
 var result = addNumbers(1, 3, 4);
```
christopherallanperry: And again

week-1-practise/level-1/4-logic-error-1.js
```diff
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return "CodeYourFuture";
 }
 
 var result = trimWord("  CodeYourFuture ");
```
christopherallanperry: This one needs another look. Remember that functions are intended to be re-usable, so for example, think what would happen if you called the function as follows:
```
Var result = trimWord(“  it is autumn     “;
```

What would you get back from your function as it’s currently written?

Take a look at how the trim method should be used at https://developer.mozilla.org/en-US/

week-1-practise/level-1/5-logic-error-2.js
```diff
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return word.trim;
+  return "CodeYourFuture";
 }
 
 var result = trimWord("  CodeYourFuture ");
```
christopherallanperry: The same comment applies here.

week-1-practise/level-1/6-logic-error-3.js
```diff
@@ -1,10 +1,10 @@
...
 function trim(word) {
-  return "word".trim();
+  return "CodeYourFuture";
 }
 
-var result = trim("CodeYourFuture  ");
+var result = trim("CodeYourFuture");
 console.log(result);
 // Expected result "CodeYourFuture"
 // Run the program and make sure it works
```
christopherallanperry: The only error that needs to be fixed is inside the function. You don’t need to do anything to the line where the function is called. Again, back to the MDN page on trim.

week-1-practise/level-1/7-logic-error-4.js
```diff
@@ -1,8 +1,8 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+  
+  return a * b * c; 
 }
 
 var result = multiply(1, 3, 4);
```

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -11,4 +11,11 @@ console.log(result);
...
-// - How can this function be improved?
\ No newline at end of file
+// - How can this function be improved?
+/*-Response
+the programme concatenate two parameter
+the function take two parameter
+ the name is concatenate
+ the function is called in line 5, code and your
+
+the function return codeyour*/
```
christopherallanperry: After you amended the function, did you try running it to see whether it did as you expected it to? You might want to do that and keep a close eye on any errors that appear in the console.

Also, you don’t appear to have answered the final question on improving the function - it’s looking for a narrative answer, not just changes to the code itself.

PR #158, opened 2018-11-17
homework

Ibraham00:


week-1-practise/level-1/1-syntax-error-1.js
```diff
@@ -1,14 +1,16 @@
...
-
 // Answer these questions:
 // 1. How many parameters does the function take?
+// 3 parameters
 // 2. What is the function name?
-// 3. Where is the function called? with what parameters?
\ No newline at end of file
+// addNumbers()
+// 3. Where is the function called? with what parameters?
+// result , 1,3,4.
```
christopherallanperry: All good

week-1-practise/level-1/10-function-output-3.js
```diff
@@ -1,6 +1,6 @@
 function concatenate(firstWord, secondWord, thirdWord) {
-    // Write the body of this function to concatenate three words together
+  return firstWord + secondWord + thirdWord;
 }
 
-var result = concatenate('code', 'your', 'future');
-console.log(result); // expected result = "codeyourfuture"
\ No newline at end of file
+var result = concatenate("code", "your", "future");
+console.log(result); // expected result = "codeyourfuture"
```
christopherallanperry: Have a look at MDN ( https://developer.mozilla.org/en-US/) and see if you can find a JavaScript 'String' method for achieving the same thing.

week-1-practise/level-1/2-syntax-error-2.js
```diff
@@ -1,8 +1,8 @@
...
 
-function addNumbers(a, b, c)
-return a + b + c;
-
+function addNumbers(a, b, c) {
+  return a + b + c;
+}
 
 var result = addNumbers(1, 3, 4);
 console.log(result); // Expected result 8 - Run the program and make sure it works
```
christopherallanperry: Good

week-1-practise/level-1/3-syntax-error-3.js
```diff
@@ -1,7 +1,7 @@
 // There is a syntax error with this code. Fix it.
 
-function(a, b, c) {
-    return a + b + c;
+function addNumbers(a, b, c) {
+  return a + b + c;
 }
 
 var result = addNumbers(1, 3, 4);
```
christopherallanperry: Good

week-1-practise/level-1/4-logic-error-1.js
```diff
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
 
 var result = trimWord("  CodeYourFuture ");
```
christopherallanperry: Sometime the simplest or error's are the hardest to find. Well done.

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -1,14 +1,20 @@
...
 
 // Answer these questions:
 // - What does this program do? Run it and see, or Google some of the methods used ot understand them.
+// Merge two parameters or arrays together.
 // - How many parameters does the function take?
+// 2 parameters.
 // - What is the function name?
+// s
 // - Where is the function called? with what parameters?
+// result , code , your
```
christopherallanperry: Again, it's looking for the line number.

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -1,14 +1,20 @@
...
 // - What does the function return?
-// - How can this function be improved?
\ No newline at end of file
+// codeyour
+// - How can this function be improved?
```
christopherallanperry: Any thoughts on this one?

week-1-practise/level-2/1-tax.js
```diff
@@ -5,7 +5,9 @@
   Sales tax is 20% of the price of the product
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  return (20 * price) / 100;
+}
```
christopherallanperry: There are many ways of achieving the same end goal here
```
return price * 0.2;

return price / 5;
```
You could even modify the parameters the function uses to take in a value for 'tax'. That way you could store a tax rate elsewhere (a database maybe) where it could be stroed once and used many time.

week-1-practise/level-2/2-format-currency.js
```diff
@@ -6,7 +6,9 @@
   Write a function that transforms numbers into the format £0.00
 */
 
-function formatCurrency() {}
+function formatCurrency(price) {
+  return "£" + price.toFixed(2);
+}
```

week-1-practise/level-2/3-currency-conversion.js
```diff
@@ -5,7 +5,9 @@
   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
 */
 
-function convertToUSD() {}
+function convertToUSD(price) {
+  return "$" + (price * 1.4).toFixed(2);
+}
```
christopherallanperry: As per the comment on the previous question, what happens if this function is being used to convert currencies other than "pound" to $? Would the way the function is written still make sense, particularly when read by another programmer at a later date?

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -6,9 +6,15 @@
...
-function convertToBRL(price) {}
+function transactionFee(price) {
+  return price * 0.01;
+}
+function convertToUSD(price) {
+  return price * 1.4 + transactionFee(price);
+}
+function convertToBRL(price) {
+  return price * 4.85 + transactionFee(price);
+}
```
christopherallanperry: See previous comments. Going forward, do you think that it would be considered good practice to hard code specific numbers into your functions like this? What would be the consequence of the exchange rateor the transaction fee, changing for example?

Alisina92:
Hi Ibrahim, 
Try to rename your homework something meaningful like Ibrahim- Week 2 homework

PR #160, opened 2018-11-22
Week 2 and its exercises

Abdoulrazack95:


week-1-practise/level-1/1-syntax-error-1.js
```diff
@@ -1,14 +1,16 @@
...
-
 // Answer these questions:
 // 1. How many parameters does the function take?
+// The fonction takes three parameters which are a, b, c.
 // 2. What is the function name?
-// 3. Where is the function called? with what parameters?
\ No newline at end of file
+//The function's name is addNumbers.
+// 3. Where is the function called? with what parameters?
+// The function is called outside of the function e.g. : var result = addNumbers();. The parameters are called with these three agreements : 1, 3, 4;
```
christopherallanperry: Good. For the last question, you could have just quoted the line number on which the function is called.

week-1-practise/level-1/10-function-output-3.js
```diff
@@ -1,6 +1,7 @@
...
-    // Write the body of this function to concatenate three words together
+  // Write the body of this function to concatenate three words together
+  return firstWord.concat(secondWord, thirdWord);
 }
 
-var result = concatenate('code', 'your', 'future');
-console.log(result); // expected result = "codeyourfuture"
\ No newline at end of file
+var result = concatenate("code", "your", "future");
+console.log(result); // expected result = "codeyourfuture"
```
christopherallanperry: Good, no problems there.

week-1-practise/level-1/2-syntax-error-2.js
```diff
@@ -1,8 +1,8 @@
...
 
-function addNumbers(a, b, c)
-return a + b + c;
-
+function addNumbers(a, b, c) {
+  return a + b + c;
+}
 
 var result = addNumbers(1, 3, 4);
 console.log(result); // Expected result 8 - Run the program and make sure it works
```
christopherallanperry: No problems again.

week-1-practise/level-1/3-syntax-error-3.js
```diff
@@ -1,7 +1,7 @@
 // There is a syntax error with this code. Fix it.
 
-function(a, b, c) {
-    return a + b + c;
+function addNumbers(a, b, c) {
+  return a + b + c;
 }
 
 var result = addNumbers(1, 3, 4);
```
christopherallanperry: Good.

week-1-practise/level-1/4-logic-error-1.js
```diff
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
 
 var result = trimWord("  CodeYourFuture ");
```
christopherallanperry: Well spotted.

week-1-practise/level-1/5-logic-error-2.js
```diff
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return word.trim;
+  return word.trim();
 }
 
 var result = trimWord("  CodeYourFuture ");
```
christopherallanperry: Good

week-1-practise/level-1/6-logic-error-3.js
```diff
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trim(word) {
-  return "word".trim();
+  return word.trim();
 }
 
 var result = trim("CodeYourFuture  ");
```
christopherallanperry: No problems there.

week-1-practise/level-1/7-logic-error-4.js
```diff
@@ -1,8 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+  return a * b * c;
 }
 
 var result = multiply(1, 3, 4);
```

week-1-practise/level-1/8-function-output-1.js
```diff
@@ -1,12 +1,16 @@
...
-    return Math.random() * 10;
+  return Math.random() * 10;
 }
 
 var result = getNumber();
 console.log(result);
 
 // Answer these questions:
 // - What does this program do?
+// Math.random() returns a random number in the range 0 and 10.
```
christopherallanperry: Check out the MDN (https://developer.mozilla.org/en-US/) page for `Math.random` - does it really return a value that would allow 10 to be the result of this function?

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -1,14 +1,20 @@
...
 // - What is the function name?
+// the function name is s.
 // - Where is the function called? with what parameters?
+// The function is called at line 5 which is s, and the parameters are called with the following arguments : code, your, future.
 // - What does the function return?
-// - How can this function be improved?
\ No newline at end of file
+// the function return "codeyour".
+// - How can this function be improved?
+// this function can be improved strings with space inside the parenthese of concat.
```
christopherallanperry: You could also improve the names of the variables to make it easier to read. Otherwise, good answers.

week-1-practise/level-2/1-tax.js
```diff
@@ -5,7 +5,9 @@
   Sales tax is 20% of the price of the product
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  return price * (20 / 100);
+}
```
christopherallanperry: There are many ways of achieving the same end goal here
```
return price * 0.2;

return price / 5;
```

You could even modify the parameters the function uses to take in a value for 'tax'. That way you could store a tax rate elsewhere (a database maybe) where it could be stored once and used many time.

week-1-practise/level-2/2-format-currency.js
```diff
@@ -6,7 +6,9 @@
   Write a function that transforms numbers into the format £0.00
 */
 
-function formatCurrency() {}
+function formatCurrency(price) {
+  return "£" + price.toFixed(2);
+}
 
 /* DO NOT EDIT BELOW THIS LINE */
 /* --------------------------- */
```

week-1-practise/level-2/3-currency-conversion.js
```diff
@@ -5,7 +5,9 @@
   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
 */
 
-function convertToUSD() {}
+function convertToUSD(priceInGBP) {
+  return price * (1.29631).toFixed(2);
+}
```
christopherallanperry: I'm curious as to why you've answered the question as you have - surely multiplying `price` by 1.4 would have got you the result you wanted. Also, did you test your solution before posting it? Did it give you the answer you expected?

Lastly, as per the comment on the previous question, what happens if this function is being used to convert currencies other than "pound" to $? Would the way the function is written still make sense, particularly when read by another programmer at a later date?

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,8 +7,12 @@
...
 */
 
-function convertToUSD(price) {}
-function convertToBRL(price) {}
+function convertToUSD(priceInGBP) {
+  return price * 1.3 * 1.01;
+}
+function convertToBRL(priceInGBP) {
+  return price * 3.78 * 1.01;
+}
```
christopherallanperry: Whilst the functions work as intended, it would be worth you looking at the part of the question that asks you to "Find a way to add 1% to all currency conversions (think about the DRY principle)" - have you met that requirement, or does it need a little more thought?

PR #161, opened 2018-11-24
Update first and last array get/set test

saralk:
I saw a few people hard coding the number 2 into their code to get this example to work. This fixes that.

* Fix a code error in the test
* Test two different lengths of array for the last element example, in order to make it not possible to hard code a value


40thieves:
This has conflicts now. Closing in favour of https://github.com/CodeYourFuture/js-exercises/pull/378

40thieves:
Thanks for the contribution!

PR #166, opened 2018-12-16
Ivina - JS exercises

ivinapontes:


week-1/G-strings-methods/exercise.js
```diff
@@ -1,3 +1,5 @@
 // Start by creating a variable `message`
-
-console.log(message);
+var myName = "Ivina";
+console.log(
+  " My name is " + myName + " and my name has " + myName.length + " letters"
```
40thieves: Good usage of string concatenation 👍

week-1/G-strings-methods/exercise2.js
```diff
@@ -1,3 +1,8 @@
-var name = " Daniel  ";
+var name = " Daniel  ".trim();
+
+var message =
+  " My name is" + name + "and my name has " + name.length + " letters";
+
+console.log(name.trim());
```
40thieves: Do you think using `.trim()` a second time here is necessary? What happens if we remove it? Can you explain why?

week-1/H-numbers/exercise.js
```diff
@@ -1 +1,8 @@
 // Start by creating a variables `numberOfStudents` and `numberOfMentors`
+var numberOfStudents = 26;
+var numberOfMentors = 15;
+var total = numberOfMentors + numberOfStudents;
```
40thieves: Nice usage of a variable here. And it's well named 👏

week-1/I-floats/exercise.js
```diff
@@ -1,2 +1,11 @@
-var numberOfStudents = 15;
-var numberOfMentors = 8;
+function percentage(numberOfMentors, numberOfStudents) {
```
40thieves: Excellent usage of a function! Well done 💪 

Do you think this is the best name for the function? Can you think of a better one?

week-1/K-functions-parameters/exercise4.js
```diff
@@ -1,5 +1,8 @@
 // Declare your function first
-
+function adding(a, b) {
```
40thieves: This looks good, but how could you name the function a bit better to make it clearer. Normally we name functions as verbs, e.g. `create`, `jump`, `doTask`

week-1/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,34 @@
...
+
 var mentor1 = "Daniel";
 var mentor2 = "Irina";
 var mentor3 = "Mimi";
 var mentor4 = "Rob";
 var mentor5 = "Yohannes";
+
+//Return names in uppercase
+function greetingMentor(mentor) {
+  console.log("HELLO " + mentor.toUpperCase());
```
40thieves: Nice work! 🏅

week-1/I-floats/exercise.js
clems27: @ivinapontes Nice work

week-1/I-floats/exercise.js
```diff
@@ -1,2 +1,11 @@
...
+  var total = numberOfMentors + numberOfStudents;
+
+  var percentageStudents = Math.round((numberOfStudents * 100) / total);
+  var percentageMentors = Math.round((numberOfMentors * 100) / total);
+
+  console.log("Percentage Students: " + percentageStudents + "%");
```
Atayo: I like that you have included the percentage sign "%"

ivinapontes: Thank you for the feedback Tayo @Atayo! :smile: I appreciate it!

week-1/L-functions-nested/exercise.js
ivinapontes: Yay! Thank you! :smile_cat:

week-1/I-floats/exercise.js
ivinapontes: Thank you for the feedback @clems27 !

ivinapontes: Do you think that naming this function getPercentage or getPercent for short would be better? Thank you for your feedback, Ali @40thieves!  I appreciate it!

week-1/G-strings-methods/exercise2.js
ivinapontes: That's great that you saw this. I used the method twice because I was trying to see if it would work the same way. But then I found out a better and clearer way to solve it by updating the variable instead of defining with the method inside. What do you think? Thank you!

40thieves: Nice fix 👍

week-1/I-floats/exercise.js
40thieves: I think `getPercentages` is a good name, but we usually try to name functions after the core of what they do. In this case the function ultimately has 1 thing that it does at the end. It prints out the percentages, so do you think a name would be `printPercentages`?

40thieves: `getPercent` is a reasonable name, but in my experience, most of the times I've tried to shorten a variable/function name it has been worse for readability in the long term.

week-1-practise/level-1/10-function-output-3.js
```diff
@@ -1,6 +1,7 @@
 function concatenate(firstWord, secondWord, thirdWord) {
-    // Write the body of this function to concatenate three words together
+  // Write the body of this function to concatenate three words together
+   return firstWord.concat(secondWord, thirdWord);
```
40thieves: Ha, I actually had forgotten that there was a `concat` function on strings! You've taught me something 😄 

Is there another style of string concatenation that you could use?

week-1-practise/level-1/3-syntax-error-3.js
```diff
@@ -1,8 +1,11 @@
...
+function addNumbers(a, b, c) {
+  return a + b + c;
 }
 
 var result = addNumbers(1, 3, 4);
-console.log(result); // Expected result 8 - Run the program and make sure it works
+console.log(result);
+// Expected result 8 - Run the program and make sure it works:
+
+//It was missing the function name
```
40thieves: 👍

week-1-practise/level-1/7-logic-error-4.js
```diff
@@ -1,8 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+  return a * b * c;
```
40thieves: Great!

week-1-practise/level-1/8-function-output-1.js
```diff
@@ -1,12 +1,13 @@
 function getNumber() {
-    return Math.random() * 10;
+    return (Math.random() * 10);
```
40thieves: You aren't wrong, but the extra parentheses (`(` and `)`) here are kind of unnecessary.

Sometimes when we're doing maths with our code we want to add extra parentheses to make it easier to read, but in this case, the extra parens don't really add anything.

week-1-practise/level-1/9-function-output-2.js
```diff
@@ -5,10 +5,28 @@ function s(w1, w2) {
 var result = s('code', 'your');
 console.log(result);
 
+/* --------------- Function improved-----------------------
```
40thieves: Nice improvement 👏 And I like the use of the block comment too 🙂

week-1-practise/level-2/3-currency-conversion.js
```diff
@@ -5,8 +5,10 @@
   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
 */
 
-function convertToUSD() {}
-
+function convertToUSD(price) {
+  price = "$" + (price * 1.4).toFixed(2);
```
40thieves: If I was being **very** picky, then I would say that in general we don't reassign a function parameter to have a different value. Most of the time we make a new variable, just to make it clear that we're changing it. This also means that we can still access the value of the parameter lower in the function if we need to.

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,8 +7,17 @@
...
-function convertToUSD(price) {}
-function convertToBRL(price) {}
+function convertToUSD(price) {
+  price = "$ " + (price * 1.4).toFixed(2);
+  return price;
+}
+
+function convertToBRL(price) {
+  price = price * 4.92;
+  var feeForTransactions = (price * 1) / 100;
```
40thieves: Well named variable 🍰

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,8 +7,17 @@
   Find a way to add 1% to all currency conversions (think about the DRY principle)
 */
 
-function convertToUSD(price) {}
-function convertToBRL(price) {}
+function convertToUSD(price) {
+  price = "$ " + (price * 1.4).toFixed(2);
```
40thieves: If you read the exercise description carefully, it says that the business is adding a 1% fee to **all** foreign transactions, not just Brazilian ones. How could you change your `convertToUSD` function to handle this?

Once you've done that, have a look at both of your functions. Do they have similar lines? How could we change them to reduce this? (May be google the "DRY principle")

week-1-practise/level-3/1-piping.js
```diff
@@ -11,6 +11,10 @@
 */
 
 var startingValue = 5;
+startingValue = multiply(startingValue, 2);
+startingValue = add(startingValue, 10);
+startingValue = format(startingValue);
+var result = startingValue;
```
40thieves: I think my advice about not reassigning function parameters applies here too. If we wanted to `console.log` the starting value (before it changes), we can't now. How could we change this to keep the `startingValue` the same?

40thieves: > As a learning exercise (you wouldn't do this normally), do all of the above in 1 line of code 

I think you missed this exercise? Can you combine all of the function calls into a single line? (I would put that as a separate bit of code)

week-1/K-functions-parameters/exercise5.js
```diff
@@ -3,6 +3,6 @@ function createLongGreeting(name, age) {
   return "Hello, my name is " + name + " and I'm " + age + " years old.";
 }
 
-const greeting = createLongGreeting("Daniel", 30);
+var greeting = createLongGreeting("Daniel", 30);
```
40thieves: This isn't wrong, and you are correct that we are just looking at the `var` keyword in class at the moment. In the future classes we will cover the `const` and `let` keywords. So you can leave it as it is if you want.

week-1-practise/level-1/10-function-output-3.js
ivinapontes: I could use just the plus sign to concatenate all the arguments, it's more efficient than using _concat method_ :smile:

40thieves: I think they are probably equally efficient in terms of how long it takes the computer to run them. But yes, it is slightly easier to type. Either way is good though!

week-1-practise/level-2/3-currency-conversion.js
ivinapontes: Thank you for pointing this out. I reassigned according to each currency. :dollar: :bank:

week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -6,9 +6,40 @@
   They have also decided that they should add a 1% fee to all foreign transactions
   Find a way to add 1% to all currency conversions (think about the DRY principle)
 */
+function convertToUSD(price) {
+  usPrice = price * 1.4;
```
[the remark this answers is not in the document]
ivinapontes: Oh! I think it's missing the var or let keyword, is it correct?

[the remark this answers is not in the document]
40thieves: Yep that's right. Can you fix it?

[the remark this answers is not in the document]
40thieves: Perfect!

PR #180, opened 2018-12-16
Ismail JS Exercises A-L

IsmailCtg:
It was cool

### DIFF
--- a/week-1/E-strings/exercise.js
+++ b/week-1/E-strings/exercise.js
@@ -1,4 +1,5 @@
 // Start by creating a variable `message`
-var message = 'Hello. This is a string.';
-console.log(message_type);
-
+var message = 'This is a string.';
+var messageType = typeof message;
+console.log(message);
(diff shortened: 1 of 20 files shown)


week-1/E-strings/exercise.js
```diff
@@ -1,3 +1,4 @@
 // Start by creating a variable `message`
+var message = 'Hello. This is a string.';
+console.log(message_type);
```
AlexanderBraithwaite: Try 'typeof' to get the data type of the variable message

week-1/G-strings-methods/exercise.js
```diff
@@ -1,3 +1,5 @@
 // Start by creating a variable `message`
+var name = "Daniel";
+var nameLength = name.length;
 
 console.log(message);
```
AlexanderBraithwaite: Check that you're console logging the correct variable 👍

PR #182, opened 2018-12-16
Mussie class exercise

Mussie14:


khan-zareen:
nice work

week-1/K-functions-parameters/exercise.js
```diff
@@ -1,9 +1,9 @@
 // Complete the function so that it takes input parameters
-function multiply() {
-  // Calculate the result of the function and return it
+function multiply(a, b) {
+  return a * b;
 }
 
 // Assign the result of calling the function the variable `result`
-var result = multiply(3, 4);
+var result = multiply((a = 3), (b = 4));
```
laperla11: hi Mussie, your function definitions is good, however but "var result = multiply((a = 3), (b = 4));" should not work.
instead you need to write "var result = multiply(3, 4); console.log(result);"

week-1-practise/level-1/4-logic-error-1.js
```diff
@@ -1,7 +1,7 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
 
 var result = trimWord("  CodeYourFuture ");
```
bmc82: Nice work, Mussie.

theseafaringturtle:
Good job. Just one nitpick: in part L you forgot to return/print the message. By the way, 65 and 35 are the expected results rather than the function parameters, but they're outdated anyway since the students/mentors are no longer 20/11.

week-1-practise/level-1/4-logic-error-1.js
Mussie14: thank you bart.

week-1-practise/level-2/2-format-currency.js
```diff
@@ -6,7 +6,9 @@
   Write a function that transforms numbers into the format £0.00
 */
 
-function formatCurrency() {}
+function formatCurrency(price) {
+  return "£" + price.toFixed(2);
```
clems27: Nice work with your function

week-1/E-strings/exercise.js
```diff
@@ -1,3 +1,7 @@
 // Start by creating a variable `message`
 
-console.log(message);
+var message = "you can do this";
+var messageType = typeof message;
```
clems27: Nice line of code

week-1/K-functions-parameters/exercise2.js
```diff
@@ -1,5 +1,7 @@
-// Declare your function first
+function divide(a, b) {
+  return a / b;
+}
 
-var result = divide(3, 4);
+var result = divide((a = 3), (b = 4));
```
LucyMac: This line was fine, not need to change it. When we call the function by saying `divide(3, 4)`, the program knows that `3` is `a` and `4` is `b` because you defined it that way on line 1.

week-1/K-functions-parameters/exercise4.js
```diff
@@ -1,5 +1,7 @@
-// Declare your function first
+function add(a, b) {
+  return a + b;
+}
 
-// Call the function and assign to a variable `sum`
+var sum = add((a = 13), (b = 124));
```
LucyMac: Same here: your function definition on lines 1-3 is perfect, however line 5 is incorrect. When you call a function, you do not need to say what value corresponds to which parameter: javascript just follows the order that you provide your arguments in and knows which one is which 🔮

PR #191, opened 2018-12-22
Ermi-javascript exercises

ermiii:


week-1/G-strings-methods/exercise2.js
```diff
@@ -1,3 +1,4 @@
-var name = " Daniel  ";
+var name = " Ermiyas!  ";
+var message= name + "Hello mates";
 
 console.log(message);
```
martynamaron: Nice!
But when you run this exercise, it prints: " Ermiyas!  Hello mates". How can you change it to say something more meaninfgul? More of a greeting or introducing yourself? :)

week-1/H-numbers/exercise.js
```diff
@@ -1 +1,4 @@
 // Start by creating a variables `numberOfStudents` and `numberOfMentors`
+var numberOfStudents=26;
+var numberOfMentors=15;
+console.log(numberOfStudents+numberOfMentors);
```
martynamaron: Thast prints the total number, yes, great!
But the instructions say that this should be the output:
```Number of students: 15
Number of mentors: 8
Total number of students and mentors: 23```

So how can you fix it?

File: week-1/K-functions-parameters/exercise3.js [hunk 3]
```diff
@@ -1,5 +1,8 @@
 // Write your function here
+function createGreeting(){
+return "Hello! my name is"
 
-var greeting = createGreeting("Daniel");
+   
+var greeting = createGreeting + "Daniel";}
```
martynamaron: Having the `}` here at the end means that the `greeting` variable belongs to the function - and it shouldn't.  In fact, are you sure your code runs here? Visual Studio should underline your code in red if it thinks there are errors.

Rethink what's happening here :) Remember that `createGreeting` is a function, not a variable and that it should take a name (a string) as parameter and return a greeting

week-1/K-functions-parameters/exercise4.js
```diff
@@ -1,4 +1,9 @@
 // Declare your function first
+function add (num1,num2){
+    return num1 + num2;
+
+}
+sum= 13+124;
```
martynamaron: Remember the proper way to create variables - this isn't quite right.
Also, please use the `add` function here instead of adding the numbers directly

week-1/K-functions-parameters/exercise5.js
```diff
@@ -1,5 +1,8 @@
 // Declare your function here
+function createLongGreeting(){
+    return "my name is";
 
-const greeting = createLongGreeting("Daniel", 30);
+
+const greeting = createLongGreeting("Daniel", "i am " + "30");}
```
martynamaron: Can you double check if this code runs? I think it doesn't - can you tell me why?

PR #205, opened 2019-01-09
BartC Week 2

bmc82:


week-2/F-logical-operators/exercise2.js
```diff
@@ -5,17 +5,31 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
-
+function isNegative(number) {
+  if (number < 0) return true;
```
40thieves: This is technically correct, but 99% of the time when we are using `if` statements we also want to use curly braces (`{` and `}`).

week-2/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return arr[0]; // complete this statement
 }
 
 function last(arr) {
-  return; // complete this statement
+  return arr[arr.length - 1]; // complete this statement
```
40thieves: Well done! 👏

week-2/L-array-methods-2/exercise3.js
```diff
@@ -7,15 +7,15 @@
 var ukNations = ["Scotland", "Wales", "England", "Northern Ireland"];
 
 function isInUK(country) {
-  return; // complete this statement
+  if (ukNations.includes(country)) return true;
+  else return false; // complete this statement
```
40thieves: This is correct but there is some unnecessary code here.

What is the value of `ukNations.includes(country)`? Is it a boolean? If it is, do need to write out `return true` or ` return false`? Or can we just return the boolean?

week-2/M-array-map/exercise.js
```diff
@@ -10,7 +10,10 @@ function double(num) {
 }
 
 var numbers = [1, 2, 3, 4];
-var numbersDoubled; // complete this statement (use map and the double function)
+var numbersDoubled = numbers.map(function(num) {
+  return num * 2;
```
40thieves: Correct! Well done! 🍰

week-2-practise/level-1/6-map.js
```diff
@@ -11,7 +11,7 @@ function tidyUpString(str) {
 }
 
 var mentors = ["/Daniel ", "irina ", " Gordon", "ashleigh "];
-var mentorsTidy; // ONLY EDIT THIS LINE
+var mentorsTidy = mentors.map(tidyUpString); // ONLY EDIT THIS LINE
```
40thieves: Nice work 👍

week-2-practise/level-2/2-validation.js
```diff
@@ -4,8 +4,20 @@
 // - is less than or equal to 100
 // Tip: write other small functions for each requirement
 
-function validate(num) {}
-
+function validate(num) {
+  return typeof num === "number";
+}
+function validate(num) {
+  if (num > 0) return true;
```
40thieves: When using an `if` statement we usually use curly braces (`{` and `}`)

week-2-practise/level-2/4-slice.js
```diff
@@ -5,7 +5,7 @@
 // - remove the item at the specified index
 
 function remove(arr, index) {
-  return; // complete this statement
+  return arr.slice(0, index).concat(arr.slice(index + 1)); // complete this statement
```
40thieves: Great work 🎉

week-2-practise/level-2/6-map.js
```diff
@@ -3,8 +3,13 @@
 // 2. Declare a new array, `percentagesFormatted`, containing
 //    - each item in `percentages` formatted by your function
 
+function formatPercentage(number){
+  return number+="%";
```
40thieves: The `.map()` method creates a **new** array from the old array transforming it based on the callback. That means we don't need to update the `number` variable here. We are only using it to calculate the new string value in the new array.

That means the `+=` operator is unnecessary. You can just use the `+` operator.

PR #209, opened 2019-01-10
Mussie week 2 exercises

Mussie14:


martynamaron:
Hi @Mussie14, well done for opening a PR with just the week-2 exercises! Please add your homework here so it can be reviewed before Sunday! thanks :)

week-2/G-conditionals-2/exercise-2.js
```diff
@@ -8,7 +8,12 @@
 */
 
 function studentPassed(grade) {
-
+  if (grade < 50) {
+    return "fail";
+  }
+  if (grade >= 50) {
```
40thieves: Could we make this code a little shorter? What if we try using a `else` statement?

week-2/G-conditionals-2/exercise-3.js
```diff
@@ -9,7 +9,10 @@
 */
 
 function calculateGrade(mark) {
-
+  if (mark > 80) return "A";
+  else if (mark >= 60 && mark <= 80) return "B";
```
40thieves: This is correct! 🎉 

But there are a few improvements we could make here. Usually when we use an `if` statement we also want to use curly braces (`{` or `}`).

The other thing is that we can make our `else if` statement a little shorter. You're checking that `mark` is greater than 80 first, so that means that if it reaches the `else if` then `mark` cannot be greater than 80.

That means that checking that `mark` is less than 80 in the `else if` is unnecessary.

week-2/G-conditionals-2/exercise-4.js
```diff
@@ -9,7 +9,7 @@
 */
 
 function containsCode(sentence) {
-
+  return sentence.includes("code");
```
40thieves: Nice work!

week-2/K-array-methods/exercise2.js
```diff
@@ -7,7 +7,7 @@
 var mentors = ["Daniel", "Irina", "Rares"];
 var students = ["Rukmini", "Abdul", "Austine", "Swathi"];
 
-var everyone; // complete this statement
+var everyone = mentors.concat(students); // complete this statement
```
40thieves: 🎉 Correct!

week-2/G-conditionals-2/exercise-2.js
NajlaSoltana: hi Mussie  I find the G exercise a bit confusing in terms there are 4 exercises in G conditionals. you could have pushed the 2 ones separately. that make them clearer to follow. but overall well done ;)

PR #210, opened 2019-01-10
Elamin- js-week2-exercises-new 

aldeez988:


week-2/D-predicates/exercise.js
```diff
@@ -7,12 +7,14 @@
...
 function isNegative(number) {
-
+  if (number < 0) return true;
+  else return false;
 }
 
 // Finish the predicate function to test if the passed number is between 0 and 10
 function isBetweenZeroAnd10(number) {
-
+  if (number <= 10 && number >= 0) return true;
```
clems27: Nice function declaration and good use of logical operators, well done

week-2-practise/level-1/6-map.js
```diff
@@ -11,7 +11,7 @@ function tidyUpString(str) {
 }
 
 var mentors = ["/Daniel ", "irina ", " Gordon", "ashleigh "];
-var mentorsTidy; // ONLY EDIT THIS LINE
+var mentorsTidy = mentors.map(tidyUpString); // ONLY EDIT THIS LINE
```
laperla11: Nice use of callback functions with the .map () method.

week-2-practise/level-1/2-validation.js
```diff
@@ -2,7 +2,7 @@
 // TIP: You should write an expression that returns a boolean value
 
 var num = 10;
-var isBigEnough; // ONLY EDIT THIS LINE
+var isBigEnough = true; // ONLY EDIT THIS LINE
```
theseafaringturtle: Hi, the expression is supposed to be num >= 10 rather than true so that the If would be correctly evaluated even when num is not 10

week-2-practise/level-2/4-slice.js
```diff
@@ -5,7 +5,7 @@
 // - remove the item at the specified index
 
 function remove(arr, index) {
-  return; // complete this statement
+  return arr.slice(0, index).concat(arr.slice(index + 1));
```
theseafaringturtle: It's a very good independently achieved solution, although array.splice() would have been easier

week-2-practise/level-2/5-indexOf.js
```diff
@@ -5,7 +5,7 @@
 // TIP: Use the .indexOf() method
 
 function remove(arr, valueToRemove) {
-  return; // complete this statement
+  return arr.slice(arr.indexOf(valueToRemove), arr.indexOf(valueToRemove) + 1); // complete this statement
 }
```
theseafaringturtle: - you should use a similar solution from 4-slice to remove the item, this one returns the item
- keep the index as separate variable for readability
- don't forget to make a copy of the array using array.slice(0) so as to not modify the original array

week-2-practise/level-3/5-indexOf.js
```diff
@@ -6,7 +6,8 @@
 // - return a new array with the replacement value inserted
 
 function replace(arr, valueToReplace, newValue) {
-  return; // complete this statement
+  index = arr.indexOf(valueToReplace);
+  return arr.slice(0, index).concat(newValue, arr.slice(index + 1)); // complete this statement
```
theseafaringturtle: Very good that you've kept the index as a variable here and modified the previous solution

week-2-practise/level-1/2-validation.js
theseafaringturtle: Hi, the expression should be num >= 10 rather than true so that it would still work even if num wasn't 10

week-2-practise/level-2/4-slice.js
theseafaringturtle: Very good independently achieved solution, although i recommend looking into array.splice() since it's easier for this task.

week-2-practise/level-2/5-indexOf.js
```diff
@@ -5,7 +5,7 @@
 // TIP: Use the .indexOf() method
 
 function remove(arr, valueToRemove) {
-  return; // complete this statement
+  return arr.slice(arr.indexOf(valueToRemove), arr.indexOf(valueToRemove) + 1); // complete this statement
```
theseafaringturtle: - you should reuse the solution from 4-slice to remove the item, this one returns the item
- best keep the index as separate variable for readability
- don't forget to make a copy of the array using array.slice(0) so as to not modify the original array

week-2-practise/level-3/5-indexOf.js
theseafaringturtle: Nice that you've kept the index as a separate variable and modified your previous solution for the slices.

week-2-practise/level-1/2-validation.js
aldeez988: Thanks @theseafaringturtle  for your helpful comments for the return statement I do write it in separate lines but  because I use Visual Studio editor when I save the file It puts the return together with the "If" statement I think this because I have installed Prettier on it.

week-2-practise/level-3/5-indexOf.js
khan-zareen: Good use of variable

PR #211, opened 2019-01-10
Ívina - week 2 

ivinapontes:


theseafaringturtle:
Impressive work so far, accurate code and good naming.

theseafaringturtle:
Just noticed that in level-2/5-indexOf.js your remove method changes the original array. A good way to create a copy of the original would be to use array.slice(0)

week-2/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount >= mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = studentCount <= roomMaxCapacity ; // finish this statement
```
Atayo: Great work Ivina.  I think this is to find out if there is enough space in room for students and mentors(not just the students).

PR #212, opened 2019-01-10
Tayo-week2

Atayo:


Atayo:
For level-2/4-slice, I researched the internet for the solution. I do not completely understand it but I will continue practising and seeking for help when needed.

week-2-practise/level-1/6-map.js
```diff
@@ -11,7 +11,7 @@ function tidyUpString(str) {
 }
 
 var mentors = ["/Daniel ", "irina ", " Gordon", "ashleigh "];
-var mentorsTidy; // ONLY EDIT THIS LINE
+var mentorsTidy = mentors.map(tidyUpString); // ONLY EDIT THIS LINE
```
LucyMac: Unfortunately .map() is a bit more complicated to use - you can't just pass it a function. Have a look here for the syntax https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#Mapping_an_array_of_numbers_using_a_function_containing_an_argument

Mapping is like looping over an array, meaning that you perform the same action on each individual item in an array. In order to do that you need to 
1) choose a name to reference each item (you can choose anything, example:`item` or `mentor` or `x`),
2) tell the .map what name you've chosen to call each one of its items (here we've chosen the name 'item')
```
mentors.map(function(item) {
 //
}
```
3) inside the map(), say what you want to do to each item - and don't forget the parameters! Here, `tidyUpString` expects a parameter, which will be each item in your mentors array. 
```
mentors.map(function(item) {
   return tidyUpString(item));
}
```
This way, `tidyUpString` will be performed on each element in the mentors array, one after the other. The result will be a new array, saved under the variable `mentorsTidy`.

We will go over this in class on Sunday :)

week-2-practise/level-2/2-validation.js
```diff
@@ -4,7 +4,26 @@
...
+}
+
+function isPositive(num) {
+  if (num >= 0) return true;
+  else return false;
+}
+
+function isLessThanOrEqual_100(num) {
+  if (num <= 100) return true;
+  else return false;
```
LucyMac: Good logic, but you need to follow the `if` parenthesis with curly brackets - always! Same with `else`.

So it should be:
```
if (num <= 100) { 
   return true 
} else {
   return false
};
```

week-2-practise/level-2/6-map.js
```diff
@@ -4,6 +4,11 @@
 //    - each item in `percentages` formatted by your function
 
 var percentages = [1, 23, 92, 18];
+function percentage(number) {
+  return number + "%";
+}
+
+var percentagesFormatted = percentages.map(percentage);
```
LucyMac: Same comment on the map() syntax. This will not work.

week-2-practise/level-3/1-boolean.js
```diff
@@ -3,7 +3,7 @@
 // What was wrong with the original code?
 // Leave comments above your changes to explain
 
-var isHappy = "false";
+var isHappy = false; //false is a Boolean value and should not be written as a string.
```
LucyMac: 👍

week-2-practise/level-1/2-validation.js
```diff
@@ -2,7 +2,7 @@
 // TIP: You should write an expression that returns a boolean value
 
 var num = 10;
-var isBigEnough; // ONLY EDIT THIS LINE
+var isBigEnough = num; // ONLY EDIT THIS LINE
```
40thieves: I don't think this is quite correct. The exercise is asking you to calculate if the `num` variable is bigger or equal than 10. If you change the `num` variable to 5 it will still log "num is bigger than or equal to 10". How could you fix that?

week-2-practise/level-3/2-validation.js
```diff
@@ -6,7 +6,22 @@
...
+  else return false;
+}
+function isInteger(num) {
+  if (num % 1 === 0) return true;
+  else return false;
+}
+function notExcludedNums(num) {
+  if (!excludedNums.includes(num)) return true;
+  else return false;
+}
```
LucyMac: Remember the `if (condition) { outcome }` syntax using the curly brackets.

week-2/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = studentCount + mentorCount <= roomMaxCapacity; // finish this statement
```
LucyMac: Nice

week-2-practise/level-2/2-validation.js
```diff
@@ -4,7 +4,26 @@
 // - is less than or equal to 100
 // Tip: write other small functions for each requirement
 
-function validate(num) {}
+function validate(num) {
+  if (isNumber(num) && isPositive(num) && isLessThanOrEqual_100(num))
+    return true;
+  else return false;
```
40thieves: This does technically work, but 99% of the time we want to include the curly braces when using an `if` or `else` statement.

week-2/F-logical-operators/exercise.js
```diff
@@ -11,14 +11,14 @@ var cssLevel = 4;
...
+var htmlLevelAbove5 = htmlLevel > 5;
+var cssLevelAbove5 = cssLevel > 5;
 
 // Finish the next two statement
 // Use the previous variables and logical operators
 // Do not "hardcode" the answers
-var cssAndHtmlAbove5;
-var cssOrHtmlAbove5;
+var cssAndHtmlAbove5 = htmlLevelAbove5 && cssLevelAbove5;
+var cssOrHtmlAbove5 = htmlLevelAbove5 || cssLevelAbove5;
```
LucyMac: Good

week-2-practise/level-2/2-validation.js
```diff
@@ -4,7 +4,26 @@
 // - is less than or equal to 100
 // Tip: write other small functions for each requirement
 
-function validate(num) {}
+function validate(num) {
+  if (isNumber(num) && isPositive(num) && isLessThanOrEqual_100(num))
```
40thieves: Well done! This is a great use of small, well-named functions 👏

week-2/F-logical-operators/exercise2.js
```diff
@@ -5,17 +5,31 @@
...
+  else return false;
+}
+function isLongName(name) {
+  if (name.length >= 6) return true;
+  else return false;
+}
+function startsWithD(name) {
+  if (name[0] === "D") return true;
+  else return false;
+}
```
LucyMac: `if` / `else` syntax issue as above

week-2-practise/level-2/5-indexOf.js
```diff
@@ -5,7 +5,9 @@
 // TIP: Use the .indexOf() method
 
 function remove(arr, valueToRemove) {
-  return; // complete this statement
+  var removedValue = arr.indexOf(valueToRemove);
+  arr.splice(removedValue, 1);
```
40thieves: Nice usage of the `.splice` method

week-2/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return arr[0]; // complete this statement
 }
 
 function last(arr) {
-  return; // complete this statement
+  return arr[2]; // complete this statement
```
LucyMac: Not quite. This will return the 3rd character of the string you pass to it. How can you ensure you always know the index of the last character in a string? Hint: you'll need to know how many characters are in the string...

week-2/L-array-methods-2/exercise2.js
```diff
@@ -7,7 +7,11 @@
   Tip: use the string method .split() and the array method .join()
 */
 
-function capitalise(str) {}
+function capitalise(str) {
+  str = str.split("");
+  str[0] = str[0].toUpperCase();
+  return str.join("");
+}
```
LucyMac: Nice

week-2-practise/level-3/4-slice.js
```diff
@@ -5,7 +5,7 @@
 // - insert the replacement value at the provided index
 
 function replace(arr, index, value) {
-  return; // complete this statement
+  return arr.slice(0, index).concat(value, arr.slice(index + 1)); // complete this statement
```
40thieves: Well done. This is a nice usage of the `.concat()` method

week-2/M-array-map/exercise.js
```diff
@@ -10,7 +10,7 @@ function double(num) {
 }
 
 var numbers = [1, 2, 3, 4];
-var numbersDoubled; // complete this statement (use map and the double function)
+var numbersDoubled = numbers.map(double(num)); // complete this statement (use map and the double function)
```
LucyMac: Getting closer...

40thieves:
Oh it seems me @LucyMac we reviewing at the same time!

week-2-practise/level-3/6-map.js
```diff
@@ -11,11 +11,13 @@ function tidyUpString(str) {
 }
 
 function captialise(str) {
-  // complete this function
+  str = str.split("");
+  str[0] = str[0].toUpperCase();
```
mtbdata711: Hey! This is how I wish I'd done the splitting! Very clean , very quick, very simple. 

Were you able to get the code to execute both the tidyUpString and capitalise function? I can't see tidyUp being called?

week-2-practise/level-2/2-validation.js
```diff
@@ -5,24 +5,35 @@
 // Tip: write other small functions for each requirement
 
 function validate(num) {
-  if (isNumber(num) && isPositive(num) && isLessThanOrEqual_100(num))
+  if (isNumber(num) && isPositive(num) && isLessThanOrEqual_100(num)) {
```
Atayo: updated version

week-2-practise/level-2/2-validation.js
Atayo: Thank you Ali, this has now been updated

week-2-practise/level-2/2-validation.js
Atayo: Thank you Lucy, this has now been updated

PR #216, opened 2019-01-11
Fix JS1 week2 exercises

martynamaron:


40thieves:
I'm going to go ahead and merge this so it doesn't get missed. But @djgrant if you want to make some changes then go ahead.

PR #217, opened 2019-01-11
JS1 week 2 readme array method fix

martynamaron:
Reflects changes made here: https://github.com/CodeYourFuture/syllabus/pull/264/

### DIFF
--- a/week-2/K-array-methods/README.md
+++ b/week-2/K-array-methods/README.md
@@ -12,16 +12,22 @@ Arrays also have several methods that you can use.
 
 ### `.sort()`
 
-_An array method that sorts the values in an array into ascending alphabetical order._
+_An array method that sorts the values in an array into ascending alphabetical or numerical order._
 
 ```js
 var unorderedLetters = ["z", "v", "b", "f", "g"];
 var orderedLetters = unorderedLetters.sort();
 
+var unorderedNumbers = [8, 5, 1, 4, 2];
+var orderedNumbers = unorderedNumbers.sort();
+
 console.log(orderedLetters); // logs [ 'b', 'f', 'g', 'v', 'z' ]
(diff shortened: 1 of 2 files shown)


40thieves:
See comment on the other review: https://github.com/CodeYourFuture/syllabus/pull/264#discussion_r247270228

40thieves:
Sorry! Clicked the wrong button!

djgrant:
Sorry for the late review. `.sort` is a bit of a pain in JS because it doesn't actually sort numerically. For example `[8, 50, 10, 40, 20].sort()` returns `[ 10, 20, 40, 50, 8 ]`. 

To sort numbers you need to use a callback: `[8, 50, 10, 40, 20].sort((a, b) => a > b)`.

We realised this in the first using js-exercises class and just removed numbers from the exercise, probably because it was too early to be introducing callbacks.

martynamaron:
Hey @djgrant thanks! I didn't realise that, sorry... But aren't there callbacks used in `map()`? There's also this exercise that requires the use of a callback in `sort()` - https://github.com/CodeYourFuture/js-exercises/blob/master/week-2-practise/level-3/3-sort.js

40thieves:
D'oh 🤦‍♂️ I totally forgot about that behaviour. We'll have to correct that in class on Sunday. Thanks for catching that @djgrant!

> But aren't there callbacks used in `map()`

I think the idea is to introduce them in `.map()`. In my experience students can grasp the idea behind a callback a lot better with `.map()` than with `.sort()`. I think this is because sort is operating on 2 elements in the array at once, where as `.map()` is only operating on 1. Additionally the idea of returning 1 or -1 has really confused students in the past. If it were up to me, I would stay away from sort with callbacks until later in the syllabus.

PR #218, opened 2019-01-11
Fix typo

40thieves:
Fixes a typo


40thieves:
It was Sola who spotted it for me :)

PR #222, opened 2019-01-12
Deepa JS-exercise week- 2

deepaakhil84:


week-2/F-logical-operators/exercise2.js
```diff
@@ -5,17 +5,28 @@
...
+  return number < 0;
+}
+function isBetween5and10(number) {
+  return number >= 5 && number <= 10;
+}
+function isLongName(name) {
+  return name.length > 5;
+}
+function startsWithD(name) {
+  return name[0] === "D";
```
40thieves: Well done for working out that `[0]` works with strings too 🍰

week-2/G-conditionals-2/exercise-3.js
```diff
@@ -9,7 +9,15 @@
 */
 
 function calculateGrade(mark) {
-
+  if (mark >= 80) {
+    return "A";
+  } else if (mark >= 60 && mark < 80) {
```
40thieves: This is correct, but we can make our code a bit simpler here. Because we check if `mark` is more than 80 in the first `if` statement, we know that if it gets to the `else if` statement then `mark` cannot be more than 80.

That means we don't need to check if `mark` is less than 80 in the `else if` statement.

week-2/J-array-get-set/exercise.js
```diff
@@ -5,20 +5,20 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return arr[0]; // complete this statement
 }
 
 function last(arr) {
-  return; // complete this statement
+  return arr[arr.length - 1]; // complete this statement
```
40thieves: Great work! 👍

week-2/J-array-get-set/exercises2.js
```diff
@@ -7,6 +7,9 @@
 */
 
 var numbers = [2, 2, 3]; // Don't change this array literal declaration
+numbers.push(4);
+numbers.shift();
+numbers.unshift(1);
```
40thieves: This is correct, well done for finding an alternative solution. However can actually do this in 1 statement, using array indexes (e.g. `[0]`). Do you know how that might work?

week-2/K-array-methods/exercise2.js
```diff
@@ -4,10 +4,10 @@
   The variable everyone should be an array containing both mentors and students.
 */
 
-var mentors = ["Daniel", "Irina", "Rares"];
+var mentors = ["Daniel", "Irina", "Rares", ","];
 var students = ["Rukmini", "Abdul", "Austine", "Swathi"];
 
-var everyone; // complete this statement
+var everyone = mentors + students; // complete this statement
```
40thieves: Unfortunately this isn't correct in JavaScript (it works in some other programming languages!).

To connect 2 arrays together we need to use the `.concat()` method.

week-2-practise/level-1/4-slice.js
```diff
@@ -5,7 +5,7 @@
   DO NOT EDIT BELOW THIS LINE
   --------------------------- */
 var numbers = [1, 2, 3, 4, 5, , 6, 7, 8];
-var first5Numbers = first5(numbers);
+var first5Numbers = numbers.slice(0, 5);
```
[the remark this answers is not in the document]
NajlaSoltana: I had the same problem when I created a PR for week 2, i find out that my week 1 exercices were there. maybe it is better to delete week-1 exercises from the new Pr. ;)

PR #223, opened 2019-01-12
Alisina92-jsweek2

Alisina92:


week-2/C-comparison-operators/exercise.js
```diff
@@ -7,16 +7,16 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = enoughSpaceInRoom === roomMaxCapacity; // finish this statement
```
40thieves: I don't think this is quite right. The exercise asks you to check if `studentCount` plus `mentorCount` is greater than the `roomMaxCapacity`. How could you fix that?

week-2/D-predicates/exercise.js
```diff
@@ -7,14 +7,14 @@
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
 function isNegative(number) {
-
+  return false;
```
40thieves: This isn't quite right. What happens if we change `number` to be -10 instead of 5? We need to check if `number` is less than 0 instead of just hard-coding `true`. When you fix this, remember to also update the `isBetweenZeroand10` function below.

week-2/F-logical-operators/exercise.js
```diff
@@ -13,24 +13,25 @@ var cssLevel = 4;
 // (hint: use the comparison operator from before)
 var htmlLevelAbove5;
 var cssLevelAbove5;
+htmlLevelAbove5 = htmlLevel > 5;
+cssLevelAbove5 = cssLevel == 5;
```
40thieves: This isn't quite right - this checks if `cssLevel` is exactly 5 not whether it is above 5. How could we fix this?

PR #224, opened 2019-01-12
Abz - week-2 exercise 

ZaimAbz:


week-2/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return arr[0]; // complete this statement
 }
 
 function last(arr) {
-  return; // complete this statement
+  return arr[arr.length - 1]; // complete this statement
```
40thieves: Nice work here!

week-2/I-array-properties/exercise.js
```diff
@@ -6,7 +6,11 @@
 */
 
 function isEmpty(arr) {
-  return; // complete this statement
+  if (arr.length > 0) {
+    return false; // complete this statement
```
40thieves: This is correct but we can improve the code here to make it a bit shorter. Remember that `arr.length > 0` evaluates to a boolean - that is why you can use it in an `if` statement. But here we're checking that boolean to decide whether to return another boolean. We could just return the original boolean. Could you try fixing that up?

week-2/G-conditionals-2/exercise-3.js
```diff
@@ -9,7 +9,15 @@
 */
 
 function calculateGrade(mark) {
-
+  if (mark > 80) {
+    return "A";
+  } else if (80 > mark && mark > 60) {
```
40thieves: This isn't quite correct. Here you're checking that `mark` is greater than 60 **and** it is greater than 80. But we've already checked that `mark` is greater than 80 (in the first `if`). If the computer has got to the `else if`, the `mark` cannot be greater than 80.

week-2/F-logical-operators/exercise2.js
```diff
@@ -5,7 +5,10 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(number) {
+  if (num > 0);
+  return true;
+}
```
LucyMac: There are a couple of issues here:
1) You're returning `true` if the given number is *greater* than 0. Should it be the opposite?
2) whatever word you choose to name your parameter, you *must* repeat the same word inside the function. So here, if your param is named *number*, you should be checking if `number` is greater than / lower than `0`, not `num` - the function has no idea what `num` means.

week-2/G-conditionals-2/exercise-3.js
```diff
@@ -9,7 +9,15 @@
...
+    return "B";
+  } else if (60 > mark && mark > 50) {
+    return "C";
+  } else {
+    return "F";
+  }
```
LucyMac: Good. What if they get `80` is exactly? :)

week-2/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
...
 }
```
LucyMac: Nice

PR #227, opened 2019-01-12
Mohammad week2

markazie:


week-2-practise/level-2/2-validation.js
```diff
@@ -4,7 +4,9 @@
 // - is less than or equal to 100
 // Tip: write other small functions for each requirement
 
-function validate(num) {}
+function validate(num) {
+  return (typeof num === "number" && num > 0 && num <= 100)?true:false;
```
40thieves: Nice use a of a ternary operator!

The instructions ask you to write small functions for each required check. An example might be to write function called `isNumber` that checks if the variable is a number.

week-2-practise/level-3/2-validation.js
```diff
@@ -6,7 +6,9 @@
 
 var excludedNums = [6, 14, 91, 111];
 
-function validate(num) {}
+function validate(num) {
+  return (typeof num === "number" && !excludedNums.includes(num) && Number.isInteger(num))?true:false;
```
40thieves: Again another nice use of a ternary operator. But the same advice from above about small well-named functions applies here too.

week-2-practise/level-3/4-slice.js
```diff
@@ -5,7 +5,8 @@
 // - insert the replacement value at the provided index
 
 function replace(arr, index, value) {
-  return; // complete this statement
+  newArr = Object.assign([], arr, {[index]: value});
```
40thieves: You need to remember to add `var` in front of variables when you declare them

40thieves: I don't think using `Object.assign` is the best way to do this. That method is usually only used for objects, not arrays.

I would look at using one or more of the array methods like `.splice()` and `.slice()`

week-2-practise/level-3/5-indexOf.js
```diff
@@ -6,7 +6,8 @@
 // - return a new array with the replacement value inserted
 
 function replace(arr, valueToReplace, newValue) {
-  return; // complete this statement
+  newArr = Object.assign([], arr, {[arr.indexOf(valueToReplace)]:newValue});
```
40thieves: The same advice about missing variable declarations and not using `Object.assign` with array applies here too.

week-2/D-predicates/exercise.js
```diff
@@ -7,12 +7,12 @@
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
 function isNegative(number) {
-
+if(number > 0) return false;
```
40thieves: This isn't quite right. If you pass in a positive number then you would not get `true` returned back to you. That is because you are only handling the case where the number is negative.

How can you fix this?

week-2/F-logical-operators/exercise2.js
```diff
@@ -5,17 +5,20 @@
   Update the code so that you get the expected result.
 */
 
-function isNegative() {}
+function isNegative(number) {
+  if (number == 5) return false;
```
40thieves: Usually when using the `if` statement we want to use curly braces (`{` and `}`)

week-2/G-conditionals-2/exercise-3.js
```diff
@@ -9,6 +9,7 @@
 */
 
 function calculateGrade(mark) {
+  return (mark >= 80)?"A":(mark >=60 && mark <80)?"B":(mark >= 50 && mark < 60)?"C":"F";
```
40thieves: I generally don't like chaining multiple ternary operators together. I think it makes it quite difficult to read and work out what it is doing. An `if` statement is much easier to read.

In addition I think you can improve the conditional here - because the first conditional checks that `mark` is greater than 80, then you don't need to check that again in the second conditional.

week-2/J-array-get-set/exercise.js
```diff
@@ -5,11 +5,11 @@
 */
 
 function first(arr) {
-  return; // complete this statement
+  return `first value is ${arr[0]} and last array is ${arr[arr.length-1]}`; // complete this statement
```
40thieves: That's not quite what the exercise is asking you to do - it asks you to just return what the first element in the array is, not format that element in a string.

We could then use another function to call `first()` and then format it.

week-2-practise/level-2/2-validation.js
Alisina92: good job

PR #230, opened 2019-01-13
week3

markazie:


week-3/A-array-find/exercise.js
```diff
@@ -3,11 +3,14 @@
   Any result over 60 is a pass. 
   Using .find() (and any other array methods necessary), find the lowest pass score
 */
-
+const findLowestPassScore = arr => {
+  return arr.sort((a, b)=> a-b).find(x => x > 60);
```
40thieves: Well done! This is correct!

week-3/A-array-find/exercise2.js
```diff
@@ -0,0 +1,14 @@
+/*
+  You are given an array of names.
+  Using .find(), we'd like to find the first name which starts with A and is longer than 7 letters.
+*/
+
+// write your code here
+findLongNameThatStartsWithA = names =>names.find(x => x[0] === "A" && x.length > 7);
```
40thieves: Well done! This is very close, but there's 1 small problem.

What if we changed the `names` array to `["Rakesh", "alexandra", "Annam", "Mikey", "Karim", "Ahmed"]`? Would the answer still be correct? How can we fix it?

week-3/B-array-some/exercise.js
```diff
@@ -11,6 +11,9 @@ var pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 // If there is a null value in the array exit the program with the error code
 // https://nodejs.org/api/process.html#process_process_exit_code
 // process.exit(1);
+if (pairsByIndex.some(x => x === null)) {
+  process.exit(1);
```
40thieves: 👍

week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,7 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = group.every(x => x.includes(students)); // complete this statement
```
40thieves: 👏

week-3/D-array-filter/exercise.js
```diff
@@ -7,8 +7,8 @@
 */
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
-
-var pairsByIndex; // Complete this statement
+console.log(typeof pairsByIndexRaw[4]);
+var pairsByIndex = pairsByIndexRaw.filter(x => x != null && typeof x[1] === "number" && typeof x[0] === "number"); // Complete this statement
```
40thieves: Great! Well done

week-3/E-array-map/exercise.js
```diff
@@ -3,3 +3,5 @@
 
 var numbers = [0.1, 0.2, 0.3, 0.4, 0.5];
 
+console.log(numbers.map(x => x*100));
+
```
40thieves: The exercise asks you to write the solution multiple times using different syntax. Can you try using the `function() {}` syntax too?

week-3/F-array-forEach/exercise.js
```diff
@@ -8,6 +8,7 @@
 */
 
 var arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
+console.log(arr.map(x => { return (x % 3 === 0 && x % 5 === 0 && x > 0) ? "FizzBuzz" : (x % 3 === 0 && x > 0) ? "Fizz" : (x % 5 === 0 && x > 0) ? "Buzz" : x }));
```
40thieves: My feedback from last week's homework is the same here. Using multiple chained ternary expressions is usually a bad idea. It is hard to read. Please change this to an `if`/`else if`/`else` statement

week-3-practice/exercise1.js
```diff
@@ -0,0 +1,13 @@
+/* 
+  You are given an array of student marks.
+  If all marks are over 10 the student is graduated, otherwise should try again. 
+  Using .every() (and any other array methods necessary), find out the student is graduated or not.
+  If the student is graduate print "Congratulation, You are graduated", otherwise print "You are nearly there, try again".
+*/
+
+var marks = [11, 19, 13, 20, 18, 16, 10];
+var studentResult = student(marks);
```
40thieves: I think this will throw an error. I think you're missing the `student` function. Please define this function!

week-3-practice/exercise2.js
```diff
@@ -0,0 +1,13 @@
+/* 
+  You are given an string array of numbers.
+  Using array methods (any array methods), find reverse array as a number.
+  If there is 0 before the number you should remove it and print tidy number.
+*/
+
+var stringNumber = ["2", "9", "0", "2", "4", "6", "0"];
+var number = findNumber(stringNumber);
```
40thieves: Again you are missing the `findNumber` function. Please define it!

week-3-practice/exercise3.js
```diff
@@ -0,0 +1,12 @@
+/* 
+  You are given an array of test scores.
+  Any result less than 50 is a fail. 
+  Using .find() (and any other array methods necessary), find the lowest fail score
+*/
+var testScores = [49, 50, 100, 25, 81, 21, 80];
+var lowestPassScore = findLowestPassScore(testScores);
```
40thieves: This exercise too. All of your practice exercises won't work because you are missing the function

week-3-practice/exercise2.js
```diff
@@ -1,13 +1,43 @@
...
+var ages = ["10", "12", "8", "33", "7", "50", "24"];
+
+function filterAges(name) {
+  return ;
+}
+
+var filterAges = ages.filter(x => x<20);
+
+var grandChildNamesAndAges = filterAges.map(function (item, index) {
+  return `${grandChildNames[index]} is ${item} years old`;
```
40thieves: Well done on a slightly complicated question!

week-3-practice/exercise3.js
```diff
@@ -1,12 +1,46 @@
...
+  "candidate5@jobmail.com",
+  "spam@spam.com"
+];
+var times = ["10:00", "10:15", "10:30", "10:45", "11:00", "11:15"];
+var names = ["Gaius", "Adama", "Starbuck", "Boomer", "Gaeta", "spam"];
+function filteremail(item) {
+  return item.sort((a, b) => a.slice(9, 10) - b.slice(9, 10)).filter(x => !x.includes('spam'));
+}
+var sortedArray = filteremail(emails);
+var sortedName = filteremail(names);
```
40thieves: I think this is correct but `filteremail` is a bit of a confusing name because here we are working on names not emails. Could you change it to something like `sortAndFilter`?

week-3-practice/exercise3.js
```diff
@@ -1,12 +1,46 @@
...
+  "candidate3@jobmail.com",
+  "candidate1@jobmail.com",
+  "candidate4@jobmail.com",
+  "candidate5@jobmail.com",
+  "spam@spam.com"
+];
+var times = ["10:00", "10:15", "10:30", "10:45", "11:00", "11:15"];
+var names = ["Gaius", "Adama", "Starbuck", "Boomer", "Gaeta", "spam"];
+function filteremail(item) {
+  return item.sort((a, b) => a.slice(9, 10) - b.slice(9, 10)).filter(x => !x.includes('spam'));
```
40thieves: You don't need to use `.slice()` because we can sort alphabetically using `.sort()` without a callback function.

week-3-practice/exercise3.js
```diff
@@ -1,12 +1,46 @@
...
+}
+var sortedArray = filteremail(emails);
+var sortedName = filteremail(names);
+sortedArray.forEach(function (item, index) {
+  console.log(`Email address: ${item} Dear ${sortedName[index]} your interview is at this time: ${times[index]}`);
```
40thieves: I've commented with some improvements but I think this is the correct solution to a hard question. Well done! 👏

PR #233, opened 2019-01-14
Ivina - JS - Week 3

ivinapontes:


week-3/A-array-find/exercise.js
```diff
@@ -1,13 +1,18 @@
...
-var testScores = [90, 50, 100, 25, 81, 66, 80];
-var lowestPassScore = findLowestPassScore(testScores);
+// write your code here
 
-console.log(lowestPassScore);
+var names = ["Rakeshss", "Alexandra", "Annam", "Mikey", "Karim", "Ahmed"];
+var longNameThatStartsWithA = names.find(findLongNameThatStartsWithA);
+
+function findLongNameThatStartsWithA(names) {
+  return names.length >= 7 && names[0] === "A";
```
40thieves: This is really [nitpicking](https://dictionary.cambridge.org/dictionary/english/nitpicking), but the exercise asks you to find names that are longer than 7 letters, not equal to or longer than 7 letters.

Also what happens if you change the names array so that "Alexandra" is spelled like "alexandra"? How could we fix this?

week-3/B-array-some/exercise.js
```diff
@@ -16,6 +16,9 @@ var students = ["Islam", "Lesley", "Harun", "Rukmini"];
 var mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
 
 var pairs = pairsByIndex.map(function(indexes) {
+  if (indexes === null) {
+    process.exit(1);
```
40thieves: This does work 🍰! But lets say you had to solve it using the `.some()` method. How would you do that?

week-3/C-array-every/exercise.js
```diff
@@ -5,14 +5,15 @@
...
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = group.every(checkForStudents);
 
-if (groupIsOnlyStudents) {
-  console.log("The group contains only students");
-} else {
-  console.log("The group does not contain only students");
+function checkForStudents(groupIsOnlyStudents) {
+  if (groupIsOnlyStudents.includes(students)) {
+    console.log("The group contains only students");
```
40thieves: I don't think this is quite right. Remember that the `.every()` method expects you to return a boolean, so that it find determine if every element in the array matches the condition in your predicate function.

The exercise is a little bit confusing as it doesn't run the `console.log(groupIsOnlyStudents)` for you, but you can add that after calling `.every()`

week-3/D-array-filter/exercise.js
```diff
@@ -8,7 +8,13 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-var pairsByIndex; // Complete this statement
+var pairsByIndex = pairsByIndexRaw.filter(function(indexes) {
+  if (indexes !== null || indexes !== "string" || indexes !== "Boolean") {
+    return indexes;
```
40thieves: This is close, but not exactly right. Remember that the filter method expects you to return a boolean (`true` or `false`) from the callback function. Here we're returning the element itself.

week-3/D-array-filter/exercise.js
```diff
@@ -8,7 +8,13 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-var pairsByIndex; // Complete this statement
+var pairsByIndex = pairsByIndexRaw.filter(function(indexes) {
+  if (indexes !== null || indexes !== "string" || indexes !== "Boolean") {
```
40thieves: I'm not sure this is quite right. Does the string "whoops" match the string "string"? Do we instead need to check the type?

week-3/E-array-map/exercise.js
```diff
@@ -2,4 +2,22 @@
...
+
+var multiplyNumbers = numbers.map(function(number) {
+  return number * 100;
+});
+
+/*Case 3 */
+
+var multiplyNumbers = numbers.map(function multiplyNumbers(number) {
+  return number * 100;
+});
```
40thieves: Nice work! They're all correct! There's 1 more potential that I can think of, can you think what it might be? Hint: you used it in your `.find()`  and `.every()` solutions

week-3/F-array-forEach/exercise.js
```diff
@@ -23,7 +34,7 @@ var arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
 'Fizz'
 'Buzz'
 11
-'FizzBuzz'
+'Fizz'
```
40thieves: Oops! Thanks for catching that mistake!

week-3/F-array-forEach/exercise.js
```diff
@@ -8,6 +8,17 @@
 */
 
 var arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
+arr.forEach(function(number) {
```
40thieves: Well done! This is correct 🎉 FizzBuzz is a famous exercise for testing your programming knowledge 🙂

week-3-pratise/exercise1.js
```diff
@@ -0,0 +1,16 @@
+/* Given is an array find the element with characters more than 6 */
+
+var food = ["Pizza", "Burger", "Fries", "Kebab", "Chicken Wings"];
+
+var findLongestFoodName = food.filter(function(food) {
```
40thieves: Is this correct? You might need to ask the person who created the exercise!

It says in the description to *find the element*, so I think they mean you should only find 1 answer. But your solution could find multiple answers it's just chance that there's only 1 element in the array that matches (if you added "Lasagne" to the array would it still work?)

But I might be reading the question wrong. The *intention* is not clear

week-3-pratise/exercise3.js
```diff
@@ -0,0 +1,14 @@
+/* Given is an array return an array containing only even numbers */
+
+var numbers = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12];
+
+var filterNumber = numbers.filter(function(number) {
+  if (number % 2 === 0) {
+    return number;
```
40thieves: This has the same problem as the other `.filter()` exercise - you need to return a boolean, not the element itself.

week-3-pratise/exercise3.js
```diff
@@ -0,0 +1,14 @@
+/* Given is an array return an array containing only even numbers */
+
+var numbers = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12];
+
+var filterNumber = numbers.filter(function(number) {
+  if (number % 2 === 0) {
```
40thieves: Excellent job working out how to calculate if a number is divisible by 2!

PR #237, opened 2019-01-15
Tayo-JS-week3

Atayo:


week-3/A-array-find/exercise.js
```diff
@@ -1,13 +1,17 @@
...
+/*
+  You are given an array of names.
+  Using .find(), we'd like to find the first name which starts with A and is longer than 7 letters.
 */
 
-var testScores = [90, 50, 100, 25, 81, 66, 80];
-var lowestPassScore = findLowestPassScore(testScores);
+// write your code here
+function findLongNameThatStartsWithA(name) {
+  return name.length > 7 && name[0] === "A";
```
40thieves: Nice work :)

There is a little problem though. What happens if we change the `names` array to have "alexandra" instead of "Alexandra"? Is our answer still correct? How could we fix it?

week-3/B-array-some/exercise.js
```diff
@@ -15,6 +16,11 @@ var pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 var students = ["Islam", "Lesley", "Harun", "Rukmini"];
 var mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
 
+function isNull(value) {
+  if (value === null) {
+    process.exit(1);
```
40thieves: This is technically correct, but not quite what the exercise was asking for. Remember that the `.some()` method will return either true (which means that the array does contain some null values) or false (which means that the array does not contain some null values).

We could then use this boolean to decide if we should exit the program.

week-3/C-array-every/exercise.js
```diff
@@ -5,8 +5,9 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
-
+var groupIsOnlyStudents = group.every(function(name) {
+  return students.includes(name); // complete this statement
```
40thieves: Excellent! Well done!

week-3/D-array-filter/exercise.js
```diff
@@ -8,7 +8,17 @@
...
 
-var pairsByIndex; // Complete this statement
+var pairsByIndex = pairsByIndexRaw.filter(function(indexes) {
+  // Complete this statement
+  if (
+    indexes !== null &&
+    indexes.length == 2 &&
+    indexes !== "Boolean" &&
+    indexes !== "string"
+  ) {
```
40thieves: A very through conditional! 👍

week-3/D-array-filter/exercise.js
```diff
@@ -8,7 +8,17 @@
...
+    return indexes;
```
40thieves: This will work, but isn't exactly right. Remember that `.filter()` expects your callback function to return a boolean (`true` or `false`). Here we're returning the element in the array

week-3/E-array-map/exercise.js
```diff
@@ -2,4 +2,6 @@
 // Write multiple solutions using different syntax (as shown in the README)
 
 var numbers = [0.1, 0.2, 0.3, 0.4, 0.5];
+var numbersByHundred = numbers.map(number => number * 100);
```
40thieves: This is correct, but the exercise asks you to try writing the solution multiple times using different syntax. Can you try that?

week-3/F-array-forEach/exercise.js
```diff
@@ -8,6 +8,17 @@
 */
 
 var arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
+arr.forEach(function(number) {
+  if (number % 3 === 0 && number % 5 === 0) {
```
40thieves: Really nicely done! 👏 FizzBuzz is a famous exercise for testing your programming skill 🙂

week-3/week-3-practice/london.js
```diff
@@ -0,0 +1,23 @@
...
+Console.log has been provided
+*/
+
+var londonLocation = [
+  ["London Bridge", "tube", "bus", "river boat"],
+  ["Tower bridge", "tube", "bus"],
+  ["Greenwich", "tube", "bus", "river boat"]
+];
+
+var londonLocationTransport = londonLocation.map(function(item) {
```
Atayo: This works better and I tried to replace it but did not work for some reason 🤔
`var londonLocationTransport = [];`
`londonLocation.forEach(function(item) {`
  `if (item.length === 4) {`
   ` londonLocationTransport.push([item[0], item[3]]);`
 ` }`
`});`

PR #240, opened 2019-01-15
Bart Week 3 Homework 

bmc82:


bmc82:
@40thieves Have added the completed week 3 practise exercises given by Ivina. Thanks for your input on the others. Will have a go at your suggested changes in the morning before class. I'm whacked after that training today and working all week!

PR #241, opened 2019-01-15
zareen js week-3

khan-zareen:


week-3/A-array-find/exercise.js
```diff
@@ -1,13 +1,16 @@
...
 
-var testScores = [90, 50, 100, 25, 81, 66, 80];
-var lowestPassScore = findLowestPassScore(testScores);
+// write your code here
 
-console.log(lowestPassScore);
+var names = ["Rakesh", "Alexandra", "Annam", "Mikey", "Karim", "Ahmed"];
+var longNameThatStartsWithA = names.find(findLongNameThatStartsWithA);
+function findLongNameThatStartsWithA(name) {
+  return name[0].includes("A") && name.length > 7;
```
40thieves: This is very correct, but you can make a small readability improvement. When using `name[0]` that will return a string only 1 length long. The `.includes()` method will search the **whole** string to check if any of the string includes the "A" letter.

So that means that using `.includes()` is a bit unnecessary to check a 1 length string. It might be better to directly compare `name[0]` against `"A"`.

40thieves: There is another small improvement that you could make. What happens if you change the `names` array so that "Alexandra" starts with a lower case "a" (`"alexandra"`)? How could we fix this?

week-3/B-array-some/exercise.js
```diff
@@ -15,6 +15,10 @@ var pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 var students = ["Islam", "Lesley", "Harun", "Rukmini"];
 var mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
 
+var pairs = pairsByIndex.some(noPair);
+function noPair(index) {
+  if (index == null) return process.exit(1);
```
40thieves: Nice work! Well done 🙂

week-3/C-array-every/exercise.js
```diff
@@ -5,8 +5,12 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
-
+var groupIsOnlyStudents = group.every(isStudent); // complete this statement
+function isStudent(index) {
+  if (group.includes(students)) {
```
40thieves: Very nice usage of `.includes()`! 🏅

week-3/C-array-every/exercise.js
```diff
@@ -5,8 +5,12 @@
...
+    return true;
+  }
```
40thieves: This is technically correct, but usually when we're writing a *predicate* function we want to return either `true` or `false`. Here you are only returning `true` (which is correct!), but you can't returning anything at all if the condition doesn't match.

week-3/D-array-filter/exercise.js
```diff
@@ -8,8 +8,10 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-var pairsByIndex; // Complete this statement
-
+var pairsByIndex = pairsByIndexRaw.filter(indexValue); // Complete this statement
+function indexValue(arr) {
+  return arr != null && arr.length === 2;
```
40thieves: Excellent!

week-3/E-array-map/exercise.js
```diff
@@ -2,4 +2,16 @@
...
+  return number * 100;
+}
+var numberMultipliedBy100 = numbers.map(multipliedBy100);
+var numberMultipliedBy100 = numbers.map(function multipliedBy100(number) {
+  return number * 100;
+});
+var numberMultipliedBy100 = numbers.map(number => {
+  return number * 100;
+});
+var numberMultipliedBy100 = numbers.map(number => number * 100);
```
40thieves: This is really nice! Well done 🍰

week-3/F-array-forEach/exercise.js
```diff
@@ -8,7 +8,19 @@
...
+function multiples(index) {
+  if (index % 3 === 0 && index % 5 === 0) {
+    console.log("FizzBuzz");
+  } else if (index % 5 === 0) {
+    console.log("Buzz");
+  } else if (index % 3 === 0) {
+    console.log("Fizz");
+  } else {
+    console.log(index);
+  }
```
40thieves: Great! Well done!

FizzBuzz is a famous way of testing your programming skills 😄

week-3/B-array-some/exercise.js
Alisina92: well done Zareen.

khan-zareen: thanks

week-3/C-array-every/exercise.js
ZaimAbz: sorry do you mean thats he need to complete it by adding else statement?!

week-3/A-array-find/exercise.js
khan-zareen: the code can be changed like this for both things from
return name[0].includes("A") && name.length > 7;
to 
return name[0]==="A" && name.length > 7;

week-3/C-array-every/exercise.js
khan-zareen: correction done to {return group.includes(students);}

week-3-practice/exercise1.js
```diff
@@ -0,0 +1,18 @@
...
+  return number * 10;
+}
+var numbersMultiplied = numbers.map(function(number) {
+  return number * 10;
+});
+var numbersMultiplied = numbers.map(number => {
+  return number * 10;
+});
+var numbersMultiplied = numbers.map(number => number * 10);
+console.log(numbersMultiplied);
```
40thieves: Well done!

week-3-practice/exercise2.js
```diff
@@ -0,0 +1,15 @@
+/*
+we have a list of grades, we would like to know
+all the greads that 80 and above.
+*/
+
+var grades = [90, 80, 77, 78, 60, 56, 86, 79, 55];
+
+var list = grades.filter(grade => grade >= 80); //Complete this line
```
40thieves: Correct 👍

week-3-practice/exercise3.js
```diff
@@ -0,0 +1,27 @@
...
+    .replace("/", "");
+}
+
+function captialise(str) {
+  return str.split("")[0].toUpperCase() + str.slice(1); // complete this function
+}
+
+var cars = [" vauxhall/ ", "bMw ", " ferrari", "nissan "];
+var carsTidy = cars.map(tidyUpString);
+var carsTidyAndCapitalised = carsTidy.map(captialise);
```
40thieves: Nice work :)

PR #243, opened 2019-01-16
Ahmed-week3-exercises

ahmedahashim:


week-3-practise/exercise-2/exercise-2.js
```diff
@@ -0,0 +1,28 @@
...
+var details = [
+  ["Mildred", 20, "Chemist"],
+  ["Jerry", 24, "Engineer"],
+  ["Choke", 30, "Doctor"],
+  ["Murice", 28, "House wife"],
+  ["Daniel", 30, "Baron"],
+  ["Micheal", 34, "Judge"]
+];
+
+var nameAndDepartment = details.map(names => [names[0], names[2]]); //complete the statement
```
40thieves: Nice work!

week-3-practise/exercise-3/exercise-3.js
```diff
@@ -0,0 +1,29 @@
...
+  "Marcathy",
+  "Sammy",
+  "Deadly",
+  "carmel",
+  "Esus",
+  "Damina"
+];
+var teamName = ["Frank", "Junior", "Smart", "Gene", "Marcus", "zack", "Fabio"];
+
+var teamOnlyList = names.every(name => teamName.includes(name) !== true); // complete this statement
```
40thieves: This is correct, but you can make the code a bit shorter. Remember that `.includes()` always returns a boolean. So you don't need to compare against `true`.

week-3/A-array-find/exercise.js
```diff
@@ -1,13 +1,17 @@
...
-  Using .find() (and any other array methods necessary), find the lowest pass score
+  You are given an array of names.
+  Using .find(), we'd like to find the first name which starts with A and is longer than 7 letters.
 */
 
-var testScores = [90, 50, 100, 25, 81, 66, 80];
-var lowestPassScore = findLowestPassScore(testScores);
+// write your code here
+function longNameThatStartsWithA(name) {
+  return name[0].includes("A") && name.length > 7;
```
40thieves: This is correct but you could make a few small improvements.

Firstly when you use `[0]` on a string it will return 1 character. `.includes()` checks if the whole string includes the parameter that you pass to it. But we know that it is just 1 character, so we can just compare directly using `=== "A"`.

The second small problem is what happens if we change the `names` array so that "Alexandra" is changed to "alexandra"? Does it still work? How could we fix that?

week-3/B-array-some/exercise.js
```diff
@@ -12,6 +12,11 @@ var pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 // https://nodejs.org/api/process.html#process_process_exit_code
 // process.exit(1);
 
+var containNull = pairsByIndex.some(function(include) {
+  include === null;
+  console.log(" null value is found in this array");
+  return process.exit(1);
```
40thieves: I don't think this is quite right. We aren't checking if the value is actually null - we need to use an `if` statement for that. How could we fix that?

week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,15 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+function Check(names) {
+  if (students == group) {
```
40thieves: This isn't quite right. Using `==` to compare 2 arrays doesn't work. And here we don't want to compare the whole `group` array against `students`. We want to compare each individual element in the `group` array instead. How can we fix that?

week-3/D-array-filter/exercise.js
```diff
@@ -8,8 +8,15 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-var pairsByIndex; // Complete this statement
+var filterarr = [];
 
+pairsByIndexRaw.filter(function(item) {
+  if (item !== null && item !== false && item !== "whoops" && item.length > 1) {
+    filterarr.push(item);
```
40thieves: This will work, but it's not what we really want to do here. Remember that the `.filter()` method will create a new array for you, based on whether you return `true` or `false` for each element in the `pairsByIndexRaw` array.

week-3/E-array-map/exercise.js
```diff
@@ -2,4 +2,37 @@
...
+//solution 4
+
+var multipliedNumbers = numbers.map(numbers => {
+  return numbers * 100;
+});
+console.log(multipliedNumbers);
+
+//solution 5
+
+var multipliedNumbers = numbers.map(numbers => numbers * 100);
```
40thieves: Yes those are all correct! Well done :)

week-3/F-array-forEach/exercise.js
```diff
@@ -8,6 +8,17 @@
 */
 
 var arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
+arr.map(function isMultipleOf3(number) {
+  if (number % 3 == 0 && number != 0) {
+    console.log("Fizz");
+  } else if (number % 5 == 0 && number != 0) {
+    console.log("Buzz");
+  } else if (number % 3 == 0 && number % 5 == 0 && number != 0) {
```
40thieves: Have you tested this? Did you look carefully at the expected output below? Is it exactly the same as the output you see when you run this?

Remember that we compare the `if` statement first, then any `else if` statements in order.

PR #245, opened 2019-01-16
clement-JS-week3

clems27:


week-3/A-array-find/exercise.js
```diff
@@ -6,7 +6,11 @@
 // write your code here
 
 var names = ["Rakesh", "Alexandra", "Annam", "Mikey", "Karim", "Ahmed"];
-var longNameThatStartsWithA = findLongNameThatStartsWithA(names);
+var longNameThatStartsWithA = names.find(longNameThatStartsWithA)
+
+function longNameThatStartsWithA(name) {
+    return name[0] === "A" && name.length > 7;
```
40thieves: This is correct, but we can make a small improvement. What happens if we change the `names` array so that it has "alexandra" instead of "Alexandra"? Does it still work? How could we fix it?

week-3/B-array-some/exercise.js
```diff
@@ -11,6 +11,11 @@ var pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 // If there is a null value in the array exit the program with the error code
 // https://nodejs.org/api/process.html#process_process_exit_code
 // process.exit(1);
+var containNull = pairsByIndex.some(function (include) {
+  include === null;
```
40thieves: I don't think this is quite right. We're not actually checking if the `include` variable is null. We need to use an `if` statement to do that.

week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,11 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+function studentsGroup(element) {
+  return students.includes(element);
```
40thieves: Excellent! Well done!

week-3/D-array-filter/exercise.js
```diff
@@ -8,7 +8,10 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-var pairsByIndex; // Complete this statement
+var pairsByIndex = pairsByIndexRaw.filter(function (pairs) {
+  return pairs !== null && pairs.length === 2;
```
40thieves: 👍 :)

week-3/E-array-map/exercise.js
```diff
@@ -3,3 +3,14 @@
...
 
+// var number100 =numbers.map(double);
+
+// function double(numbers) {
+//     return numbers * 100;
+// }
+// console.log(number100);
+
+var number100 = numbers.map(function (num) {
+    return num * 100;
```
40thieves: This is correct, but the exercise asks you to write multiple solutions using different syntax. Can you try that?

week-3/F-array-forEach/exercise.js
```diff
@@ -9,6 +9,23 @@
...
+  }
+  else if (nums % 5 === 0) {
+    console.log("Buzz");
+  }
+  else if (nums % 3 === 0) {
+    console.log("Fizz");
+  }
+  else {
+    console.log(nums);
+  }
```
40thieves: Well done! 😄 FizzBuzz is a famous way of testing your programming skill!

week-3/A-array-find/exercise.js
clems27: @40thieves Thanks for pointing that out, I could improve the code by just returning name[0] ===name[0] && name.length >7

week-3/B-array-some/exercise.js
clems27: @40thieves Not quite sure how the process.exit(1), work...

week-3/C-array-every/exercise.js
clems27: Thanks

week-3/E-array-map/exercise.js
clems27: @40thieves the lines commented out above from line number 6 to line number 11 is another method, please do take a look

week-3/F-array-forEach/exercise.js
clems27: thanks

PR #246, opened 2019-01-16
Elamin-Week-3-exercises

aldeez988:


week-3-practise/exercise1.js
```diff
@@ -0,0 +1,13 @@
+/* convert the array money to british pounds from us dollars
+ */
+
+var convertToPound = [];
+var Money = ["$50", "$2000", "$300", "$500"];
+/* write your answer here */
+convertToPound = Money.map(money => {
+  return "£" + (0.78 * parseFloat(money.slice(1))).toFixed(2);
```
40thieves: Well done!

week-3-practise/exercise2.js
```diff
@@ -0,0 +1,13 @@
+/*  print the ages greater than 15 */
+
+var filterAge = [];
+var ages = [1, 2, 3, 4, 5, 10, 15, 20, 25, 30, 35];
+// write your answer here :
+filterAge = ages.filter(age => age > 15);
```
40thieves: Excellent!

week-3-practise/exercise3.js
```diff
@@ -0,0 +1,10 @@
+// find all the names thats start with an A :
+
+var names = ["Alice", "benjamin", "Angela", "Adam", "Bob"];
+// write answer here
+var nameStartWithA = names.filter(name => name[0] === "A");
```
40thieves: This is technically correct, but I think the *intention* of the question is a little different.

What happens if I change "Alice" to "alice"? Is it still correct? How could we fix it?

week-3/A-array-find/exercise.js
```diff
@@ -1,13 +1,15 @@
...
-  Any result over 60 is a pass. 
-  Using .find() (and any other array methods necessary), find the lowest pass score
+  You are given an array of names.
+  Using .find(), we'd like to find the first name which starts with A and is longer than 7 letters.
 */
 
-var testScores = [90, 50, 100, 25, 81, 66, 80];
-var lowestPassScore = findLowestPassScore(testScores);
+function firstLongNameStartWithA(name) {
+  return name[0] === "A" && name.length > 7;
```
40thieves: This is a similar problem to the solution above: what happens if we change "Alexandra" to "alexandra"?

week-3/B-array-some/exercise.js
```diff
@@ -6,8 +6,14 @@
...
 
+//The function checkNullValue  will exit if it find an element that has a null value
+function checkNullValue(arrElement) {
+  if (arrElement === null) {
+    process.exit();
+  }
+}
 var pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
-
+pairsByIndex.some(checkNullValue);
```
40thieves: This is technically correct but the exercise is asking you to find out whether the array contains any null values, then use that to decide whether to exit the process.

Remember that `.some()` returns a boolean that you can decide whether to exit the process or not.

week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,9 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = students.every(student => {
+  return group.includes(student) !== true;
```
40thieves: This is correct, but we can make the code a bit shorter. Remember that `.includes()` always returns a boolean. So we don't need to compare it against `true`

week-3/D-array-filter/exercise.js
```diff
@@ -8,7 +8,12 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-var pairsByIndex; // Complete this statement
+var pairsByIndex = pairsByIndexRaw.filter(arrayElemnet => {
+  if (arrayElemnet !== null)
+    return (
+      typeof arrayElemnet[0] === "number" && typeof arrayElemnet[1] === "number"
```
40thieves: This is correct but we could make this shorter by *chaining* the conditionals together. Something like this:

```js
return myVariable !== null && typeof myVariable[0] === "number"
```

week-3/E-array-map/exercise.js
```diff
@@ -1,5 +1,27 @@
...
+
+var mumultipliedBy100_v3 = numbers.map(function(number) {
+  return number * 100;
+});
+
+var mumultipliedBy100_v4 = numbers.map(number => {
+  return number * 100;
+});
+
+var mumultipliedBy100_v4 = numbers.map(number => number * 100);
```
40thieves: Well done! That is all of the different syntaxes 💪

week-3/F-array-forEach/exercise.js
```diff
@@ -8,7 +8,17 @@
...
-
+arr.forEach(number => {
+  if (number % 5 === 0 && number % 3 === 0 && number !== 0) {
+    console.log("FizzBuzz");
+  } else if (number % 3 === 0 && number !== 0) {
+    console.log("Fizz");
+  } else if (number % 5 === 0 && number !== 0) {
+    console.log("Buzz");
+  } else {
+    console.log(number);
```
40thieves: Well done! 🍰 FizzBuzz is a famous way of testing programming skill!

PR #247, opened 2019-01-16
Ermi-week3-js exercises

ermiii:


week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,7 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = false; // complete this statement
```
40thieves: This isn't quite right. If we change the contents of the array, then it may be incorrect.

We instead need to use the `.every()` method to check if every element within the `group` array is part of the `students` array.

Remember that we can check if an individual value (e.g. a group member) is part of an array (e.g. students) using the `.includes()` method

week-3/E-array-map/exercise.js
```diff
@@ -1,5 +1,19 @@
 // Using the .map() method, create a new array with `numbers` multiplied by 100
 // Write multiple solutions using different syntax (as shown in the README)
-
 var numbers = [0.1, 0.2, 0.3, 0.4, 0.5];
 
+var multipliedBy100 = numbers.map(function (number) {
+    return number * 100;
+});
+console.log(multipliedBy100);
```
40thieves: This is correct! Well done.

The exercise asks you to write multiple solutions using different syntax. Can you try that?

week-3/F-array-forEach/exercise.js
```diff
@@ -6,8 +6,18 @@
 
   An array with numbers 0-15 has been provided.
 */
-
 var arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
+arr.forEach(number => {
+  if (number % 5 === 0 && number % 3 === 0 && number !== 0) {
```
40thieves: Well done! 🍰 FizzBuzz is a famous way of testing programming skill

PR #249, opened 2019-01-16
Week 3 exercises

Mussie14:


week-3-practice/exercise-2.js
```diff
@@ -0,0 +1,21 @@
+/* Odd or Even: write a program that checks if  
+the numbers in an array are odd or even and
+ returns one of the following messages:
+“All Odd!”, “All Even !” or “All Mixed”
+An array of numbers is provided */
+
+var numbers = [7, 10, 4, 5, 12];
+
```
theseafaringturtle: I would've done it with .every() but this is a good solution too!

PR #250, opened 2019-01-16
Ahmet-JS-Week3-Exercises Review-1

Ahmet-K:


week-3/A-array-find/exercise.js
```diff
@@ -1,13 +1,19 @@
...
-var testScores = [90, 50, 100, 25, 81, 66, 80];
-var lowestPassScore = findLowestPassScore(testScores);
+// write your code here
 
-console.log(lowestPassScore);
+var names = ["Rakesh", "Alexandra", "Annam", "Mikey", "Karim", "Ahmed"];
+var longNameThatStartsWithA = names.find(findLongNameThatStartsWithA);
+
+function findLongNameThatStartsWithA(name) {
+  return name[0].includes("A") && name.length > 7;
```
martynamaron: Good! One point about checking for the letter "A" though - if `name` is a string, you know that `name[0]` will be just one character - so it can't really "include" anything. Checking if that character is equal to something is good enough :)

week-3/D-array-filter/exercise.js
```diff
@@ -7,8 +7,12 @@
 */
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
+var pairsByIndex=pairsByIndexRaw.filter(function(pair){
+  return typeof (pair) == "object" && pair !== null && pair.length==2;
```
martynamaron: A very through check, well done :) 👏

week-3/E-array-map/exercise.js
```diff
@@ -3,3 +3,10 @@
 
 var numbers = [0.1, 0.2, 0.3, 0.4, 0.5];
 
+var newArraySolution1= numbers.map(number=>number*2);
+console.log(newArraySolution1);
+
+var newArraySolution2= numbers.map(function(number){
+    return number*2;
+})
```
martynamaron: You were meant to multiply the numbers by 100 :)

There's also a couple more way you can write a callback, could you add those in as well? :)

week-3/F-array-forEach/exercise.js
```diff
@@ -8,6 +8,17 @@
 */
 
 var arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
+var newArr = arr.forEach(function(number){
```
martynamaron: What does the `newArr` variable hold? Do you need it?

week-3/F-array-forEach/exercise.js
```diff
@@ -8,6 +8,17 @@
...
+  if ((number%3===0) && (number%5===0)){
```
martynamaron: Minor point, but you don't need the double brackets here :)

Also, your output prints out "FizzBuzz" where 0 is. The expected output doesn't include this - could you fix that at all? :)

martynamaron:
Well done with your exercises! :) Good work on the `some()` method - I think this one is a little tricky. I've left some comments for minor improvements, but it would be good if you could have a look :) Thanks!

PR #251, opened 2019-01-17
Sola js week3

solasobowale:


week-3/A-array-find/exercise.js
```diff
@@ -1,13 +1,19 @@
...
 
-var testScores = [90, 50, 100, 25, 81, 66, 80];
-var lowestPassScore = findLowestPassScore(testScores);
+// write your code here
 
-console.log(lowestPassScore);
+var names = ["Rakesh", "Alexandra", "Annam", "Mikey", "Karim", "Ahmed"];
+
+function findLongNameThatStartsWithA(name) {
+return (name[0] === "A" && name.length > 7) ;
```
martynamaron: Great stuff! :)

week-3/B-array-some/exercise.js
```diff
@@ -6,7 +6,7 @@
   - Do not edit any of the existing code
 */
 
-var pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
+var pairsByIndex = [[0, 3], [1, 2], [2, 1],[3, 0]];
```
martynamaron: Please put that `null` value back, it's meant to be here 😄

week-3/B-array-some/exercise.js
```diff
@@ -15,6 +15,13 @@ var pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
 var students = ["Islam", "Lesley", "Harun", "Rukmini"];
 var mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
 
+function isNull(value) {
+  if (value === null) {
+    process.exit(1);
+  }
+}
+pairsByIndex.some(isNull);
```
martynamaron: I think this got a little mixed up - `some()` will return you a boolean. If you use `value === null` in a function that you put into `some()` - that will check every **value** in the `pairsByIndex` array - and if any of them is `null`, it'll return `true`. Then you can use that boolean in a condition, and only if it's `true` - call `process.exit(1);`

Hope this makes more sense now - can you please try again? :)

week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,7 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = group.every(pupil => students.includes(pupil)); // complete this statement
```
martynamaron: Nice and concise too! 🎉

week-3/D-array-filter/exercise.js
```diff
@@ -8,7 +8,9 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-var pairsByIndex; // Complete this statement
+var pairsByIndex = pairsByIndexRaw.filter(
+  entry => entry !== null && entry.length === 2
```
martynamaron: Yay! 🎉 🎉 🎉

week-3/E-array-map/exercise.js
```diff
@@ -3,3 +3,30 @@
...
+
+var numbersTimesHundred5 = numbers.map(number => number * 100);
+
+console.log(
+  numbersTimesHundred1,
+  numbersTimesHundred2,
+  numbersTimesHundred3,
+  numbersTimesHundred4,
+  numbersTimesHundred5
+);
```
martynamaron: Awesome! All bases covered :)

week-3/F-array-forEach/exercise.js
```diff
@@ -9,6 +9,20 @@
...
+      console.log("FizzBuzz");
+    } else if (number % 3 === 0) {
+      console.log("Fizz");
+    } else if (number % 5 === 0) {
+      console.log("Buzz");
+    } else {
+      console.log(number);
+    }
+  }
+});
```
martynamaron: Perfect! I like your check for `0` at the top and then nesting the other checks inside of it :)

martynamaron:
Well done @solasobowale !! Nice and exhaustive solutions :) Only one hiccup with `some()` but I'm sure you get it. Very well done 👏

week-3-practise/exercise-1.js
```diff
@@ -0,0 +1,11 @@
+/* The following numbers are ages for a group of students. Create a function so that you’ll find out their ages “in total” after 10 years
+Hint: Use find method */
+var numbers = [18, 17, 19, 16];
+var checkingAges = numbers.map(number => number + 10);
+var totalAgeInTotal = checkingAges.reduce((a, b) => a + b, 0);
```
40thieves: Wow! Great usage of the `.reduce()` method! This is a very powerful but quite complex method

week-3-practise/exercise-2.js
```diff
@@ -0,0 +1,11 @@
+/* You are working a project in the council to give some benefits to the aged people older than 65. Find out them in a new array in the following list */
+var numbers = [56, 45, 32, 67, 89, 41, 65, 62, 59, 73];
+
+var findTheAgedOnes = numbers.filter(number => number > 65);
+
+console.log(findTheAgedOnes.toString());
```
40thieves: I don't think you need to convert the array to a string here. `console.log` is very good at outputting the right thing even if you give it an array.

week-3-practise/exercise-2.js
```diff
@@ -0,0 +1,11 @@
+/* You are working a project in the council to give some benefits to the aged people older than 65. Find out them in a new array in the following list */
+var numbers = [56, 45, 32, 67, 89, 41, 65, 62, 59, 73];
+
+var findTheAgedOnes = numbers.filter(number => number > 65);
```
40thieves: Nice! Well done

week-3-practise/exercise-3.js
```diff
@@ -0,0 +1,14 @@
+/* In Turkish Language, some of the words are joined words and 
+they are written together. Create a new function in order to get a 
+new array with the joined words of each items in the array*/
+
+/*var newWords = [
+  { firstword: "Ata", lastword: "turk" },
```
40thieves: This is a challenging question! We are going to look at objects on Sunday.

week-3-practise/exercise-3.js
```diff
@@ -0,0 +1,14 @@
...
+  { firstword: "Galata", lastword: "saray" },
+  { firstword: "Eski", lastword: "sehir" }
+]; */
+
```
40thieves: Have you tried implementing this one?

If we break it down, we can think about it as **transforming** each element in the array from an object into a string. Do you remember what we talked about in class about the `.map()` method? If we are transforming each element in the array into something else that is when we want to use the `.map()` method.

I'll give you another hint about working with objects. We need to use the dot (`.`), like with methods to get a *property* of the object. So in this case to join the first and last words together we could write something like this:

```js
var joinedWord = word.firstword + word.lastword;
```

PR #252, opened 2019-01-17
week-3 Abz A, B, C, D, E and F

ZaimAbz:
I have just finished week 3 now I will add the exercises from Semantha to week-3 practise and create my exercises and I will do a new PR

### DIFF
--- a/week-3/A-array-find/exercise.js
+++ b/week-3/A-array-find/exercise.js
@@ -4,10 +4,11 @@
   Using .find() (and any other array methods necessary), find the lowest pass score
 */
 
-var testScores = [90, 50, 100, 25, 81, 66, 80];
-var lowestPassScore = findLowestPassScore(testScores);
+var names = ["Rakesh", "Alexandra", "Annam", "Mikey", "Karim", "Ahmed"];
+var longNameThatStartsWithA = names.find(function(name) {
+  return name[0] === "A" && name.length > 7;
+});
 
-console.log(lowestPassScore);
-
+console.log(longNameThatStartsWithA);
 /* EXPECTED OUTPUT */
 // 66
--- a/week-3/B-array-some/exercise.js
+++ b/week-3/B-array-some/exercise.js
@@ -16,6 +16,9 @@ var students = ["Islam", "Lesley", "Harun", "Rukmini"];
 var mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
 
 var pairs = pairsByIndex.map(function(indexes) {
+  if (indexes === null) {
+    process.exit(1);
+  }
   var student = students[indexes[0]];
   var mentor = mentors[indexes[1]];
   return [student, mentor];
--- a/week-3/C-array-every/exercise.js
+++ b/week-3/C-array-every/exercise.js
@@ -5,7 +5,9 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = students.every(function(student) {
+  return group.includes(student) !== true;
+}); // complete this statement
 
 if (groupIsOnlyStudents) {
   console.log("The group contains only students");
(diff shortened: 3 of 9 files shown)


week-3/A-array-find/exercise.js
```diff
@@ -4,10 +4,11 @@
   Using .find() (and any other array methods necessary), find the lowest pass score
 */
 
-var testScores = [90, 50, 100, 25, 81, 66, 80];
-var lowestPassScore = findLowestPassScore(testScores);
-
-console.log(lowestPassScore);
+var names = ["Rakesh", "Alexandra", "Annam", "Mikey", "Karim", "Ahmed"];
+var longNameThatStartsWithA = names.find(function(name) {
+  return name[0] === "A" && name.length > 7;
```
martynamaron: That's right! :)

week-3/B-array-some/exercise.js
```diff
@@ -16,6 +16,9 @@ var students = ["Islam", "Lesley", "Harun", "Rukmini"];
 var mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
 
 var pairs = pairsByIndex.map(function(indexes) {
+  if (indexes === null) {
+    process.exit(1);
```
martynamaron: Well, that's clever! Indeed this is doing what the exercise was asking :) However, I'd like you to not alter this code at all, and write your function on line 10. Please use the array method `some()` as well, thanks!

week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,9 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = students.every(function(student) {
+  return group.includes(student) !== true;
```
martynamaron: I think you had a good idea, but what you're doing here is that you're checking if all students are included in the `group` array. What you need to check is if the `group` array contains only students - subtle difference :D

week-3/D-array-filter/exercise.js
```diff
@@ -8,7 +8,9 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-var pairsByIndex; // Complete this statement
+var pairsByIndex = pairsByIndexRaw.filter(function(pair) {
+  return pair !== null && pair.length === 2;
```
martynamaron: Great! 🎉

week-3/E-array-map/exercise.js
```diff
@@ -2,4 +2,7 @@
 // Write multiple solutions using different syntax (as shown in the README)
 
 var numbers = [0.1, 0.2, 0.3, 0.4, 0.5];
-
+var multipliedBy = numbers.map(number => {
+  return number * 100;
```
martynamaron: Well done!
The instructions mention trying the same thing with different syntax - can you add those in as well please?

week-3/F-array-forEach/exercise.js
```diff
@@ -8,7 +8,17 @@
...
+  if (num % 5 === 0 && num % 3 === 0 && num !== 0) {
+    console.log("FizzBuzz");
+  } else if (num % 5 === 0 && num !== 0) {
+    console.log("Buzz");
+  } else if (num % 3 === 0 && num !== 0) {
+    console.log("Fizz");
+  } else {
+    console.log(num);
+  }
+});
```
martynamaron: Good! Well done for checking the `0` as well! 
A more efficient way of doing this though would be to nest the if statements - you can start with checking if it's a `0`, and if it's not - proceed with all the other checks. Try that :)

martynamaron:
Hey @ZaimAbz well done on the class exercises - left a few comments for improvement :)

Also, I see the exercises you posted in `week-3-practice` are the exercises you've written and posted in Slack and tagged Zareen in. So those are for Zareen to solve - you should be adding here the exercises you've received from someone else - including the solutions. Who gave you exercises? Could you solve them here please? Thanks!

week-3/E-array-map/exercise.js
ZaimAbz: sorry but not sure what do you mean what I need to do

martynamaron: Look at the README.md file in the E-array-map folder - there is a "Callback functions" section in there - there are examples of writing the same function in different ways. Please add those to your exercise here.

PR #255, opened 2019-01-17
Madiha - Java script week 3

diya2214:


week-3/A-array-find/exercise.js
```diff
@@ -1,13 +1,17 @@
...
+/*
+  You are given an array of names.
+  Using .find(), we'd like to find the first name which starts with A and is longer than 7 letters.
 */
 
-var testScores = [90, 50, 100, 25, 81, 66, 80];
-var lowestPassScore = findLowestPassScore(testScores);
+// write your code here
+function findLongNameThatStartsWithA(name){
+ return name.length >7 && name[0]==="A"
```
40thieves: Well done!

There is a little problem though, what happens if we change the `names` array to have "alexandra" instead of "Alexandra"? How can we fix that?

week-3/B-array-some/exercise.js
```diff
@@ -6,17 +6,34 @@
...
 // process.exit(1);
 
 var students = ["Islam", "Lesley", "Harun", "Rukmini"];
 var mentors = ["Daniel", "Irina", "Mozafar", "Luke"];
-
+// /////////////checking for null value in array
+var  isNull = pairsByIndex.some(checkNull);
+if (isNull) {
+ process.on('exit', function (code){
+  console.log("Null value found , terminating program code:"+ code);
```
40thieves: Well done! This is some very nice error handling 🍰

week-3/B-array-some/exercise.js
```diff
@@ -6,17 +6,34 @@
   - Do not edit any of the existing code
 */
 
-var pairsByIndex = [[0, 3], [1, 2], [2, 1], null, [3, 0]];
+function checkNull(index){
+  //console.log(index);
+  return index==null || index == "";
```
40thieves: The exercise description is a little vague, but I think it means to only check for values that are `null`. That is slightly different from an empty string.

week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,11 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = group.every(function (name) {
+  return students.find(function (stName){
+    return name === stName
```
40thieves: This is correct!

There is a shorter solution though. Remember that the `.includes()` method will check if the array (`students` in this case) includes a particular value (`name` in this case). Can you try implementing that solution?

week-3/D-array-filter/exercise.js
```diff
@@ -8,7 +8,11 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-var pairsByIndex; // Complete this statement
+var pairsByIndex = pairsByIndexRaw.filter(function(pairs){
+   if (Array.isArray(pairs) && pairs.length==2){
```
40thieves: Really nice work using `Array.isArray`! That's one of the JavaScript weirdnesses that I talked about last week 🙂

week-3/D-array-filter/exercise.js
```diff
@@ -8,7 +8,11 @@
...
+    return pairs;
```
40thieves: This will technically work, but most of the time with `.filter()` we want to return a boolean. Here we're returning the element of the array instead. How could we fix this?

week-3/E-array-map/exercise.js
```diff
@@ -1,5 +1,21 @@
...
+//************Method 2 **************/
+var newNumbers2 = numbers.map(function(number){
+    return number*100})
+console.log(newNumbers2);
 
+////*******************Method 3 */
+ var newNumber3 = numbers.map(number=>{
+     return number*100
+ })
+ console.log(newNumber3);
```
40thieves: Excellent work! All three solutions are correct 🎆

week-3/F-array-forEach/exercise.js
```diff
@@ -9,6 +9,13 @@
 
 var arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
 
+arr.forEach(number=>{
+  /********** using ternary operators to return either an empty string or 
+  Fizz / Buzz  concatinating result and addind OR if cancatination returns empty string*/
+  console.log((number%3?"":"Fizz")+(number%5?"":"Buzz")||number)
```
40thieves: This isn't quite right. Have a careful look at the expected output below, especially for the values 10 and 15 in the `arr` array. Are they the same in your output? How could we fix that?

Generally, in this situation I would usually use an `if` statement instead of a ternary. It makes it a little easier to read.

week-3/A-array-find/exercise.js
diya2214: by checking name[0]==="a"..... Will fix it thanks

week-3/C-array-every/exercise.js
diya2214: ohhh yes i was not clear which function returns what but now i revised them all ..... much clear concept now .... Thanks

week-3/F-array-forEach/exercise.js
diya2214: i am getting output of Buzz for 10 and FixxBuzz for 15 
which i think is correct in given scenario 10 is not a multiple of 3

PR #256, opened 2019-01-18
mohbeen - first attempt at week-3

mohbeen29:


week-3/A-array-find/exercise.js
```diff
@@ -1,13 +1,16 @@
...
+// You are given an array of names.
+//   Using.find(), we'd like to find the first name which starts with A and is longer than 7 letters.
 
-var testScores = [90, 50, 100, 25, 81, 66, 80];
-var lowestPassScore = findLowestPassScore(testScores);
+// write your code here
 
-console.log(lowestPassScore);
+function findLongnameThatStartsWithA(names) {
+  return names.slice[0] === "A" && names.length >= 7;
```
martynamaron: `slice` is a method, so it needs round brackets `()`. Square brackets `[]` are only used with an index of an array. Helpfully, a string is treated in JavaScript as an array of letters (or characters). How can you get the first element of that string then?

Just to be clear `names.slice[0]` is not valid code - you get "Alexandra" as output because it's the only one that satisfies the `names.length >= 7` condition :) Please try again!

week-3/C-array-every/exercise.js
```diff
@@ -5,7 +5,7 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = students.every(group); // complete this statement
```
martynamaron: I'm afraid this code doesn't run :( `group` is not a function, it's an array. The `every()` method needs a function (or the name of a function) as parameter. That function needs to return a boolean based on a certain check. Here this check would be to see if the element of the array on which the method is being called on belongs in the `students` array. 

The interpretation of the task is: check if every element in the group array is included in the students array.

Hope this helps! :)

week-3/D-array-filter/exercise.js
```diff
@@ -8,7 +8,11 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-var pairsByIndex; // Complete this statement
+var pairsByIndex = pairsByIndexRaw.filter(function(item) {
+  if (item !== null || item !== boolean || item !== "string") {
```
martynamaron: `item !== boolean` - I think you meant to check the `typeof` that item? Same for the "string" check. Remember that `typeof item` will always return a string.

Also, the function that is passed into the array method `filter` has to return a boolean, not the item :)

Good syntax though!

week-3/E-array-map/exercise.js
```diff
@@ -3,3 +3,22 @@
 
 var numbers = [0.1, 0.2, 0.3, 0.4, 0.5];
 
+function multiply(number) {
+  return number * 100;
+}
+
+var numbersMultiplied = numbers.map(multiply);
```
martynamaron: Yeah! That's right! 👏

week-3/E-array-map/exercise.js
```diff
@@ -3,3 +3,22 @@
...
+
+//or
+
+var numbersMultiplied = numbers.map(function multiply(number) {
```
martynamaron: Yep, also correct! :) 
Be careful with declaring 2 variables with the same name though - they might clash.

week-3/E-array-map/exercise.js
```diff
@@ -3,3 +3,22 @@
...
+  return number * 100;
+});
+console.log(numbersMultiplied);
+//or
+var numbersMultiplied = numbers.map(number => {
+  return number * 100;
+});
+
+//or
+var numbersMultiplied = numbers.map(number => number * 100);
```
martynamaron: Great! Good syntax :)

week-3/F-array-forEach/exercise.js
```diff
@@ -6,9 +6,47 @@
 
   An array with numbers 0-15 has been provided.
 */
+//create a function that performs on each element in the array.
 
 var arr = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15];
 
+arr.forEach(multiples(item){
```
martynamaron: I think you're missing an important keyword on this line (line 13) :) Look at your E-array-map/exercise.js to help you figure out what it is.

week-3/F-array-forEach/exercise.js
```diff
@@ -6,9 +6,47 @@
...
+} else if (item %5 === 0){
+  return "Buzz";
+} else if (item %3 === 0 && item %5 === 0){
+  return "FizzBuzz";
+} else {
+  return item;
+}
+});
+
+console.log(item);
```
martynamaron: `item` is not reachable from this line - it's enclosed in the `forEach` function - so I'm afraid this code doesn't run because `item` is undefined.

week-3/F-array-forEach/exercise.js
```diff
@@ -6,9 +6,47 @@
...
 
+arr.forEach(multiples(item){ 
+if (item %3 === 0){
+  return "Fizz";
+} else if (item %5 === 0){
+  return "Buzz";
+} else if (item %3 === 0 && item %5 === 0){
+  return "FizzBuzz";
+} else {
+  return item;
```
martynamaron: I think your checks are good here, but here's a couple of points:
- you don't have to `return` anything - you can just print it to the console
- I suggest starting with the "strictest" of the checks - for example with the number `15` it's meant to print "FizzBuzz". But if it encounters the `item % 3 === 0` check first, it'll print "Fizz". So the checks just need a reshuffle :)

martynamaron:
Hey Mohbeen - I think the ideas are there, but I think maybe you've forgotten the syntax :) I've left some tips, so maybe you could try again? Also, I think you're missing the `B-array-some` exercise?
Please if you need help, don't hesitate to ask in the channel or work with your study group! 🙌

mohbeen29:
Hey martyna,
I did this on tuesday as i needed guidance to complete it altho i pushed i
forgot to pull. My mistake i was worried it got deleted as i couldnt see it
on github.
Thank you for reviewing apolgies its late.

PR #266, opened 2019-01-21
Ivina js week4

ivinapontes:


week-4-practise/level-2/exercise-1.js
```diff
@@ -9,17 +9,19 @@ We made a start on this for you here:
...
-		// this function should return true if your bottles volume = 0
-	}
-}
+  volume: 0,
+  fill: function() {
+    return (this.volume = 100);
+  },
+  drink: function() {
+    return (this.volume = this.volume - 10);
+  },
```
nbogie: `fill` and `drink` don't need to return anything, it's only important here that they have the right side-effect - changing the value of the bottle's `volume` property.  They do - good!

week-4-practise/level-2/exercise-1.js
```diff
@@ -9,17 +9,19 @@ We made a start on this for you here:
...
+  empty: function() {
+    if (this.volume <= 0) {
+      return true;
+    }
+  }
```
nbogie: There's a small bug here, waiting to bite you.  

What will your code return if the volume is greater than zero?  

When you see `if (condition) { return }` but no `else`, this should ring alarm bells in your head because there's a path through your code which doesn't explicitly return anything.  It's not *always* wrong, but mostly.

week-4-practise/level-1/exercise-1.js
```diff
@@ -38,7 +38,9 @@ var writers = [
     alive: true
   },
 ];
-
+writers.forEach(function (writer) {
+  console.log("Hi, my name is " +writer.firstName+ " " +writer.lastName+". I am "+writer.age+ " years old, and work as a "+writer.occupation);
+});
```
nbogie: Good.  When we later cover template strings, you'll see we can alternatively write this:

```
console.log(`Hi, my name is ${writer.firstName} ${writer.lastName}.  I am ${writer.age} years old, and work as a ${writer.occupation}`);
```

week-4/A-objects-intro/exercise-part-2.js
```diff
@@ -5,17 +5,18 @@ The objects below have some syntax issues - try and fix them all!
...
-    operating system "iOS",
-    hasStylus: true,
-    megapixels 12
-    "batteryLife": "24 hours"
\ No newline at end of file
+  operatingSystem: "iOS",
+  hasStylus: true,
+  megapixels: 12,
+  batteryLife: "24 hours"
+};
```
nbogie: all good

week-4/B-objects-get-set/exercise-2.js
```diff
@@ -5,14 +5,15 @@
...
-    is Unlocked: true
-;
+    isUnlocked: true
+}
 
-var phoneBrand = phone.bbrand;
-var phoneLaunchYear = phone[launchYear];
+
+var phoneBrand = phone.brand;
+var phoneLaunchYear = phone['launchYear'];
```
nbogie: yep

week-4/B-objects-get-set/exercise-4.js
```diff
@@ -3,21 +3,21 @@
...
-    wantsToPlay: false
+  name: "Billy",
+  wantsToPlay: false
 };
 
 // WRITE CODE BELOW THIS LINE
 
-
+dog.name = "Rex";
+dog.wantsToPlay = true;
```
nbogie: good

week-4/C-more-complex-objects/exercise-1.js
```diff
@@ -22,12 +22,20 @@ var house = {
 // - change the address of "house" to '51 Berkley Road'
 // - change the previous owners of "house" to ["Brian M.", "Fiona S."]
 // - change the last name of the current owner of "house" to "Montgomery"
-
-
+house.address = "51 Berkley Road";
+house.previousOwners = ["Brian M.", "Fiona S."];
+house.currentOwner.lastName = "Montgomery";
 /*
```
nbogie: right!

week-4/C-more-complex-objects/exercise-2.js
```diff
@@ -28,15 +28,22 @@ var newCurrentOwner = {
 // - assign the value of the variable 'newCurrentOwner' as the value to the house's "currentOwner"
 // - from the list of previous owners, replace only "John A." with "Stephen B."
 // - give the house a new property called 'isForSale' with the value 'false'
-
-
-
-
+newCurrentOwner.firstName = "Margaret";
```
nbogie: We want Georgina Hernandez to be the `house.currentOwner`.  Also, no need to assign it string by string.

week-4/C-more-complex-objects/exercise-2.js
```diff
@@ -28,15 +28,22 @@ var newCurrentOwner = {
...
 // - from the list of previous owners, replace only "John A." with "Stephen B."
 // - give the house a new property called 'isForSale' with the value 'false'
-
-
-
-
+newCurrentOwner.firstName = "Margaret";
+newCurrentOwner.lastName = "Conway";
+newCurrentOwner = house.currentOwner;
+house.previousOwners = ["Claire M.", "Stephen B."];
```
nbogie: `replace ONLY "John A." with "Stephen B."`  Can you think of a way to do this?

week-4/C-more-complex-objects/exercise-3.js
```diff
@@ -32,23 +32,43 @@ var parkAvenueHouse = {
 
 // returns the full name (first name + last name) of the owner of the house
 function getOwnerFullName(house) {
-
+  return (
+    kinningParkHouse.currentOwner.firstName +
```
nbogie: bug.  This should return full name of the owner of the house passed in as the parameter `house`!  Currently it ALWAYS does `kinningParkHouse's owner's name, no matter what it is passed!

week-4/C-more-complex-objects/exercise-3.js
```diff
@@ -32,23 +32,43 @@ var parkAvenueHouse = {
...
+    " " +
+    kinningParkHouse.currentOwner.lastName
+  );
 }
 
 // returns an array of the owners' email addresses of the two houses
 function getEmailAddresses(house1, house2) {
-
+  return (
+    kinningParkHouse.currentOwner.email +
```
nbogie: bug: This should work on house1, and house2 - and you shouldn't have to know what they're going to be in advance!  

The syntax is good, though.

week-4/C-more-complex-objects/exercise-3.js
```diff
@@ -32,23 +32,43 @@ var parkAvenueHouse = {
...
 }
 
 // returns the address for the cheapest house out of the two
 function getCheapestAddress(house1, house2) {
-
+  if (kinningParkHouse.price < parkAvenueHouse.price) {
+    return kinningParkHouse.address;
+  } else {
+    return parkAvenueHouse.address;
+  }
```
nbogie: good logic, but again, test `house1` and `house2` parameters and return one of *them*.  

You don't know if people will always call your function with kinningParkHouse and parkAvenueHouse - indeed that's why we have parameters!

week-4/D-methods/exercise-2.js
```diff
@@ -4,15 +4,18 @@ Add a method "sayName" so this person can say their own name.
...
 
-
 var person = {
-    name: "Alice",
-    age: 25
+  name: "Alice",
+  age: 25,
+  sayName: function() {
+    return "Hi, my name is " + this.name;
+  }
```
nbogie: good

week-4/D-methods/exercise-3.js
```diff
@@ -3,26 +3,26 @@ The following code contains syntax errors - try and fix them!
...
-    }
+  name: "Alice",
+  age: 25,
+  currentAddress: "Glasgow",
+  changeAddress: function(newAddress) {
+      this.currentAddress = newAddress;
+  },
+  celebrateBirthday: function() {
+    this.age = this.age + 1;
+  }
```
nbogie: good!  we will see this a LOT more.

week-4/D-methods/exercise-4.js
```diff
@@ -6,7 +6,10 @@ Define a method "makeFriend" to add a new friend to her list.
 
 var person = {
     name: "Alice",
-    friends: ["John", "Nina"]
+    friends: ["John", "Nina"],
+    makeFriend: function (newFriend) {
+        this.friends.push(newFriend);
+    }
```
nbogie: great.

week-4/D-methods/exercise-5.js
```diff
@@ -9,35 +9,54 @@ only if the inserted amount is greater or equal than the price of the coffee!
...
+  brand: "Super Coffee",
+  prices: {
+    cappuccino: 2.4,
+    blackCoffee: 1.5,
+    flatWhite: 3.0
+  },
+  insertedAmount: 0,
+  insertMoney: function(amount) {
+      return this.insertedAmount = amount;
+
```
nbogie: insertMoney: good!

week-4/D-methods/exercise-5.js
```diff
@@ -9,35 +9,54 @@ only if the inserted amount is greater or equal than the price of the coffee!
...
+    blackCoffee: 1.5,
+    flatWhite: 3.0
+  },
+  insertedAmount: 0,
+  insertMoney: function(amount) {
+      return this.insertedAmount = amount;
+      
+  },
+    getCoffee: function (coffee) {
+        if (coffee === "cappuccino" && this.insertedAmount >= this.prices.cappuccino) {
```
nbogie: Looks like this works (good!), but there's a way we can do this with many fewer lines of code, instead of a big if... else if... construction.  

Imagine there were 40 different types of product being vended, or 4000... We couldn't keep adding to the if... else if... without going insane.  
See if you can improve it.

week-4/E-arrays-of-objects/exercise-1.js
```diff
@@ -5,42 +5,65 @@
...
+}
 
-var personsYoungerThan28YearsOld = // Complete here
+//[person1.name, person2.name, person3.name]; // Complete here
 
+var personsYoungerThan28YearsOld = persons.filter(checkingAge); // Complete here
+
+function checkingAge(persons) {
+  return persons.age <= 28;
+}
```
nbogie: *Younger than*.  28 is not younger than 28...  Always be careful with inequalities and check whether they're inclusive or exclusive.  Many bugs hide here.

week-4/E-arrays-of-objects/exercise-2.js
```diff
@@ -7,50 +7,92 @@ Each destination has a name, a distance from Glasgow, and a list of transportati
...
+  destination3,
+  destination4
+];
 
 /* 
 DO NOT EDIT ANYTHING ABOVE THIS LINE
 WRITE YOUR CODE BELOW
 */
+//Question 1
+// Find retorna só o primeiro que satisfaz a condiçao. Filter retorna mais!
```
nbogie: É assim!

week-4/E-arrays-of-objects/exercise-2.js
```diff
@@ -7,50 +7,92 @@ Each destination has a name, a distance from Glasgow, and a list of transportati
...
+
+function checkingDistance(travelDestinations) {
+  if (travelDestinations.distanceKms <= 500) {
+    return travelDestinations;
+  }
+}
```
nbogie: Some problems here.  
checkingDistance() only ever takes ONE object at a time.  YES, it will be called repeatedly by filter, but each time it is called, it takes ONE travel destination, so it's parameter should not be plural - it does not take an array of travelDestinations.

Secondly, what you want to return for a function used by filter (a "predicate function") is simply true or false.  You can (and should) convince yourself of this by reading the official docs: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter#Syntax
"Function is a predicate, to test each element of the array. Return true to keep the element, false otherwise."

Thirdly, you again have a path through your code where you probably can't tell me what will be returned.

Fourthly, there's an easier way to do this without using an `if` statement at all.

week-4/E-arrays-of-objects/exercise-2.js
```diff
@@ -7,50 +7,92 @@ Each destination has a name, a distance from Glasgow, and a list of transportati
...
+function getName(travelDestinations) {
+  return travelDestinations.destinationName;
+}
```
nbogie: good.  again this takes ONE destination, not an array, so `travelDestinations` is a misleading name for the parameter.

week-4/E-arrays-of-objects/exercise-2.js
```diff
@@ -7,50 +7,92 @@ Each destination has a name, a distance from Glasgow, and a list of transportati
...
+//Question 2
 
+function isItReachableByFerry(travelDestinations) {
+  return travelDestinations.transportations.includes("ferry");
+}
 
-var destinationNamesWithin500Kms = // Complete here
+var destinationNameReachableByFerry = travelDestinations
+  .filter(isItReachableByFerry)
+  .map(getName);
```
nbogie: I think maybe you're supposed to use `find` here - as it's destinationName... not destinationNames... 
minor issue.

week-4/E-arrays-of-objects/exercise-2.js
```diff
@@ -7,50 +7,92 @@ Each destination has a name, a distance from Glasgow, and a list of transportati
...
+function destinationMoreThan300andByTrain(travelDestinations) {
+  return (
+    travelDestinations.transportations.includes("train") &&
+    travelDestinations.distanceKms > 300
+  );
+}
 
+var destinationNamesMoreThan300KmsAwayByTrain = travelDestinations
+  .filter(destinationMoreThan300andByTrain)
+  .map(getName);
```
nbogie: Good.  

Minor point: again, your predicate function takes one parameter, so better to call it `destination`.  After all, the statement `travelDestinations.distanceKms > 300` makes no sense if travelDestinations is really multiple.

week-4/E-arrays-of-objects/exercise-3.js
```diff
@@ -53,33 +53,62 @@ var restaurants = [restaurant1, restaurant2, restaurant3];
...
-    }
+  applicationName: "Restaurant Finder",
+  applicationVersion: "1.0",
+  restaurants: restaurants,
+  findAvailableRestaurants: function(numberOfPeople) {
+    return restaurants.filter( function getRestAvailable(rest) {
+      return rest.totalSeats - rest.numberOfCustomers >= numberOfPeople;
+      })
+      .map(getRestaurantName);
+  },
```
nbogie: good, this will work, but you have getRestAvailable defined twice, and are only using one.  

There is also no need to name the function you are creating and passing in-line to filter on line 68.

More importantly, note that the `restaurants` variable referred to on line 68 is NOT the restaurants property of this object!  It is the global `restaurants` variable.   This is probably not what you meant to do.

week-4/E-arrays-of-objects/exercise-3.js
```diff
@@ -53,33 +53,62 @@ var restaurants = [restaurant1, restaurant2, restaurant3];
...
+  findRestaurantServingDish: function(dishName) {
+    return restaurants.filter(function(restaurant) {
+        return restaurant.menu.includes(dishName);
+      })
+      .map(getRestaurantName);
+  },
```
nbogie: good.  We'll learn about arrow functions in ES6 which make this more readable: 
```
findRestaurantServingDish: function(dishName) { 
   return restaurants.filter(r => r.menu.includes(dishName))
               .map(getRestaurantName);
}
```

week-4/E-arrays-of-objects/exercise-3.js
```diff
@@ -53,33 +53,62 @@ var restaurants = [restaurant1, restaurant2, restaurant3];
...
+  countNumberOfRestaurantsInArea: function(area) {
+      return restaurants.filter(function(restaurant){
+          return restaurant.address.area === area
+      }).length;
+      }
+
+
```
nbogie: good

week-4/F-object-keys/exercise-part-1.js
```diff
@@ -15,10 +16,13 @@ var mentorsAges = {
...
 
-var mentorsNames = ;
+var mentorsNames = Object.keys(mentorsAges) ;
 
-var mentorsNamedCapitalised = ;
+var mentorsNamedCapitalised = mentorsNames.map(getName); ;
 
+function getName(mentorsName) {
+  return mentorsName.toUpperCase();
+}
```
nbogie: `getName` isn't the best name for this function, because it does something more than simply getting the name.

week-4-practise/level-2/exercise-1.js
ivinapontes: Thanks for pointing it out!!

week-4/C-more-complex-objects/exercise-2.js
ivinapontes: Aham, I've changed the name by assigning the array index.

week-4/E-arrays-of-objects/exercise-2.js
ivinapontes: hahaha yes! :smile:

PR #269, opened 2019-01-22
BartC-Week4

bmc82:


week-4-practise/level-2/exercise-1.js
```diff
@@ -35,7 +35,7 @@ bottle.fill();
...
 bottle.drink();
 bottle.drink();
-if(!bottle.empty()){
-	console.log('bottles volume = ' + bottle.volume);
+if (!bottle.empty()) {
+  console.log("bottles volume = " + bottle.volume);
 }
-console.log("Above volume should be: 70");
\ No newline at end of file
+console.log("Above volume should be: 70");
```
timrooke1991: This does what it needs to - and good effort! 👍 

However, there are some additional considerations to look at here. The explanation says: `It will be 100 when full and 0 when empty`. What happens if drink a lot without re-filling? Does the bottle remain empty? Also, what if I fill the bottle twice? What is volume number?

How about looking at these edge cases? 🙂

PR #272, opened 2019-01-22
Madiha Week 4 JS

diya2214:


week-4/E-arrays-of-objects/exercise-3.js
```diff
@@ -60,14 +60,19 @@ var restaurantFinderApplication = {
...
-        // Complete here
+        var dishInResturant = restaurants.filter(dish=>dish.menu.includes(dishName)).map(rest=>rest.name);
+        return dishInResturant;
     },
     countNumberOfRestaurantsInArea: function(area) {
-        // Complete here
-    }
+     var restInGlasgow  = restaurants.filter(gArea=>gArea.address.area === area).length;
+     return restInGlasgow;
+      }
```
timrooke1991: Great job - this runs smoothly and returns what we want. 👏

Like before, I think we are putting too much on one line here. And also, we have some issues with variable names. 

- `dish` isn't ideal because you then write: `dish.menu`, which doesn't make sense. A menu has dishes, a dish doesn't have a menu. At that point in time, you'll dealing with an array of restaurants even if you objective is to reference the dish eventually. 
- `gArea` - what does this mean? I guess short for glasgow area? I would avoid short acronyms that might not be obvious to other developers working with your code. Better be verbose and obvious than short and obscure.

I don't want to sound like I am repeating myself, but these are important things to think about. When you start programming with other developers, they will need to be able to understand what is going on with your code 😊 That being said, the code works great! 👍

week-4/F-object-keys/exercise-part-0.js
```diff
@@ -20,8 +20,8 @@ var highScores = {
 
 // ONLY EDIT BELOW HERE
 
-var capitalCitiesKeys = ;
-var highScoresKeys;
+var capitalCitiesKeys = Object.keys(capitalCities);
+var highScoresKeys = Object.keys(highScores);
 
 // ONLY EDIT ABOVE HERE
```
timrooke1991: Great 😃

week-4/F-object-keys/exercise-part-1.js
```diff
@@ -15,11 +15,8 @@ var mentorsAges = {
...
-
-var mentorsNamedCapitalised = ;
-
-// ONLY EDIT ABOVE THIS LINE
-
+var mentorsNames = Object.keys(mentorsAges);
+console.log(mentorsNames);
+var mentorsNamedCapitalised = mentorsNames.map(name=>name.toUpperCase());
 console.log(mentorsNamedCapitalised);
 // prints [ 'JAMES', 'JOSH', 'JAMIE', 'MOZAFAR' ]
```
timrooke1991: `name` is a perfect variable name 😊

week-4/F-object-keys/exercise-part-2.js
```diff
@@ -35,14 +35,14 @@ var storeBranches = {
...
 // prints [ 'manager', 'assistant', 'interns' ]
-console.log()
+console.log(Object.keys(storeBranches.glasgow))
 
 // # 3
 // prints [ 'head_intern', 'intern' ]
-console.log()
+console.log(Object.keys(storeBranches.edinburgh.interns))
 
 // ONLY EDIT ABOVE THIS LINE
```
timrooke1991: Perfect! 🙌

PR #276, opened 2019-01-24
Sola-exercise-week4

solasobowale:


week-4-practise/level-1/exercise-1.js
```diff
@@ -36,9 +36,45 @@ var writers = [
...
     alive: true
-  },
+  }
 ];
+var livingWriters = [];
+
+var writer = "writer";
+
+function isAlive(writer) {
+  return writers
```
foaly-nr1: There is a small difference in spelling between the `writers` variable you're using in line 46 and the `writer` parameter you've defined in line 45. Therefore you're *not* using the value that is passed into the function but instead you're accessing a variable from outside the function scope, called the global scope. Try to avoid accessing any values from the global scope and instead pass all values a function needs in as an argument. Hope this makes sense?

week-4-practise/level-1/exercise-1.js
```diff
@@ -36,9 +36,45 @@ var writers = [
     occupation: "writer",
     age: 64,
     alive: true
-  },
+  }
 ];
+var livingWriters = [];
+
+var writer = "writer";
```
foaly-nr1: This variable doesn't seem to be required 🙂

week-4-practise/level-1/exercise-1.js
```diff
@@ -36,9 +36,45 @@ var writers = [
...
+  }
 ];
+var livingWriters = [];
+
+var writer = "writer";
+
+function isAlive(writer) {
+  return writers
+    .filter(
+      authour => authour.occupation === "writer" && authour.alive === true
```
foaly-nr1: You said in class you couldn't figure out how to only test `alive`. Try this: ```.filter(author => author.alive === true)```

week-4-practise/level-1/exercise-1.js
```diff
@@ -36,9 +36,45 @@ var writers = [
...
+    " " +
+    livingWriters[0].lastName +
+    ". " +
+    "I am " +
+    livingWriters[0].age +
+    " years old, and work as a " +
+    livingWriters[0].occupation +
+    "."
+);
+console.log(
```
foaly-nr1: You're repeating a lot of code in these two calls of `console.log()`. Image you had 20 more writers that were alive – you'd have to add so much code! Can you think of a way to make this more DRY (don't repeat yourself)? Maybe use a function.

week-4-practise/level-2/CODEWARS.md
```diff
@@ -3,4 +3,33 @@ Complete the following CodeWars exercises. Go to the webpages below and follow t
...
+latvian: 'Gaidits',
+lithuanian: 'Laukiamas',
+polish: 'Witamy',
+spanish: 'Bienvenido',
+swedish: 'Valkommen',
+welsh: 'Croeso'
+};
+
+function greet(language) {
+if (audience.hasOwnProperty(language)) {
```
foaly-nr1: As above, don't access variables from outside of the function that aren't passed in as an argument. Instead, you could define the `audience` variable inside of `greet()`.

week-4-practise/level-2/exercise-1.js
```diff
@@ -7,19 +7,22 @@ drinking some of it, and emptying it.
...
-	drink: function(){
-		// calling this function should decrease your bottles volume by 10;
-	},
-	empty: function(){
-		// this function should return true if your bottles volume = 0
-	}
-}
+  volume: 0,
+  fill: function() {
+    return (bottle.volume = 100);
```
foaly-nr1: When calling `fill()` on your `bottle` object what return type would you expect? Would you expect any at all?

week-4/A-objects-intro/exercise-part-1.js
```diff
@@ -6,4 +6,23 @@ can describe with a JavaScript object
...
+  name: "alice"
+};
+var myCat = {
+  colour: "grey",
+  type: "tabby",
+  age: 3,
+  cuddly: true
+};
 
+var mySibblings = {
```
foaly-nr1: Typically, when you have a number of siblings with their various properties, you'd use an array of objects. That would look like this:
```
var mySiblings: [
  {
    name: "Toyin",
    gender: "F",
    location: "USA",
   },
   ...
];
```

week-4/D-methods/exercise-3.js
```diff
@@ -3,26 +3,26 @@ The following code contains syntax errors - try and fix them!
...
-        currentAddress = newAddress;
-    },
-    celebrateBirthday: function {
-        that.age = that.age + 1;
-    }
+  name: "Alice",
+  age: 25,
+  currentAddress: "Glasgow",
+  changeAddress: function(newAddress) {
+    return (this.currentAddress = newAddress);
```
foaly-nr1: Does this function need to return something at all? What would you use the returned value for?

week-4/E-arrays-of-objects/exercise-1.js
```diff
@@ -5,42 +5,55 @@
...
+var persons = [person1, person2, person3]; // Complete here
 
-var persons = // Complete here
-
-var personNames = // Complete here
-
-var personsYoungerThan28YearsOld = // Complete here
+var personNames = persons.map(person => person.name); // Complete here
 
+var personsYoungerThan28YearsOld = persons.filter(person => person.age < 28); // Complete here
```
foaly-nr1: Good use of `filter()` and the arrow function here 👏🏻

week-4/E-arrays-of-objects/exercise-2.js
```diff
@@ -7,50 +7,87 @@ Each destination has a name, a distance from Glasgow, and a list of transportati
...
 DO NOT EDIT ANYTHING ABOVE THIS LINE
 WRITE YOUR CODE BELOW
 */
+/*function getDestination(destination) {
+  if (destination.distanceKms < 500) {
+    return destination.destinationName;
+  }
+} */
 
+function getDestination(destination) {
```
foaly-nr1: Can you think of a more meaningful name for this function that would help me guess/understand what the function returns?

PR #291, opened 2019-01-27
Ahmet-JS-Week5

Ahmet-K:


Ahmet-K:
Ahmet-JS-Week-5

week-5-practise/js/main.js
```diff
@@ -0,0 +1,66 @@
+//// PART-1 ////
+
+var placeToJumbotron = document.querySelector(".jumbotron");
+var placeToDonateABike = document.querySelector(".btn-primary"+".btn-lrg");
```
foaly-nr1: Just a minor detail: What's the `"+"` for? I think it would be easier to read without it.

week-5-practise/js/main.js
```diff
@@ -0,0 +1,66 @@
...
+var placeToVolunteer = document.querySelector(".btn-secondary");
+
+var blueButton = document.querySelector("#blueBtn");
+blueButton.addEventListener("click", blueButtonFunction);
+function blueButtonFunction() {
```
foaly-nr1: Can you think of a more meaningful name for your function? Usually your function names should describe what the function does. In this case it's changing colours, right?

week-5-practise/js/main.js
```diff
@@ -0,0 +1,66 @@
...
+    placeToJumbotron.style.backgroundColor = "#87ca8a";
+    placeToDonateABike.style.backgroundColor = "black";
+    placeToVolunteer.style.backgroundColor = "#8c9c08";
+}
+
+////////// PART-2 /////
+
+var formTag = document.querySelector("form");
+formTag.lastElementChild.addEventListener("click", submitButtonFunction); 
+event.preventDefault();
```
foaly-nr1: Did this give you a runtime error? I can't see `event` being declared anywhere. Let me know what you were trying to achieve with this line of code and I can help you out.

week-5-practise/js/main.js
```diff
@@ -0,0 +1,66 @@
...
+    if (nameInput.value.length === 0) {
+        nameInput.style.backgroundColor = "red";
+    } else{
+        nameInput.style.backgroundColor = "white";
+    }
+    if (describeTextInput.value.length === 0) {
+        describeTextInput.style.backgroundColor = "red";
+    } else{
+        describeTextInput.style.backgroundColor = "white";
+    }
```
foaly-nr1: This is really good, Ahmet! 👏🏻 Well done for also turning the fields back to white once they have been filled in.

week-5-practise/js/main.js
```diff
@@ -0,0 +1,66 @@
...
+    if ((emailInput.value.length !== 0) && (emailInput.value.includes("@")) && (nameInput.value.length !== 0) && (describeTextInput.value.length !== 0)){
```
foaly-nr1: This works perfectly fine. As a best practice, when you have more than one negative condition in an if statement, you would work with a negative group of conditions instead. This has the exact same outcome as your line of code but should be easier to read/comprehend:
```
iff (!(emailInput.value.length === 0 || !emailInput.value.includes("@") || nameInput.value.length === 0 || describeTextInput.value.length === 0)){
```
If this doesn't make a lot of sense don't worry and just ignore this comment.

week-5-practise/js/main.js
Ahmet-K: I am asked that "In your function that handles clicks on the `Submit` button you will need to call `event.preventDefault()` to stop the browser from refreshing the page." in the question. That's why I have used it.

week-5-practise/js/main.js
Ahmet-K: Thanks. You are right. I have gotten Naming issues.

week-5-practise/js/main.js
Ahmet-K: My code doesn't work without "+".

foaly-nr1: If you take off all three characters (`"+"`) it will work.

week-5-practise/js/main.js
foaly-nr1: In that case, you'll need to move it into the function. Outside the function, where it is now, you'll find that `event` is undefined.

PR #294, opened 2019-01-27
Miles - Week-4

milesuno:


week-4/A-objects-intro/exercise-part-1.js
```diff
@@ -7,3 +7,18 @@ Assign each of them to a separate variable
...
+};
+
+const gymStats = {
+    bench: "130 KG",
+    squat: "195 KG",
+    deadlift: "200 KG",
+    overHeadPress: "85 KG"
+};
+
+console.log(miles.nationality+gymStats.bench);
```
foaly-nr1: This line made me curious. Did it work? What happened? JavaScript does often allow to do operations with two values of different types with sometimes peculiar outcomes. Here are some of them if you're interested: https://charlieharvey.org.uk/page/javascript_the_weird_parts

week-4/C-more-complex-objects/exercise-3.js
```diff
@@ -32,18 +32,20 @@ var parkAvenueHouse = {
...
+  return house1.currentOwner.email +", "+ house2.currentOwner.email;
 }
 
 // returns the address for the cheapest house out of the two
 function getCheapestAddress(house1, house2) {
-
-}
+if(house1 < house2){
+  return "Buy " + house1.address;
+} return "Buy " + house2.address;
```
foaly-nr1: That's exactly right! And you don't need an `else` here. Just a minor detail: To make it easier to read your code, make sure to have a line break before the second `return` statement.

week-4/E-arrays-of-objects/exercise-1.js
```diff
@@ -19,17 +19,24 @@ var person3 = {
...
 DO NOT EDIT ANYTHING ABOVE THIS LINE
 WRITE YOUR CODE BELOW
 */
+function isName(person){
+    return person.name;
+}
+
+function isUnder28(person){
+    if(person.age < 28){
+     return person;
```
foaly-nr1: I know when you executed this it worked... to the dismay of most coders JavaScript often allows you to do things that weren't meant to work such as in this case. By definition, the callback used for `filter()` needs to return a boolean value: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter.

week-4/E-arrays-of-objects/exercise-3.js
```diff
@@ -60,7 +60,9 @@ var restaurantFinderApplication = {
     applicationVersion: "1.0",
     restaurants: restaurants,
     findAvailableRestaurants: function(numberOfPeople) {
-        // Complete here
+        if(numberOfPeople < this.restaurants.totalSeats){
+            return this.restaurants.name;
+        }// Complete here
     },
```
foaly-nr1: Not quite right. `this.restaurants` is an array so it won't have a `totalSeats` or `name` property.

week-4/F-object-keys/exercise-part-1.js
```diff
@@ -15,11 +15,11 @@ var mentorsAges = {
 
 // ONLY EDIT BELOW THIS LINE
 
-var mentorsNames = ;
+var mentorsNames = Object.keys(mentorsAges);
 
-var mentorsNamedCapitalised = ;
+var mentorsNamedCapitalised = mentorsNames.map(function(arr){return arr.toUpperCase()});
```
foaly-nr1: This is correct. Can you think of a more meaningful name for your callback argument? `arr` suggests it's an array you expect here.

PR #295, opened 2019-01-27
Mussie JS-week-5  

Mussie14:


week-5-practise/js/main.js
```diff
@@ -0,0 +1,59 @@
+//  ### Part 1
+
+// Blue Click
+
+var blueButton = document.querySelector("#blueBtn");
+
+var jumbo = document.querySelector(".jumbotron");
+var bike = document.querySelector(".jumbotron .btn-primary");
```
BexB1: Don't forget to use descriptive variable names! When you're working as part of a team of developers, you need to write code that makes sense not only to you. So it's good practice to keep things descriptive, even if you're working alone.

PR #296, opened 2019-01-27
Elamin-week-5-exercises

aldeez988:


BexB1:
Hi Elamin, this is very good work. Well done!

You did very well to write the validations for the form, that clearly took some research. The only thing I would add is that where you use a `for` loop, you can often use array methods like `forEach`. They're not always better to use, but it's worth doing some research to learn which you use and when: https://davidtang.io/2016/07/30/javascript-for-loop-vs-array-foreach.html

week-5-practise/js/main.js
```diff
@@ -0,0 +1,63 @@
+var blueButton = document.querySelector("#blueBtn");
+var jumbotron = document.querySelector(".jumbotron");
+var donateAndVolunteerButtons = document.querySelectorAll(".btn-lrg");
```
BexB1: Very nice use of descriptive variable names. 👍

PR #303, opened 2019-01-29
zareen js exercises week5

khan-zareen:


week-5/A-dom-manipulation/exercise.js
```diff
@@ -4,30 +4,46 @@ Task 1
 
 Write JavaScript below that changes the background colour of the page when the 'Change colour' button is clicked.
 */
-
-
+var updateTitleBtn = document.querySelector("#bgrChangeBtn");
+updateTitleBtn.addEventListener("click", function() {
+  var myElement = document.querySelector("#bgrChangeBtn");
```
40thieves: This line looks identical to line 7, does that tell us something? We already have a reference to the button, so if we wanted to change that then we can just change it 🙂. We don't need to select it again within the event handler.

However it looks like the exercise asks you to change the background colour of the *page*, not the button when the button is clicked.

week-5/A-dom-manipulation/exercise.js
```diff
@@ -4,30 +4,46 @@ Task 1
...
 ======
-
 When a user clicks the 'ALERT' button, an alert box should pop up with the text "Thanks for visiting Bikes for Refugees!"
 */
-
-
-
+var updateTitleBtn = document.querySelector("#alertBtn");
+updateTitleBtn.addEventListener("click", function() {
+  alert("Thanks for visiting Bikes for Refugees!");
```
40thieves: Correct! Well done!

week-5/A-dom-manipulation/exercise.js
```diff
@@ -4,30 +4,46 @@ Task 1
...
 */
-
-
-
+var updateTitleBtn = document.querySelector("#addTextBtn");
+updateTitleBtn.addEventListener("click", function() {
+  var myElement = document.querySelector(".buttons");
+  var myParagraph = document.createElement("p");
+  document.querySelector("#mainArticles").appendChild(myParagraph);
+  myParagraph.innerText = "Read more below.";
```
40thieves: I think this does work, but it does the operations in a strange order. We want to create the `myParagraph` element with all of it's text, then append it into the DOM.

So I would swap lines 32 and 33.

week-5/A-dom-manipulation/exercise.js
```diff
@@ -4,30 +4,46 @@ Task 1
...
-*/
\ No newline at end of file
+*/
+var largeButtonText = document.querySelector("#largerLinksBtn");
+largeButtonText.addEventListener("click", largeTextBIncrease);
+function largeTextBIncrease() {
+  var aElems = document.querySelectorAll("a");
+  console.log(aElems.length);
+  aElems.forEach(function(aElem) {
+    aElem.style.fontSize = "3em";
```
40thieves: Nice work! 🍰

week-5/A-dom-manipulation/exercise.js
khan-zareen: yes i did

week-5/A-dom-manipulation/exercise.js
khan-zareen: thanks

PR #304, opened 2019-01-30
Matthew HW Week 5

mtbdata711:


week-5/A-dom-manipulation/exercise.js
```diff
@@ -5,7 +5,15 @@ Task 1
 Write JavaScript below that changes the background colour of the page when the 'Change colour' button is clicked.
 */
 
+var btn = document.getElementById("bgrChangeBtn");
 
+btn.addEventListener("click", changeBackground);
+
+function changeBackground() {
+  var randomColour = ["blue", "purple", "yellow", "green", "white"];
```
40thieves: Nitpick: this variable name implies that it itself is a single colour. If you renamed it to something like `colors` then that frees up the variable below to be named `randomColor`.

Incidentally, in my experience it's usually better to use the American spelling of colour in code. That's how the computer spells it (see `backgroundColor`), so it can become confusing switching between the 2 😄

week-5/A-dom-manipulation/exercise.js
```diff
@@ -14,20 +22,45 @@ Task 2
...
 
 When a user clicks the 'Add some text' button, a new paragraph should be added below the buttons that says "Read more below."
 */
 
+var addText = document.querySelector("#addTextBtn");
+
+addText.addEventListener("click", addMoreText);
 
+function addMoreText() {
+  var JumboElement = document.querySelector(".jumbotron");
```
40thieves: Another little nitpick: remember camelCase variables start with a lowercase letter

week-5/A-dom-manipulation/exercise.js
```diff
@@ -14,20 +22,45 @@ Task 2
...
+  var addParagraph = document.createElement("p");
 
+  addParagraph.innerHTML = `<p>Read more below.</p>`;
```
40thieves: This is correct, but what is the resulting DOM structure? I think it might be a little odd. How could we fix that?

Extra bonus question: if we didn't want to use `innerHTML` here, what could we use instead? Hint: text node.

week-5/A-dom-manipulation/exercise.js
```diff
@@ -14,20 +22,45 @@ Task 2
...
+*/
+
+var largerLinks = document.querySelector("#largerLinksBtn");
+
+largerLinks.addEventListener("click", makeLinksLarger);
+
+function makeLinksLarger() {
+  var grabLinks = document.querySelectorAll("a");
+  grabLinks.forEach(function(item) {
+    item.style.fontSize = "2em";
```
40thieves: Nice work 👏

week-5-practise/js/main.js
```diff
@@ -0,0 +1,88 @@
...
+}
+
+function describeValid() {
+  return describeInput.value.length > 0;
+}
+
+
+//Button Event Listener
+submitBtn.addEventListener(
+  "click",
```
40thieves: This will work, but we can actually be a bit smarter about how we listen for the form being submitted. One of the useful features of `<form>`s is that you can hit the Enter key to trigger the submission - you don't have to click  on the button.

This matters here because we are only listening for the `click` event on the submit button. Instead we really want to listen for the `submit` event that is first **on the form**. Then if we click a submit button or if we hit the Enter key, the event handler is fired 🎉.

week-5-practise/js/main.js
```diff
@@ -0,0 +1,88 @@
...
+  function(event) {
+    if (!emailValid()) {
```
40thieves: Really nice use of small, well named functions 🏅

week-5-practise/js/main.js
```diff
@@ -0,0 +1,88 @@
...
+      setFormDefaults();
+    }
+    event.preventDefault();
+  },
+  false
+);
+
+//Reset form to default
+function setFormDefaults() {
+  emailInput.value = "";
```
40thieves: Nice work! 🍰

week-5-practise/js/main.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+    if (!describeValid()) {
+      describeInput.style.border = "1px solid red";
+      //describeInput.style.backgroundColor = "red";
+    }
+    if (emailValid() && nameValid() && describeValid()) {
+      alert("Thank you for filling out the form (:");
+      setFormDefaults();
+    }
+    event.preventDefault();
```
40thieves: You can technically put it anywhere in the function, but the convention is to put the `event.preventDefault()` as one of the first things in the event handler function.

week-5-practise/js/main.js
mtbdata711: Hey Ali! I've pushed this again with your suggestions. Please let me know what you think!

For the eventListener on the button I have put two different options. I found that having the button event and then a key event seemed to be the way to get the form working the best. 

With the submit event, while it would submit the form on the enter key from within the form if you clicked outside of the form or you were in the describe text box hitting enter would do nothing.

week-5/A-dom-manipulation/exercise.js
mtbdata711: Hey Ali!

This file has also been updated. Please let me know if you have any additional comments.

PR #305, opened 2019-02-01
Alisina's week 5 and week 5 practise solutions

Alisina92:
Pull request is ready.

### DIFF
--- a/week-5-practise/js/main.js
+++ b/week-5-practise/js/main.js
@@ -1,0 +1,54 @@
+/* The Blue Button*/
+var blueBtn = document.querySelector("#blueBtn");
+
+blueBtn.addEventListener("click", function() {
+  document.getElementsByClassName("jumbotron")[0].style.backgroundColor =
+    "#588fbd";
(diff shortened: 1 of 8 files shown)


week-5-practise/js/main.js
```diff
@@ -0,0 +1,54 @@
...
+
+  if (email.value.length === 0) {
+    email.style.backgroundColor = "red";
+  } else if (name.value.length === 0) {
+    name.style.backgroundColor = "red";
+  } else if (description.value.lengh === 0) {
+    description.style.backgroundColor = "red";
+  } else if (!email.value.includes("@")) {
+    email.style.backgroundColor = "red";
+  }
```
theseafaringturtle: A few suggestions:
+ chain if statemtents instead of else if in this case, so that all the invalid fields are highlighted rather than just one at a time
+ if they're valid set the background to white, so it doesn't stay red, and also display an alert message
+ nitpick: best to pack !email.value.includes("@") and email.value.length == 0 into a single if statement with an ||

PR #308, opened 2019-02-01
Joan week5 A

lexyking:


week-5-practise/index.html
```diff
@@ -1,157 +1,170 @@
...
+</head>
+
+<body>
+	<a href="#content" class="sr-only sr-only-focusable">Skip to main content</a>
+	<div class="site-header container">
+		<div class="navbar navbar-toggleable-md navbar-light">
+			<button class="navbar-toggler navbar-toggler-right" type="button" data-toggle="collapse" data-target="#navbarSupportedContent"
+			 aria-controls="navbarSupportedContent" aria-expanded="false" aria-label="Toggle navigation">
+				<span class="navbar-toggler-icon"></span>
+			</button>
```
nbogie: You've checked large changes to the .html  file (just, i think, the formatting).  This is a problem when you do `git add .` rather than carefully review which files you are intentionally changing.
It's distracting!

week-5/A-dom-manipulation/exercise.js
```diff
@@ -4,7 +4,24 @@ Task 1
...
 */
+////////////METHOD 1//////
+// var bgrChangeBtn = document.querySelector("#bgrChangeBtn");
+// bgrChangeBtn.addEventListener("click", changeBackground);
 
+// function changeBackground() {
+//     document.body.style.backgroundColor = "red";
+
+// }
+
```
nbogie: this looks good

week-5/A-dom-manipulation/exercise.js
```diff
@@ -4,7 +4,24 @@ Task 1
...
+//////////METHOD 2: CREATING AN ARRAY WITH THE CHILDREN OF THE DIV//////////
+var divButtonChild = document.querySelector(".buttons").children;
+//console.log(divButtonChild);
```
nbogie: Minor point: This is an array, so even if you think it's going to have one element in it, it should be called `divButtonChildren`, not `...Child`.

week-5/A-dom-manipulation/exercise.js
```diff
@@ -4,7 +4,24 @@ Task 1
...
+divButtonChild[0].addEventListener("click", changeBackground);
+
+function changeBackground() {
+    document.body.style.backgroundColor = "red";
+
+}
```
nbogie: It's good you can do it two ways.  The first way was much clearer to read.

week-5/A-dom-manipulation/exercise.js
```diff
@@ -13,7 +30,22 @@ Task 2
...
 When a user clicks the 'ALERT' button, an alert box should pop up with the text "Thanks for visiting Bikes for Refugees!"
 */
+///////////METHOD 1////////////
+// var alertBtn = document.querySelector("#alertBtn");
+// alertBtn.addEventListener("click", alertMessage);
 
+// function alertMessage() {
+//     alert("Red light");
+//     alertBtn.style.backgroundColor = "red";
+// }
```
nbogie: the above is good.

week-5/A-dom-manipulation/exercise.js
```diff
@@ -13,7 +30,22 @@ Task 2
...
+
+/////////////METHOD 2///////////////
+divButtonChild[1].addEventListener("click", alertMessage);
+
```
nbogie: Imagine your boss asks you to reorder the buttons, or add a new one at the start.  Suddenly you have to change many lines of code!

week-5/A-dom-manipulation/exercise.js
```diff
@@ -22,12 +54,57 @@ Task 3
...
 */
+// var addTextBtn = document.querySelector("#addTextBtn");
+// addTextBtn.addEventListener("click", addText);
 
+////////////METHOD 2//////////
+divButtonChild[2].addEventListener("click", addText);
+
+function addText() {
+    var paragraph = document.createElement('p'); // here we're just creating it, element is not visible yet
+    addTextBtn.appendChild(paragraph);
```
nbogie: You haven't declared or initialised `addTextBtn` (you've commented that out).

week-5/A-dom-manipulation/exercise.js
```diff
@@ -22,12 +54,57 @@ Task 3
...
+    paragraph.innerText = "Read me below.";
+    addTextBtn.removeEventListener("click", addText); // to stop at one paragraph
+    addTextBtn.style.backgroundColor = "Yellow"; //Just an extra
+}
```
nbogie: Good.  I like that you remove the event listener. 

If you read the spec carefully, it says the text should be added below the *buttons*, not below the text of one button!  The mechanism will be the same, anyway.

week-5/A-dom-manipulation/exercise.js
```diff
@@ -22,12 +54,57 @@ Task 3
...
+//     large.forEach(function (element) {
+//         if (element.style.fontSize == "xx-large") {
+//             element.style.fontSize = "medium";
+
+//         } else {
+//             element.style.fontSize = "xx-large";
+//         }
+//     });
+
+// }
```
nbogie: Nice idea to make it toggle between xx-large and medium.  (Perhaps you could update the button's text so it is clear what will happen when you next press it.)

week-5-practise/js/main.js
```diff
@@ -0,0 +1,66 @@
+////////Part 1//////////
+var jumbotron = document.querySelector(".jumbotron");
+var jumbotronButton = document.querySelector(".buttons").children;
+
```
nbogie: Maybe name this `jumbotronButtons` not `jumbotronButton` - it's an array.

I'd define `volunteerBtn` at the start, and `donateBtn`.  Much more readable than `jumbotronButton[1]`, etc, and if we have to reorder the buttons or add a new one at the front, we only have to change 2 lines of code, instead of 8.

week-5-practise/js/main.js
```diff
@@ -0,0 +1,66 @@
...
+    jumbotronButton[1].style.backgroundColor = "#8c9c08";
+
+}
+
+
+////////part 2//////////
+var emailCheck = document.querySelector("#exampleInputEmail1");
+var nameCheck = document.querySelector("#example-text-input");
+var textCheck = document.querySelector("#exampleTextarea");
+
```
nbogie: These would be better named as inputs or fields.  `emailInput` or `emailField`, for example.  `textTA` or `textField`.  Something named a `check` is more likely to be a function.

week-5-practise/js/main.js
```diff
@@ -0,0 +1,66 @@
...
+//var labelText = document.querySelectorAll(".form-control");
+// console.log(labelText); ////3 nodes
+
+var formBtn = document.querySelector("form").children; /// 3 div and 1 form
+formBtn[3].addEventListener("click", validateForm);
+
+function validateForm() {
+    event.preventDefault();
+    if ((emailCheck.value != "") && (nameCheck.value != "") && (textCheck.value != "")) {
+
```
nbogie: you could just add `&& emailCheck.value.includes("@")` to the condition of your top if statement, and get rid of the nested if.

week-5-practise/js/main.js
```diff
@@ -0,0 +1,66 @@
...
+        } else {
+
+            emailCheck.style.backgroundColor = "red";
+        }
+    } else {
+        emailCheck.style.backgroundColor = "red";
+        nameCheck.style.backgroundColor = "red";
+        textCheck.style.backgroundColor = "red";
+    }
+}
```
nbogie: There's a bug or two here.  
Bug 1) If I enter something for the name field and press Submit, all fields go red, but only the invalid fields (email and description) should turn red.
The specification says:  `For all the fields that invalid, it should make their background color 'red'.`

Bug 2) If I first submit with an invalid field and cause all fields to turn red, and THEN I correct all fields and submit, it will submit correctly and empty the fields, but the fields *remain red*! 

Be careful to test various paths through your finished work.

PR #320, opened 2019-02-08
Tayo week5

Atayo:


Atayo:
I managed to include week 4 into week 5. I have communicated with @foaly-nr1 who has advised me on further improvements.

PR #341, opened 2019-02-27
Tayo-week3practice-map-exercise

Atayo:


week-2-practise/level-2/6-map.js
```diff
@@ -2,9 +2,13 @@
 // 1. Write a function that formats the numbers into a string with the percentage symbol appended e.g. "10%"
```
martynamaron: Sorry @Atayo but this file shouldn't be included in this PR

week-4/A-objects-intro/exercise-part-0.js
```diff
@@ -4,4 +4,11 @@ Describe your own laptop as a JavaScript object
```
martynamaron: This file shouldn't be included in the PR either

PR #344, opened 2019-02-27
Clement-JS-week8

clems27:


week-8-practise/1-code-refactoring.js
```diff
@@ -8,28 +8,28 @@
...
+  return text
+    .split("")
+    .reverse()
+    .join("")
+    .toUpperCase();
+}
 
-// Function that returns text with every other letter capitlised
+//this function will transform all text to upperCase,
 export function everyOtherLetterCapitilised(text) {
```
sansaid: What do you think about the naming of the function?

week-8-practise/1-code-refactoring.js
```diff
@@ -8,28 +8,28 @@
...
-  for (i==0; i++; i<text.length) {
-      if (i/2==0) {
-       text[i] = text[i].toUpperCase() 
-    }    
-  } 
+  return text.toUpperCase();
```
sansaid: The intention of the function is to capitilise every other letter (i.e. "practise" becomes "PrAcTiSe"). With that, the for loop should stay, though there is a logical error in the for loop.

week-8-practise/3-classes.js
```diff
@@ -29,14 +29,92 @@
  */
 
 class House {
-  constructor(windows, colour) {
-    this.windows = windows
+  constructor(windows, colour = "blue", numberOfRooms, location) {
```
sansaid: Although we haven't covered this yet, default parameters should be right-most, because JS maps the value to the argument by position.

week-8-practise/1-code-refactoring.js
clems27: @sansaid This is the toughest part, has decided to just rename the function upperCase, since that is the main purpose of the  function, will be pushing the update shortly

week-8-practise/1-code-refactoring.js
clems27: @sansaid  As the instruction was not too clear on the intention of the function, I thought I might as well just capitalise all the string, I have re-inserted the loop in the function, will appreciate if you review it again

week-8-practise/3-classes.js
clems27: @sansaid please do you mind clarifying what you mean?

week-8-practise/1-code-refactoring.js
sansaid: No worries - I'll take responsibility for that, there was definitely scope to make it clearer. Thanks!

week-8-practise/3-classes.js
sansaid: Sure! So you used `colour="blue"` to set colour to have a default value. The only problem is that default values need to be one of the last arguments. So instead of `constructor(windows, colour = "blue", numberOfRooms, location)`, you should have `constructor(windows, numberOfRooms, locatoin, colour = "blue")` - non-default arguments on the left, default arguments on the right. 

The reason for this is that when you pass values into the function but don't want to pass a value for the `colour`, Javascript will always map the second argument to `colour`. However, if you positioned the default parameters last, then all you have to do is leave the last arguments out and Javascript will know to use the default values.

clems27: Noted!  Thanks for the clarification, will update that line of code.

week-8-practise/1-code-refactoring.js
clems27: @sansaid  what do you think about how I used the loop in this exercise?..., was that your expectation?

PR #345, opened 2019-02-27
Ivina - week 8 - practice exercises

ivinapontes:


week-8-practise/1-code-refactoring.js
```diff
@@ -1,35 +1,34 @@
...
+    .split("")
+    .reverse()
+    .join("");
+  return reversed;
+}
+// Function that returns text with every other letter capitlised
 
-  return reversed
-  }
+function everyOtherLetterCapitilised(word) {
```
sansaid: What do you think about the naming of the function?

PR #376, opened 2019-06-21
These are my completed exercises of CYF from week1 to week 3-practice.

Perasperad:
One or two exercises could be released togheter with week 4.


week-2-practise/level-2/4-slice.js
```diff
@@ -5,7 +5,11 @@
 // - remove the item at the specified index
 
 function remove(arr, index) {
-  return; // complete this statement
+  if (typeof arr[arr.length - 1] == "string") {
```
jslopezgithub: You can write this by using only splice. The first method takes the index where to start removing and the second argument how many items to remove. e.g. `splice(index, 1)`

week-2-practise/level-2/6-map.js
```diff
@@ -4,7 +4,11 @@
 //    - each item in `percentages` formatted by your function
 
 var percentages = [1, 23, 92, 18];
-
+function stringPercentage(num) {
+  typeof num === "strings";
```
jslopezgithub: this line is not needed

week-2-practise/level-3/2-validation.js
```diff
@@ -6,7 +6,9 @@
 
 var excludedNums = [6, 14, 91, 111];
 
-function validate(num) {}
+function validate(num) {
+  typeof num === "number";
+}
 
 console.log(validate(6));
```
jslopezgithub: This will print out undefined instead of false or true

week-2-practise/level-3/2-validation.js
```diff
@@ -6,7 +6,9 @@
 
 var excludedNums = [6, 14, 91, 111];
 
-function validate(num) {}
+function validate(num) {
+  typeof num === "number";
```
jslopezgithub: Test this code with 3.6 (it should return false) because it is a float and it is not on excludedNums

week-2-practise/level-3/4-slice.js
```diff
@@ -5,7 +5,7 @@
 // - insert the replacement value at the provided index
 
 function replace(arr, index, value) {
-  return; // complete this statement
+  return arr.slice(arr.indexOf(index), 1, value); // complete this statement
```
jslopezgithub: This does not generate the correct output.

week-2-practise/level-3/5-indexOf.js
```diff
@@ -5,8 +5,9 @@
 // - NOT change the original array
 // - return a new array with the replacement value inserted
 
-function replace(arr, valueToReplace, newValue) {
-  return; // complete this statement
+function replace(numbers, valueToReplace, newValue) {
+  numbers.splice(numbers.indexOf(valueToReplace), 1, newValue);
+  return numbers; // complete this statement
```
jslopezgithub: You need to create a new array to avoid changing the original array. Once you fix that problem see if the output matches the expected output

week-2-practise/level-3/6-map.js
```diff
@@ -11,11 +11,12 @@ function tidyUpString(str) {
 }
 
 function captialise(str) {
-  // complete this function
+  return str.toUpperCase; // complete this function
```
jslopezgithub: In order to call the function, you need to use () at the end of the method.

week-2-practise/level-3/6-map.js
```diff
@@ -11,11 +11,12 @@ function tidyUpString(str) {
...
 }
 
 var mentors = ["/Daniel ", "irina ", " Gordon", "ashleigh "];
-var mentorsTidyAndCapitalised;
+var mentorsTidy = mentors.every(tidyUpString);
+var mentorsTidyAndCapitalised = mentorsTidy.map(captialise);
```
jslopezgithub: The output should be `[ 'Daniel', 'Irina', 'Gordon', 'Ashleigh' ]`

week-2/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = true; // finish this statement
```
jslopezgithub: this should be done based on using a comparison with studentCount and mentorCount

week-2/F-logical-operators/exercise.js
```diff
@@ -11,14 +11,14 @@ var cssLevel = 4;
...
-var cssLevelAbove5;
+var htmlLevelAbove5 = htmlLevel > 5;
 
+var cssLevelAbove5 = cssLevel > 5;
 // Finish the next two statement
 // Use the previous variables and logical operators
 // Do not "hardcode" the answers
-var cssAndHtmlAbove5;
-var cssOrHtmlAbove5;
+var cssAndHtmlAbove5 = htmlLevel > 5 && cssLevel > 5;
```
jslopezgithub: you could have also use htmlLevelAbove5 instead of using `htmlLevel > 5` again.

week-2/G-conditionals-2/exercise-4.js
```diff
@@ -9,7 +9,11 @@
 */
 
 function containsCode(sentence) {
-
+  if (sentence.includes("code")) {
```
jslopezgithub: this simplifies to `return sentence.includes("code")`

week-2/I-array-properties/exercise.js
```diff
@@ -6,7 +6,7 @@
 */
 
 function isEmpty(arr) {
-  return; // complete this statement
+  return arr.length; // complete this statement
```
jslopezgithub: This returns a number instead of a true or false

week-2/L-array-methods-2/exercise.js
```diff
@@ -15,8 +15,8 @@ var everyone = [
   "Swathi"
 ];
 
-var firstFive; // complete this statement
-var lastFive; // complete this statement
+var firstFive = everyone.slice(0, 1, 2, 3, 4); // complete this statement
```
jslopezgithub: This does not result in the correct output.

PR #379, opened 2019-07-10
Roma Class1 - Week 1 to 4

marco1774:


week-2/C-comparison-operators/exercise.js
```diff
@@ -7,14 +7,14 @@
 
 var studentCount = 16;
 var mentorCount = 9;
-var moreStudentsThanMentors; // finish this statement
+var moreStudentsThanMentors = studentCount > mentorCount; // finish this statement
 
 var roomMaxCapacity = 25;
-var enoughSpaceInRoom; // finish this statement
+var enoughSpaceInRoom = true; // finish this statement
```
jslopezgithub: This should be based in terms of the variables roomMaxCapacity, studentCount and mentorCount

week-2/G-conditionals-2/exercise-2.js
```diff
@@ -7,8 +7,12 @@
 
 */
 
-function studentPassed(grade) {
-
+function studentPassed(grade1) {
+  if (grade1 < 50) {
```
jslopezgithub: 👍 There is also a convenience ternary operator that allows coding this in this into one line
`return grade1 < 50 ? "passed" : "failed";`

condition ? <value for true> : <value for false>;

week-2/G-conditionals-2/exercise-4.js
```diff
@@ -9,7 +9,10 @@
 */
 
 function containsCode(sentence) {
-
+  if (sentence.includes("code")) {
```
jslopezgithub:
```suggestion
  return sentence.includes("code");
```

week-3-practise/level-2/1-map.js
```diff
@@ -12,7 +12,12 @@ var londonLocations = [
   ["Greenwich", "tube", "bus", "river boat"]
 ];
 
-var locationsByBoat;
+var locationsByBoat = londonLocations.map(prova);
```
jslopezgithub: This does not produce the expected output. The output is currently `[undefined, "London Bridge", undefined, "Greenwich"]`. You need to use filter to remove undefined entries

week-3-practise/level-2/2-map.js
```diff
@@ -14,7 +14,12 @@ var attendances = [
   ["Nina", 10]
 ];
 
-var eligibleStudentNames; // TODO: Complete this line.
+var eligibleStudentNames = attendances.map(check); // TODO: Complete this line.
```
jslopezgithub: This has the same problem as before with undefined values.

week-3-practise/level-3/1-practise-all.js
```diff
@@ -42,7 +42,55 @@ Expected Result:
...
+  "ttkTu.wer3",
+  "dvyyeyY!5",
+  "qwbfj76%",
+  "tytT3729."
+];
 var password = ["Se%5", "TktE.TJTU", "384HsHF", "dvyyeyY!5", "tryT3729."];
 
-console.log(PasswordValidationResult)
\ No newline at end of file
+function controllLevell1(arr) {
```
jslopezgithub: I tried running your solution but I was not able to. I'm posting myself solution so that you can compare.
```suggestion
function passwordValid1(password) {
    return (password.length > 4) 
    && password.match("[A-Z]") != null
    && password.match("[a-z]") != null
    && password.match("[0-9]") != null;
}

function passwordValid2(password) {
    return passwordValid1(password) && (password.match("[!#$%.]") != null);
}

function passwordValid3(password) {
    return passwordValid2(password) && !previousPassword.includes(password);
}

var PasswordValidationResult = password.map(passwordValid3);

console.log(PasswordValidationResult
```

week-3/D-array-filter/exercise.js
```diff
@@ -8,7 +8,11 @@
 
 var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1], false, "whoops"];
 
-var pairsByIndex; // Complete this statement
+var pairsByIndex = pairsByIndexRaw.filter(clean); // Complete this statement
+
+function clean(elem) {
```
jslopezgithub: This is not enough to remove the bad entries. This needs to clean null, [1], false and "whoops" from the array

PR #389, opened 2019-07-26
Edited week-5/F-webchat/exercise-1.js according to the comments

saeendamghanian:
I tried to edit the exercise according to the comments. I used forEach to iterate inside the fetch file and also used json() instead of text() for getting the result of fetch file. I used async-await on the promises but I don't know it is necessary or not. I changed one exercise and wanted to make PR with that but all the committed files are on the PR!! I don't know if there is a way to make PR only with the file that I changed.

### DIFF
--- a/week-5/A-dom-manipulation/exercise.js
+++ b/week-5/A-dom-manipulation/exercise.js
@@ -4,8 +4,12 @@ Task 1
 
 Write JavaScript below that changes the background colour of the page when the 'Change colour' button is clicked.
 */
+var btn_changeColor = document.querySelector("#bgrChangeBtn");
+btn_changeColor.addEventListener("click", changeBgrColor);
 
-
+function changeBgrColor() {
+  document.querySelector("body").style.backgroundColor = "red";
+}
 
 /*
 Task 2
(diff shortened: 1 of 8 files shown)


week-5/A-dom-manipulation/exercise.js
```diff
@@ -4,30 +4,64 @@ Task 1
...
-*/
\ No newline at end of file
+*/
+console.log("Hello");
+
+var btn_largerLink = document.querySelector("#largerLinksBtn");
+btn_largerLink.addEventListener("click", largeLinks);
+
+function largeLinks() {
+  var allLinks = document.querySelectorAll(
```
jslopezgithub: There were probably many answers to this problem. This is the one I used:
```
    document.querySelectorAll("a").forEach((link) => link.style.fontSize = "2em");
```

week-5/C-promises/exercise.js
```diff
@@ -55,7 +62,16 @@ function exercise3() {
   YOUR NAME"
 */
 function exercise4() {
-  var promise4 = concatPromise()
+  var promise4 = concatPromise();
+  promise4
+    .then(firstResolvedValue => {
+      return new Promise(resolve => {
```
jslopezgithub: There is no need to create a new Promise, it happens automatically. You can just do 
```suggestion
    promise4.then(firstResolvedValue => firstResolvedValue + "Saeen")
```

jslopezgithub:
Good job

week-5/C-promises/exercise.js
saeendamghanian: I did that because in the question is asked to return the concatenated string as a promise and do the rest on it.

jslopezgithub: You can return the concatenated string, it gets wrapped in a Promise automatically. I will have to re-read the question but I don't think it wanted you to explicitly crate the Promise since there is no need.

week-5/A-dom-manipulation/exercise.js
saeendamghanian: Exactly. I tried to change only the three links that exist on the page and anytime you click on the button their size will be increase.

PR #391, opened 2019-07-29
Exercises from week 4 to week 5

Perasperad:


week-5/F-webchat/exercise-1.js
```diff
@@ -33,5 +35,18 @@ When you open index.html in your browser, it should display the existing message
...
+function name() {
+  fetch("https://codeyourfuture.herokuapp.com/api/messages", {
+    method: "GET",
+    headers: { "Content-Type": "application/json" }
+  })
+    .then(res => {
+      return res.json();
+    })
+    .then(text => {
+      document.getElementById("message-list").innerHTML = JSON.stringify(text);
```
jslopezgithub: text is an array of json messages. You would want to create a displays that shows a message per line. For example, you can create a paragraph like you did earlier and add the message and the timestamp. Perhaps you might want to style the timestamp in a color and the message in a different color. `JSON.stringify()` displays the messages but it is difficult to read a conversation.

PR #395, opened 2020-02-09
Dummy PR to make a 'bad' comment on, please ignore

40thieves:


week-1/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,6 @@
-var mentor1 = "Daniel";
-var mentor2 = "Irina";
-var mentor3 = "Mimi";
-var mentor4 = "Rob";
-var mentor5 = "Yohannes";
+var mentors = ["Daniel", "Irina", "Mimi", "Rob", "Yohannes"];
+mentors.forEach(function (mentor) { console.log(`HELLO ${mentor.toUpperCase()}`); });
```
40thieves: Don't use `forEach`, it has side-effects. `foo.map(f)` is pure. Extract the function too.

PR #414, opened 2020-02-24
js-week-1

ferhatu:
week-1
1,2,3 and 4 done

### DIFF
--- a/week-1/Homework/mandatory/3-function-output.js
+++ b/week-1/Homework/mandatory/3-function-output.js
@@ -1,31 +1,45 @@
 // Add comments to explain what this function does. You're meant to use Google!
+/*Math.random() returns a number lower than 1. In this function it returns a
+random integer from 0 to 9 */
 function getNumber() {
-    return Math.random() * 10;
+  return Math.random() * 10;
 }
 
 // Add comments to explain what this function does. You're meant to use Google!
+/*This function joins w1 and w2 and returns the new value   */
 function s(w1, w2) {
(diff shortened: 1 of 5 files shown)


week-1/Homework/mandatory/4-tax.js
```diff
@@ -17,25 +20,34 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function formatCurrency() {}
+function formatCurrency(amount) {
+  let totalAmount = amount + amount * 0.2;
```
sansaid: Good first attempt @ferhatu! How can you improve this to make it more simple? Is there a way you can rewrite this line to use one `amount` reference?

week-1/Homework/mandatory/3-function-output.js
```diff
@@ -1,31 +1,45 @@
...
+/*This function joins w1 and w2 and returns the new value   */
 function s(w1, w2) {
-    return w1.concat(w2);
+  return w1.concat(w2);
 }
 
 function concatenate(firstWord, secondWord, thirdWord) {
-    // Write the body of this function to concatenate three words together
-    // Look at the test case below to understand what to expect in return 
+  return firstWord.concat(" ", secondWord, " ", thirdWord);
```
sansaid: Interesting approach! It's a good thing that you learnt about the .concat() string method! How could you rewrite this to use the interpolation we learned in class to make this more readable?

PR #416, opened 2020-02-24
Tayfun week 1

tayfuntayfun:


week-1/Homework/mandatory/3-function-output.js
```diff
@@ -1,14 +1,20 @@
...
 
 // Add comments to explain what this function does. You're meant to use Google!
+// This function concatenates the second string after the first string as seen below.
 function s(w1, w2) {
     return w1.concat(w2);
 }
 
 function concatenate(firstWord, secondWord, thirdWord) {
+
+    return firstWord.concat(` ${secondWord} ${thirdWord}`);
```
brightpixels: This works but there is a better way to do this.

week-1/Homework/mandatory/4-tax.js
```diff
@@ -5,7 +5,14 @@
   Sales tax is 20% of the price of the product
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(productPrice) {
+
```
brightpixels: Well done. This works but can you think of anything else to make break down this code so it's more readable. Imagine your work colleague doesn't know what 0.20 represents. What can you do to make code readable.

week-1/Homework/mandatory/4-tax.js
```diff
@@ -17,7 +24,14 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function formatCurrency() {}
+function formatCurrency(productPrice) {
```
brightpixels: - I like the way you make `£` a variable on its own first.
- The way you named the variable `SalesPrice` doesn't follow convention. Can you work out what you are doing wrong here?
- `formatCurrency` tests are failing. That means your function implementation is wrong. You are close but missing important functionality that was requested in the comments: `prices must have 2 decimal places`

week-1/Homework/mandatory/3-function-output.js
tayfuntayfun: Hi Abdo, here is a different way of scripting this part.

week-1/Homework/mandatory/3-function-output.js
```diff
@@ -1,14 +1,20 @@
...
+
+    //return firstWord.concat(" " + secondWord + " " + thirdWord);
     // Write the body of this function to concatenate three words together
     // Look at the test case below to understand what to expect in return 
 }
```
tayfuntayfun: Changed how concat gets the strings from the function.

week-1/Homework/mandatory/4-tax.js
tayfuntayfun: camelCase :), you are right. I have also added .toFixed function to have 2 digit decimal added to the product price. Hope I have done the right thing.

PR #423, opened 2020-02-26
JS1/Week-1/Mo

mnashawati:


week-1/Homework/extra/2-piping.js
```diff
@@ -16,43 +16,53 @@
...
-const startingValue = 2
+const startingValue = 2;
+
+let finalResult = format(multiply(add(startingValue, 10), 2));
 
 // Why can this code be seen as bad practice? Comment your answer.
-let badCode = 
+// All the code is written on one line, more difficult to read and modify.
+
+let badCode = finalResult;
```
sansaid: You applied the correct "bad code"! But how come you decided to use a different variable for the result? (i.e. `finalResult` instead of `badCode`)?

week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -43,19 +43,69 @@ Outlook not so good.
 Very doubtful.
 */
 
+let array = [
+  // Very Positive:
```
sansaid: Very good job laying out your array this way! It's very easy for me to read this as a result :)

week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -43,19 +43,69 @@ Outlook not so good.
...
+  "My reply is no",
+  "My sources say no",
+  "Outlook not so good",
+  "Very doubtful."
+];
+
+function randomAnswer(array) {
+  return array[Math.floor(Math.random() * array.length)];
+}
+console.log(randomAnswer(array));
```
sansaid: It's good practice to remove your `console.log` statements after using them, otherwise we're leaving unnecessary statements in our scripts which might confuse other developers

week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -43,19 +43,69 @@ Outlook not so good.
...
+    answerType = "very positive";
+  } else if (array.indexOf(randomAnswer(array)) < 10) {
+    answerType = "positive";
+  } else if (array.indexOf(randomAnswer(array)) < 15) {
+    answerType = "negative";
+  } else {
+    answerType = "very negative";
+  }
+  // console.log(answerType);
+  return answerType;
```
sansaid: Did you intend to modify `answer` instead of `answerType` here? `answerType` is a global variable which you're modifying, instead of `answer`.

sansaid:
Overall, good attempt at keeping your code clean! :) I've added some comments for you to review, let me know what you think :)

week-1/Homework/extra/2-piping.js
mnashawati: lol I am not sure why I did it, I removed it now.

week-1/Homework/mandatory/5-magic-8-ball.js
mnashawati: (answer) is the argument for checkAnswer function. 
(answerType) is what checkAnswer function returns i.e. "very positive", "positive" etc... 
I declared it globally in order to use it in the function.. 
Is this bad practise? 
Should I have declared it within the function instead?

week-1/Homework/mandatory/5-magic-8-ball.js
mnashawati: Removed.

week-1/Homework/mandatory/5-magic-8-ball.js
mnashawati: Thanks!

week-1/Homework/extra/2-piping.js
mnashawati: I am not sure why I added another variable lol. I removed it now.

frosas:
(@mnashawati she is 😉)

week-1/Homework/mandatory/5-magic-8-ball.js
frosas: The function returns the right kind of result (a string containing an answer type) but that answer type should be based on the `answer` parameter provided to the function.

week-1/Homework/extra/1-currency-conversion.js
```diff
@@ -16,20 +19,24 @@ function convertToUSD() {}
   Find a way to add 1% to all currency conversions (think about the DRY principle)
 */
 
-function convertToBRL() {}
+function convertToBRL(sterlingPrice) {
+  const brlPrice = sterlingPrice * 5.7;
+  const finalPrice = brlPrice + brlPrice / 100;
+  return finalPrice;
+}
```
frosas: I find the names of the variables very descriptive 👌

Notice though that, with this implementation, the 1% fee we are supposed to be applying is actually a 1% _gift_ 😱

week-1/Homework/extra/2-piping.js
```diff
@@ -16,43 +16,51 @@
...
+function add(a, b) {
+  const total = a + b;
+  return Number(total.toFixed(2));
 }
 
-function multiply() {
-
+function multiply(a, b) {
+  const total = a * b;
+  return Number(total.toFixed(2));
```
frosas: Notice that, by making `add()` and `multiply()` return a result truncated to two decimals, we unnecessarily lose precision when we combine them (like we do below). It'd be a better practice to let the formatting aspects to `format()` :)

week-1/Homework/extra/1-currency-conversion.js
mnashawati: @frosas Really appreciate you reviewing my code :D... I checked it again, I think you are right (I should probably deduct the fee from the final BRL price), but if I change the code the test will fail. Do you think the test is incorrect?

week-1/Homework/extra/2-piping.js
mnashawati: Noted, I totally agree! But again the case 2 test for (Add) contains a decimal number, so if I remove it from there the test will fail.. Unless there is another way/method that I can use instead of toFixed()?

week-1/Homework/extra/1-currency-conversion.js
frosas: Oh I see. I'd say the test is incorrect as it just makes no sense as it is, but feel free to leave this as it is in any case.

week-1/Homework/extra/2-piping.js
frosas: You're right (again).

IMO the test is implemented too naively, float numbers (which is how JS stores numbers with decimals) are not precise and you can't simply compare them to other floats with `===`.

Instead, you want to check if the two floats are close enough. The test could've been implemented like this:

```js
// `tolerance` is the maximum difference we allow between the two numbers. Notice
// we have to provide it on every comparison as there is no one size fits all.
const floatsEqual = (a, b, tolerance) => Math.abs(a - b) < tolerance;

// Here for example we decide that these two floats are equal if they are no more 
// than 0.1 different
test("add function - case 2 works", floatsEqual(add(2.4, 5.3), 7.7, 0.1));
```

In reality there are not many cases where you actually need to treat floats as precise numbers. If you find yourself needing that, you may want consider avoiding floats altogether (for example, when working with money is common to represent amounts in cents so that you don't have decimals).

PR #427, opened 2020-02-27
fixed introduceMe function: FAILED

Hanifix:
fixed introduceMe function: FAILED
i am strugling with this part

### DIFF
--- a/week-1/Homework/mandatory/2-logic-error.js
+++ b/week-1/Homework/mandatory/2-logic-error.js
@@ -1,31 +1,31 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+    return word;
 }
 
 function getWordLength(word) {
-  return "word".length()
+    return word;
 }
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+    return a * b * c;
+
 }
 
 /* ======= TESTS - DO NOT MODIFY ===== */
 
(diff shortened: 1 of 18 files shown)


Hanifix:
This is my first update for the 2-logic

week-1/Homework/mandatory/2-logic-error.js
```diff
@@ -1,31 +1,31 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+    return word;
```
amandaeareias: In this function you're supposed to find a method that trims the string (removes whitespace from both ends of it). For example:
`   hanif    ` would became `hanif`

You don't need to do it yourself, JavaScript has built in methods that you can use on strings. You can try looking for it on google and implement it here

week-1/Homework/mandatory/2-logic-error.js
```diff
@@ -1,31 +1,31 @@
...
 }
 
 function getWordLength(word) {
-  return "word".length()
+    return word;
```
amandaeareias: In this function we want to get the word's **length** (quantity of characters in it). For example: 
if **word** is `hanif` the result would be `6`.

Same as above: You don't need to do it yourself, JavaScript has built in methods that you can use on strings. You can try looking for it on google and implement it here

Hanifix:
This are some modification i have done

PR #429, opened 2020-02-28
Friday

abcelen:


week-1/Homework/mandatory/1-syntax-errors.js
```diff
@@ -1,19 +1,25 @@
...
+function introduceMe(name, age){
+return "Hello, my name is " + name + " and I am " + age + " years old";
+
+}
+
+
 
 function getRemainder(a, b) {
-    remainder = a %% b;
+    remainder = a % b;
```
frosas: Works great but you forgot something in front of this `remainder` variable and now it's implicitly a global variable 😱 Do you know what it is?

week-1/Homework/mandatory/3-function-output.js
```diff
@@ -1,17 +1,18 @@
-// Add comments to explain what this function does. You're meant to use Google!
+// returns a random number between 0 (inclusive),  and 1 (exclusive)
 function getNumber() {
     return Math.random() * 10;
 }
 
-// Add comments to explain what this function does. You're meant to use Google!
+// to join two or more arrays.
```
frosas: That's correct. Which other type of value other than arrays would this function allow us to concatenate?

week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -1,61 +1,82 @@
-/**
+//I couldnt write this code alone. we did together with Ferhat, Serkan and Tayfun
```
frosas: Thanks for letting us know 👍

week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -1,61 +1,82 @@
...
+  "My reply is no",
+  "My sources say no",
+  "Outlook not so good",
+  "Very doubtful"
+];
+function shakeBall() {
+  console.log("The ball has shaken!");
+  let i = Math.floor(Math.random() * answers.length);
+  return answers[i];
+}
```
frosas: Clean and concise. Also kudos for using `answers.length` instead of `20` 👍

week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -1,61 +1,82 @@
...
+    return "very positive";
+  } else if (positive.includes(result)) {
+    return "positive";
+  } else if (negative.includes(result)) {
+    return "negative";
+  } else {
+    return "very negative";
+  }
+}
+
```
frosas: Very good solution, you saved so much code by doing it this way!

One small detail, given how `checkAnswer()` is used in the test below, I think it was supposed to accept an `answer` parameter (eg `function checkAnswer(answer) { ... }`) instead of reading the value in the outer variable `result`.

PR #431, opened 2020-02-28
add make-a-radio exercise

jamesgorrie:
This does on the surface sound quite tricky I think.

I have tried to be as prescriptive as possible in the comments, I wonder if that might mean people just following steps and losing understanding, but maybe that's fine.

I wasn't sure if to call frequencies frequencies or channels... 

The testing logic is weird, but that's JS for you.

To not muddy the waters you can add:
```js
 /* ======= TESTS - DO NOT MODIFY ===== */ 
require("../tests/4-radio-stations.js")
```

Which would maybe make it better and allow us to iterate on the testing functions, but maybe that's not an issue that needs solving as the simple version is great.

## TODO:
- [ ] write logging test

### DIFF
--- a/week-2/Homework/mandatory/4-radio-stations.js
+++ b/week-2/Homework/mandatory/4-radio-stations.js
@@ -1,0 +1,79 @@
+/**
+ * Finding a radio station, and a good one, can be hard manually.
+ * Let's use some code to help us build a program that helps us scan
+ * the radio waves for some good music.
+ */
+
+/**
+ * First, let's create a function that creates a list of all the frequencies.
+ * Call this function `getAllFrequencies`.
+ *
+ * This function should:
+ * - Create an array  starting at 87 and ending in 108
+ * - Should return this array to use in other functions
+ */
+
+// `getAllFrequencies` goes here
+
+/**
+ * Next, let's write a function that gives us only the frequencies that are radio stations.
+ * Call this function `getStations`.
+ *
+ * This function should:
+ * - Get the available frequencies from `getAllFrequencies`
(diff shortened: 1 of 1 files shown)


week-2/Homework/mandatory/4-radio-stations.js
```diff
@@ -0,0 +1,79 @@
...
+ * Call this function `searchRadioWaves`.
+ * - "🎶 station found at ${frequency}, enjoy!"
+ * - "No station found at ${frequency}, moving on"
+ */
+// `searchRadioWaves` goes here
+
+/* ======= TESTS - DO NOT MODIFY ===== */
+
+function isRadioStation(frequency) {
+  // This is just a way of caching the random array internally
```
sansaid: I'd probably use a different word to cache for now (or start with a simple word and use "cache" in brackets). In this case I'd probably avoid using cache since that's not what it's actually doing (i.e. copying data closer to where it can be retrieved - it's actually being generated here for the first time).Suggested wording: "This is a way of storing the random frequency array inside a function as storing it outside is bad practice".

I'd also add more of an explanation of what this function does to set an example on how to comment code to the students.

week-2/Homework/mandatory/4-radio-stations.js
```diff
@@ -0,0 +1,79 @@
...
+  // so it's not available outside of this funtion
+  // Not best practise
+  if (!this.stations) {
```
sansaid: I'm a bit conflicted with introducing to OOP so soon but also want to do something more interesting than just creating function that returns a static list. I'll come back to this comment after I've thought a little more about possible alternatives. @aldeez988 and @ismaelgt , what do you guys think?

PR #432, opened 2020-02-28
Yohannes Js HW Week-1

byohannes:


week-1-practise/level-2/4-currency-conversion-2.js
```diff
@@ -7,17 +7,21 @@
   Find a way to add 1% to all currency conversions (think about the DRY principle)
 */
 
-function convertToUSD(price) {}
-function convertToBRL(price) {}
+function convertToUSD(price) {
+  return price * 1.4
+}
+function convertToBRL(price) {
+  return price * 5.77 + 0.01 * price
```
frosas: I don't understand that `0.01 * price`, what is it for? 🤔

byohannes: I am adding 1%  of the price to the currency conversion as we have been asked on the top comment

frosas: Ah, I miss that part 😅

If I understand it correctly, I see some details that don't match what's being asked:

- The 1% fee has to be applied to both USD and BRL conversions
- The 1% fee should be added to the price in GBP, and then the currency conversion should be applied to the resulting amount. It's a subtle change but it makes the final price to be different!

byohannes: Ohh okay I will do the changes then thanks frances

week-1-practise/level-3/magic-8-ball.js
```diff
@@ -0,0 +1,222 @@
...
+  test(`shakeBall returns an string answer"`, typeof answer === 'string')
+  test(
+    `checkAnswer returns the level of positivity"`,
+    ['very positive', 'positive', 'negative', 'very negative'].includes(
+      checkAnswer(answer),
+    ),
+  )
+}
+
+testAll()
```
frosas: Works great! I see you ended up not using that `eightballValid` array, I think in the next couple of classes we'll learn how to take advantage of using arrays to make exercises like this look even nicer 😎

PR #438, opened 2020-02-29
all exercises done

sabia-m:


week-1/Homework/mandatory/1-syntax-errors.js
```diff
@@ -1,18 +1,22 @@
...
-    remainder = a %% b;
+    remainder = a % b;
+}
+
+
 
     // Use string interpolation here
-    return "The remainder is %{remainder}"
-}
+    return `The remainder is ${getRemainder}`
```
theseafaringturtle: Replacing the quotes with ticks and % with $ is right, but the return should be inside the function, and the variable name is remainder.

week-1/Homework/mandatory/1-syntax-errors.js
```diff
@@ -1,18 +1,22 @@
...
 
-function addNumbers(a b c) {
+function addNumbers(a, b, c) {
     return a + b + c;
 }
 
-function introduceMe(name, age)
-return "Hello, my name is " + name "and I am " age + "years old";
+function introduceMe(name, age) {
+return "Hello, my name is " + name + "and I am" + age + "years old";
```
theseafaringturtle: The fix is correct, you only forgot the spaces inside the quotes

week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -42,20 +42,58 @@ My sources say no.
...
+  if (indexValue <= 4) {
+  return `very positive`}
+ else if  (indexValue > 4 & indexValue <=9){
+   return `positive`}
+else if (indexValue > 9 & indexValue <= 14) {
+  return `negative`}
+  else {
+    return `very negative`
+  }
+}
```
theseafaringturtle: Nice work, the only suggestion I have is to format it to make it more readable. If you wish to do it automatically the shortcut should be Alt+Shift+F on Windows or Ctrl+Shift+I on Linux

PR #439, opened 2020-02-29
Kyaw All js-week1-exercises done.

kiowd:


week-1/Homework/extra/1-currency-conversion.js
```diff
@@ -16,20 +19,24 @@ function convertToUSD() {}
   Find a way to add 1% to all currency conversions (think about the DRY principle)
 */
 
-function convertToBRL() {}
+function convertToBRL(amount) {
+  let exchangeRate = amount * 5.7;
```
illicitonion: This is great, but I have a couple of minor naming suggestions :)
My immediate thought when I read the variable name `exchangeRate` is that it's the number of BRL per GBP, i.e. 5.7 - maybe this variable could be called something like `amountInBRL` instead?

Similarly, instead of `add_commission`, perhaps that could be `brlIncludingCommission`?

week-1/Homework/extra/2-piping.js
```diff
@@ -16,26 +16,34 @@
   the final result to the variable goodCode
 */
 
-function add() {
-
+function add(a, b) {
+  let decimal_allowed = a + b;
+  let precision = decimal_allowed.toFixed(1) / 1;
```
illicitonion: One interesting question is _where_ to do the rounding here.

In general, we try to round numbers as late as we can, ideally just when we're formatting, rather than intermediate results. If you imagine:
```javascript
multiply(4, add(1.2, 1.05))
```
if we round in `add`, the answer is `4`; if don't, the answer is `5` - by leaving the rounding to the very end, we get more accurate results.

week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -42,20 +42,75 @@ My sources say no.
...
+console.log(getNumber());
 
 // This should log "The ball has shaken!"
 // and return the answer.
-function shakeBall() {}
+function shakeBall() {
+  
+  switch (randomNumber) {
+    case 1:
+      answer = "It is certain.";
```
illicitonion: Your magic 8 ball works well, but it only has one answer for each sentiment! This passes our tests, because testing randomness is hard (good lesson for the future!), but it would be great to expand this code to be able to return all 20 possible answers. Can you give it a go?

Next week we're going to learn some things about arrays and objects which will help us do this with less code, too, but it's worth practicing with what you already know :)

PR #440, opened 2020-02-29
Harris j sweek1

HarrisSheraz20:


week-1/Homework/mandatory/3-function-output.js
```diff
@@ -1,15 +1,28 @@
...
+// Add comments to explain what this function does. You're meant to use Google!,  
 function getNumber() {
     return Math.random() * 10;
 }
+// A Math.random(), only displays "floting points"(witch is baseicly just numbers using to show result), 
+// numbers that are between 0 -1, 0 is shown but not 1. With approximantly unchangeble distribution.
+// You can then change those numbers to whatever you want. Then once you changed the scale.
+// It chouses a number between the scale at random, again not incloding the maxiem number but the minimem (in this case 0)
+// witch can also be included. However, in this case there is a " * 10" on line 3.
+// This means that what ever the algorithom choses the number, it will get multipiled by 10.
```
illicitonion: This is a really good explanation of what this function is doing :) One more thing to think about - what are the limits of what this function may return? You've said that that `Math.random()` will return a value between 0 and 1 - what values may `getNumber()` return?

week-1/Homework/mandatory/4-tax.js
```diff
@@ -17,7 +20,11 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function formatCurrency() {}
+function formatCurrency(productPrice) {
```
illicitonion: This passes the tests, which is great :) Unfortunately, the tests were wrong!! (Sorry! Though this happens in real life too, it's really inconvenient!)

Please you update the last three lines of this file to instead be:
```javascript
test("formatCurrency function - case 1 works", formatCurrency(15) === "£18.00")
test("formatCurrency function - case 2 works", formatCurrency(17.5) === "£21.00")
test("formatCurrency function - case 3 works", formatCurrency(34) === "£40.80")
```

and then make them pass? If you read through the comment above, it mentions something about decimal places you'll need to make work :)

week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -45,17 +45,146 @@ Very doubtful.
 
 // This should log "The ball has shaken!"
 // and return the answer.
-function shakeBall() {}
+
+//Created a switch
```
illicitonion: This 100% works, but you may have felt as you were writing it that it was a little repetitive (you needed to write down every string twice, which took quite a lot of typing!)

In week three we're going to learn more about arrays and objects which would help us write this with less code - give some thought as to how you may want rewrite this to be smaller, as we learn those things next week!

PR #441, opened 2020-02-29
the sorting game exercise

aldeez988:


week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -39,6 +39,16 @@ Write a function that:
 
 function formatPercentage(arr) {}
 
+/*
+Inside the array ages we have mix of data, use this data to write your sort function using what
+ you have learnt from the sorting game the function should:
+- takes an array of mixed data types as input
+- remove the unwanted data without using the built-in javascript filter method
```
sansaid: Maybe we should let the students decide if they want to do this or not? What would be the benefit of asking them to use another method over filter?

sansaid: Maybe we should let the students decide if they want to do this or not? What would be the benefit of asking them to use another method over filter?

PR #443, opened 2020-02-29
Week1 alex

AlexandruBudaca:


week-1/Homework/extra/2-piping.js
```diff
@@ -16,43 +16,51 @@
   the final result to the variable goodCode
 */
 
-function add() {
-
+function add(a, b) {
+  let sum = a + b;
+  return Number(sum.toFixed(2));
```
illicitonion: One interesting question is _where_ to do the rounding here.

In general, we try to round numbers as late as we can, ideally just when we're formatting, rather than intermediate results. If you imagine:
```javascript
multiply(1000, add(1, 0.001))
```
if we round in `add`, the answer is `1000`; if don't, the answer is `1001` - by leaving the rounding to the very end, we get more accurate results.

week-1/Homework/extra/2-piping.js
```diff
@@ -16,43 +16,51 @@
...
-
+function multiply(c, d) {
+  let result = c * d;
+  return result;
 }
 
-function format() {
-
+function format(f) {
+  return "£" + f;
```
illicitonion: This would probably be a really good place to do rounding; something like:
```javascript
return `£${sum.toFixed(2)}`;
```
because we generally want to round money for display, but not during calculation!

Then you'd need to pass the expression for the amount of money as an argument to `format`, rather than always passing the empty string :)

week-1/Homework/extra/2-piping.js
```diff
@@ -16,43 +16,51 @@
...
 }
 
-const startingValue = 2
+const startingValue = `${format("")}${add(1, 3) + 10 * multiply(1, 2)}`;
```
illicitonion: This works, but can you try writing both this, and `calcAll`, without using `+` or `*`? Just using the functions you defined.

week-1/Homework/mandatory/3-function-output.js
```diff
@@ -1,31 +1,43 @@
 // Add comments to explain what this function does. You're meant to use Google!
 function getNumber() {
-    return Math.random() * 10;
+  return Math.random() * 10;
 }
+// will return a number between 0 and 1 but not 1
```
illicitonion: That's a good explanation of what `Math.random()` does - can you also explain what `getNumber()` does?

week-1/Homework/mandatory/6-playing-computer.js
```diff
@@ -8,11 +8,17 @@
 
   1. This program throws an error. Why? (If you can't find it, try executing it).
   2. Remove the line that throws the error.
+  // The line 31 throws the error because we don't have a defined variable b.
   3. What is printed to the console?
+  // The console prints first the variabels x and a, then the numbers: 4,9,6,13,8.
   4. How many times is "f1" called?
+  //Three times
```
illicitonion: Questions 1-3 are right here, but 4-7 are slightly off - can you give them another go? It may help to write out the values of all of the variables for each iteration, i.e.
```
i=0: x=2, a=3, f2(a=0, b=2), d=4
i=1: ...
...
```

PR #444, opened 2020-02-29
Me week one js exrecises

yusuf963:
Hi, here is my work for this week
thank you

### DIFF
--- a/week-1/Homework/mandatory/5-magic-8-ball.js
+++ b/week-1/Homework/mandatory/5-magic-8-ball.js
@@ -45,22 +45,45 @@ Very doubtful.
 
 // This should log "The ball has shaken!"
 // and return the answer.
-function shakeBall() {}
+let possibleAnswer = ["It is certain","It is decidedly so","Without a doubt","Yes - definitely","You may rely on it",
+  "As I see it, yes","Most likely","Outlook good","Yes","Signs point to yes","Reply hazy, try again","Ask again later",
+  "Better not tell you now","Cannot predict now","Concentrate and ask again","Don't count on it","My reply is no",
+  "My sources say no","Outlook not so good","Very doubtful",
+];
+
+function shakeBall() {
+  let answer = possibleAnswer[Math.floor(Math.random() * possibleAnswer.length)];
+  console.log(`The ball has shaken!`)
+  return answer;
+}
 
 // The answer should come from shaking the ball
-let answer;
+let answer = shakeBall();
+console.log(answer);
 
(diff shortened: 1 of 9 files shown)


week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -45,22 +45,58 @@ Very doubtful.
...
+console.log(answer);
 
 // When checking the answer, we should tell someone if the answer is
 // - very positive
 // - positive
 // - negative
 // - very negative
-function checkAnswer() {}
+function checkAnswer(answer) {
+  let getIndexNum = possibleAnswer.indexOf(answer);
```
illicitonion: This line doesn't quite work, because `possibleAnswer` is a local variable which only exists inside the `shakeBall` function, so you can't see it here. If you move `possibleAnswer` outside of `shakeBall`, so that it exists at the top-level of this file, this should work fine :)

week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -45,22 +45,58 @@ Very doubtful.
 
 // This should log "The ball has shaken!"
 // and return the answer.
-function shakeBall() {}
+
+function shakeBall() {
+  let possibleAnswer = [
```
illicitonion: This solution works really well, nicely done!

A couple of things to think about (some of which we'll learn more about in week three) which could make this code even easier to work with in the future:
This long list combines a few different types of response together. It may be useful to add a comment at the start of each section, i.e.
```javascript
let allAnswers = [
        // Very positive
        'It is certain.',
	'It is decidedly so.',
	'Without a doubt.',
	'Yes - definitely.',
	'You may rely on it.',

        // positive
	'As I see it, yes.',
	'Most likely.',
	'Outlook good.',
	'Yes.',
	'Signs point to yes.',

        // negative
	'Reply hazy, try again.',
	'Ask again later.',
	'Better not tell you now.',
	'Cannot predict now.',
	'Concentrate and ask again.',

        // very negative
	"Don't count on it.",
	'My reply is no.',
	'My sources say no.',
	'Outlook not so good.',
	'Very doubtful.'
];
```
so that it's easy to see which responses are which.

Even then, though, if you add another answer, you need to change several pieces of code (the list, and the indexes in each if statement). In week 3 we'll learn more about arrays, and about objects, which could make it even easier to work with. Have a think about how they could help in this problem so that adding a new answer only needs one code change when we're learning about them!

PR #445, opened 2020-02-29
Add reading list for JS1 week 2

ismaelgt:


sansaid:
Hey @ismaelgt , can we add reading material for merge sort as well since it's the faster and most common sorting algorithm than bubble sort? I haven't read the bubble sort reading material, but if comparisons to other sorting methods are in there, then you can ignore this message. :)

week-2/Homework/README.md
```diff
@@ -15,3 +15,9 @@ For your homework, we'll be using [**test driven development**](https://medium.c
 To challenge yourself even further, you can get ahead for next week's class by reading some of the below documentation:
 
 - [ ] More on the node REPL: https://hackernoon.com/know-node-repl-better-dbd15bca0af6
+- [ ] `while` loops: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/while
+- [ ] `for` loops: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for
+- [ ] Arrays: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array
+- [ ] What are algorithms: https://www.khanacademy.org/computing/computer-science/algorithms/intro-to-algorithms/v/what-are-algorithms
+- [ ] Bubble sort: https://humanwhocodes.com/blog/2009/05/26/computer-science-in-javascript-bubble-sort/
```
frosas: Are students supposed to implement a sort algorithm? I find it way too advanced. What about just using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort?

sansaid: @frosas , the idea here is that they would implement the sorting algorithm they came up with in the sorting game at the start of the class. It can be a "dumb" algorithm, but it's a medium to get them thinking algorithmically in JS. @ismaelgt has assigned them reading material on sorting algorithms and they will also be taught about the `sort` method, but this is only a thought exercise. We won't be assessing them on the efficiency of the algorithm, but on how much they've applied from what we've taught them in class and seeing if they apply anything they learn outside of class.

frosas: Looks like you have it well planned out 👌

ismaelgt:
@sansaid The reason why I want students to read about Bubble Sort is its simplicity and pedagogical value. I understand it's the least efficient one, but I am not interested in that. We could invite them to read about Merge Sort but that implies they'll need to understand recursion. This is why I haven't done it. Happy to add it if you think this is a good idea.

PR #446, opened 2020-02-29
Js exercises ola

leealex88:
need to improve the logical approach to the problems.


week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -55,12 +90,27 @@ let answer;
 // - positive
 // - negative
 // - very negative
-function checkAnswer() {}
+function checkAnswer(randomArray) {
+  console.log('second function', randomAnswear)
+  if (randomArray === answears[0]) {
```
illicitonion: This is the right idea, but unfortunately one thing that we learnt in week 2 is that you can't compare arrays with `===`, so this always returns `very negative` :( This tricked our tests, though, because it's really hard to write tests which test random behaviour (a problem we see in professional software engineering too!!)

There are a few ways that we could do this kind of check, and we're going to learn about them more in week 3 (where we'll learn more about arrays, and I think objects too) :)

An example  is that we could use a loop to go through each array, and look for `randomAnswear` in them. Something like:
```javascript
for (let i = 0; i < 5; ++i) {
  if (answears[i].includes(randomAnswear)) {
    if (i === 0) {
      return "Very positive";
    } else if (i === 1) {
      return "positive";
    }
    ...
}
```

week-2/D-predicates/exercise.js
```diff
@@ -7,12 +7,14 @@
 
 // Finish the predicate function to test if the passed number is negative (less than zero)
 function isNegative(number) {
-
+  return number === 5 && number < 0
```
illicitonion: This function always returns `false`, because a number can't both be 5 _and_ be less than 0 - can you give this one another go? :)

week-1/Homework/extra/2-piping.js
```diff
@@ -1,58 +1,58 @@
...
+  return "£" + num
 }
 
-const startingValue = 2
+// const startingValue = 2
 
-// Why can this code be seen as bad practice? Comment your answer.
-let badCode = 
+// // Why can this code be seen as bad practice? Comment your answer.
+// let badCode =
```
illicitonion: You've got the functions right, and they work - can you add in some code here which calls the functions, to prove that they work and make the tests pass? :) As a reminder of how functions work, you can write:
`format(multiply(2, 5))` to call multiply, and then format the return value of the function.

week-1/Homework/extra/2-piping.js
```diff
@@ -1,58 +1,58 @@
...
+//   -add 10 to startingValue -
+//   multiply the result by 2 -
+//   format it
+// 3. Write a more readable version of what you wrote in step 2 under the BETTER PRACTICE comment.Assign
+// the final result to the variable goodCode
+
+
+function add(num1, num2) {
+  const total = num1 + num2
+  return Number(total.toFixed(2))
```
illicitonion: This is right, and works well, but one thing to think about is _when_ to round numbers. It's often useful to keep numbers as precise as we can during things like adding and multiplying, and only rounding them (with `.toFixed`) when we're about to show them to people. Otherwise we'll keep introducing more small rounding errors each time we do anything with numbers. So I'd think about removing the `Number` and `toFixed` calls here, and instead having one in the `format` function.

Like I say, this is right, and does what the question asks, but it's something to think about :)

leealex88:
.

leealex88:
Closing this reviewed week-1 pull request. In order to open a new one for week 2 homework, because both are on the same branch.

leealex88:
Closing this reviewed week-1 pull request. In order to open a new one for week 2 homework, because both are on the same branch.

PR #448, opened 2020-03-01
node js/Bani

banirezaie:


week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -14,38 +14,23 @@ and have different levels of positivity or negativity.
...
-My sources say no.
-Outlook not so good.
-Very doubtful.
+// This should log "The ball has shaken!"
 */
 
-// This should log "The ball has shaken!"
-// and return the answer.
-function shakeBall() {}
+function shakeBall() {
```
illicitonion: You have all of the pieces you need here, but they're not _quite_ in the right places :) If you check the comments again, `shakeBall` currently returns the category of answer (e.g ."very positive") but _should_ return the actual answer (e.g. "It is certain."), and `checkAnswer` currently returns the actual answer (e.g. "It is certain.") but _should_ return what the category of answer was (e.g. "very positive"). Can you try swapping things around so that they work that way around?

week-1/Homework/mandatory/3-function-output.js
```diff
@@ -1,31 +1,43 @@
 // Add comments to explain what this function does. You're meant to use Google!
 function getNumber() {
-    return Math.random() * 10;
+  return Math.random() * 10;
+  // The Math.random() function returns a floating - point, pseudo - random number in the range 0 to less than 1(inclusive of 0, but not 1) with approximately uniform distribution over that range — which you can then scale to your desired range.
```
illicitonion: These are good definitions of what `Math.random` and `String.concat` do :) Could you also add a comment about what _these functions_ (`getNumber` and `s`) do, because they're slightly different?

For `getNumber`, one thing to think about is what the lowest and highest values it can return are, and include that in your comment :)

week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -55,8 +40,87 @@ let answer;
...
+      return "My sources say no.";
+    } else if (number === 4) {
+      return "Outlook not so good";
+    } else if (number === 5) {
+      return "Very doubtful";
+    }
+  }
+}
+// and return the answer.
+console.log(`level of posibility : ${answer}`);
```
illicitonion: This code is right, but it's slightly in the wrong place. Remember how when we define a function, we don't run any of the code in it until someone calls it, but code outside of functions gets run straight away - so this code gets called _before_ anyone calls `shakeBall` (which is what assigns a variable to the `answer` variable), which means that `answer` isn't set here (which is why you'll see "level of posibility : undefined" logged).

If you move this code to happen in a function where we know `answer` has a value, you should see the right value get logged :)

banirezaie: Hello,
Because of change in the master branch I had to make a new branch, name is: week 1 final changes. 
I fixed these mistakes there.
thank you for your comments.

banirezaie: sorry I deleted that because of a mistake, I will upload it here shortly, but I want to ask a question and that is why I couldn't answer with or(||) tag rather than repeat one command over and over.

banirezaie: it is a new pull request, named:
changes are done.

PR #454, opened 2020-03-01
Java s jreo123

jreo123:


week-1/Homework/mandatory/3-function-output.js
```diff
@@ -1,31 +1,43 @@
 // Add comments to explain what this function does. You're meant to use Google!
+//The Math.random() function returns a floating-point, pseudo-random number in the range 0 to less than 1
```
illicitonion: This is a great description of `Math.random()` - can you also explain what the function `getNumber()` does? (It's very similar!) In particular, what's the lowest and highest value `getNumber()` may return?

week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -45,17 +45,21 @@ Very doubtful.
 
 // This should log "The ball has shaken!"
 // and return the answer.
-function shakeBall() {}
+function shakeBall() {
```
illicitonion: This is a good start, but it looks like this one isn't quite done :) Can you try to finish it off? If you're stuck, we're always around on slack, or at homework club, to help out!

PR #458, opened 2020-03-02
js-1 / Week _1 + Week_2/Min Ko

minwwgo:


sansaid:
Good job @minwwgo ! Very well done for attempting the extra exercises as well! I've left some comments, let me know if you have any questions :)

PR #462, opened 2020-03-03
JS1-week 2-Alisina

Alisina92:


illicitonion:
The first couple of exercises look great, well done :) Hopefully the other two go smoothly too!

One thing I'd encourage you to think about is consistent formatting - making sure your {s and }s line up, having the same number of spaces before and after punctuation, etc, can make it a little easier to read your code and see what's going on.

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -15,8 +15,27 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
 */
 
 function sortAges(arr) {
-
-}
+   let numberArr =[];// new array to store numbera
+  for(let i = 0; i < arr.length;i++){ // loops throw the length of an array
+        if(typeof arr[i]!== typeof'string'){// checks if array elements are number
```
illicitonion: This works for the test cases we've been given, but it could be more general - what if there was a boolean in the list?

It may be more comprehensive to test `if (typeof arr[i] === "number") {` rather than looking for strings :)

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -15,8 +15,27 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
...
-
-}
+   let numberArr =[];// new array to store numbera
+  for(let i = 0; i < arr.length;i++){ // loops throw the length of an array
+        if(typeof arr[i]!== typeof'string'){// checks if array elements are number
+           numberArr.push(arr[i]);// push the numbers into new array
+        }
+      }
+      let swapped; // variable to hold a swap
+      for(let j= numberArr.length; j > 0;j--){// loops though the array desending order
```
illicitonion: We have a convention (not a rule! Nothing goes wrong if you don't follow it, but it helps other programmers understand your code quickly when reading it) that we start loop counters from `i`, and go on in the alphabet in inner loops.

Because we tend to use `i` for the outer-most `for` loop, when I first read this code I assumed that it was inside another loop (which uses `i`). This definitely isn't a programming rule, or anything wrong, but because that's a convention we have, it may be more clear to other readers to use `i` and `j` here, rather than `j` and `k`

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -7,36 +7,36 @@
...
 
-let x = 2;
-let a = 6;
+let x = 2;// variable x assigned to 2
+let a = 6;// variable a assigned to 6
 
-const f1 = function(a, b) {
-  return a + b;
+const f1 = function(a, b) {//f1 is a function with a prameters  a and b
+  return a + b;// sums a and returns the result  However this would return an error because 'b' is not defined
```
illicitonion: Are you sure that `b` isn't defined? Take another look at the arguments to the function :)

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -7,36 +7,36 @@
   Answer the following questions:
 
   1. This program throws an error. Why? (If you can't find it, try executing it).
-  2. Remove the line that throws the error.
+  2. Remove the line that throws the error. 
   3. What is printed to the console?
```
illicitonion: Your comments of what's happening in each part of the file are good :) Can you also add the answers to these questions to explain how the whole program behaves?

PR #463, opened 2020-03-03
Js1-exercises-week1-tm

tanianm:


week-1/Homework/mandatory/3-function-output.js
```diff
@@ -11,8 +11,15 @@ function s(w1, w2) {
 function concatenate(firstWord, secondWord, thirdWord) {
     // Write the body of this function to concatenate three words together
     // Look at the test case below to understand what to expect in return 
+
```
illicitonion: Your concatenate function works well! There are a couple of other exercises to do in this file at the top (have a look at the comments - can you give them a go too? :))

PR #464, opened 2020-03-03
Fix not comparing arrays properly

frosas:
Some context: https://codeyourfuture.slack.com/archives/CSE061L6L/p1583158208116000


sansaid:
Thanks @frosas ! Apologies for the plethora of mistakes. Exercise 2 has an `arraysEqual` function that we used to compare arrays for consistency. Is that alright?

PR #465, opened 2020-03-03
Week 2

ferhatu:


week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -4,7 +4,7 @@
 function mood() {
   let isHappy = true;
 
-  if (isHappy) {
+  if (typeof isHappy !== "boolean") {
```
frosas: It may not be very clear what's supposed to be fixed in this function but I don't think we want to check the type of value of `isHappy` (otherwise when `isHappy` is `false` we would return `"I am happy"`, which wouldn't be correct).

What else could we change in the function so that it makes sense (and makes the test pass)?

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -13,9 +13,9 @@ function mood() {
 
 function greaterThan10() {
   let num = 10;
-  let isBigEnough;
+  let isBigEnough = 12;
 
-  if (isBigEnough) {
+  if (isBigEnough >= num) {
```
frosas: Given its name, the variable `isBigEnough` implies it represents a boolean (the response to "is it big enough?").

Without changing the `if` condition, how would change that `let isBigEnough` to contain the right value based on `num`?

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -24,21 +24,21 @@ function greaterThan10() {
 
 function sortArray() {
   let letters = ["a", "n", "c", "e", "z", "f"];
-  let sortedLetters;
+  let sortedLetters = letters.sort();
```
frosas: 👍

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -24,21 +24,21 @@ function greaterThan10() {
...
 
   return sortedLetters;
 }
 
 function first5() {
   let numbers = [1, 2, 3, 4, 5, 6, 7, 8];
-  let sliced;
+  let sliced = numbers.slice(0, 5);
```
frosas: 👍

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -24,21 +24,21 @@ function greaterThan10() {
...
 
-function get3rdIndex(arr) {
-  let index = 3;
-  let element;
+function get3rdIndex() {
+  let index = ["fruit", "banana", "apple", "strawberry", "raspberry"];
+  let element = index[3];
 
   return element;
 }
```
frosas: If you look at how `get3rdIndex()` is called in the test below, you'll see the array of strings is provided as a parameter.

How would you change the function so that it uses that parameter?

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -5,7 +5,24 @@ Write a function that:
...
+}
+function tidyUpString1() {
+  let strArr = [" /Sanyia ", " Michael ", "AnTHonY ", "   Tim   "];
+  for (var i = 0; i < strArr.length; i++) {
+    strArr[i] = strArr[i].trim();
+    strArr[i] = strArr[i].replace(/\//g, "");
+    strArr[i] = strArr[i].toLowerCase();
+  }
+  return strArr;
+}
```
frosas: Notice that, similarly to one of the previous functions, `tidyUpString1()` is called in the test below with a parameter that contain the list of strings so you shouldn't recreate it inside of the function.

Other than that the function looks good 👍

Btw, I guess you created that new function to try something but don't forget to remove code like this once you're done with it :)

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -15,7 +32,13 @@ Complete the function to check if the variable `num` satisfies the following req
...
 */
 
-function validate(num) {}
+function validate(num) {
+  if (typeof num === "number" && num <= 100 && num % 2 == 0) {
+    return true;
+  } else {
+    return false;
+  }
+}
```
frosas: 👍

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -26,7 +49,11 @@ The function must:
 */
 
 function remove(arr, index) {
-  return; // complete this statement
+  let first = arr.slice(0, index);
+  let rest = arr.slice(index + 1);
+  newArr = first.concat(rest);
+
+  return newArr; // complete this statement
```
frosas: Nice 👍

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -38,7 +65,16 @@ Write a function that:
...
+  for (var i = 0; i < arr.length; i++) {
+    if (arr[i] > 100) {
+      arr[i] = 100;
+    }
+    if (!Number.isInteger(arr[i])) {
+      arr[i] = arr[i].toFixed(2);
+    }
+    arr[i] = `${arr[i]}%`;
+  }
+  return arr;
```
frosas: Looks great.

What would happen if we didn't have that `if (!Number.isInteger(arr[i]))`?

Also, can you think of a way to do the same returning a new array instead of modifying `arr`?

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -14,9 +14,19 @@
   6. What value does the "a" parameter take in the first "f1" call?
   7. What is the value of the "a" outer variable when "f1" is called for the first time?
 */
+/*This program throws an error because "b" is not defined as a function argument before
+ So it shoul be defined before execution.
```
frosas: 👍

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -14,9 +14,19 @@
...
+ Console is printing value of x, a, b
```
frosas: I think what 2. asks for is the actual strings that are printed to the console 😉

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -14,9 +14,19 @@
   6. What value does the "a" parameter take in the first "f1" call?
   7. What is the value of the "a" outer variable when "f1" is called for the first time?
 */
+/*This program throws an error because "b" is not defined as a function argument before
+ So it shoul be defined before execution.
+ Console is printing value of x, a, b 
+ "f2" is called one time
```
frosas: I think it's called more than one time

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -14,9 +14,19 @@
...
+ "f1" is called one time
```
frosas: Same here

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -14,9 +14,19 @@
...
+ "a" takes 7 in the first "f1" call
+ it is 6 (outer variable)
```
frosas: 👍

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -14,9 +14,19 @@
   6. What value does the "a" parameter take in the first "f1" call?
   7. What is the value of the "a" outer variable when "f1" is called for the first time?
 */
+/*This program throws an error because "b" is not defined as a function argument before
+ So it shoul be defined before execution.
+ Console is printing value of x, a, b 
+ "f2" is called one time
+ "f1" is called one time
+ "a" takes 7 in the first "f1" call
```
frosas: I don't think it's 7

frosas: I don't think it's 7

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -14,44 +14,75 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
...
+  for (let i = 0; i < arr.length - 1; i++) {
+    for (let j = 0; j < arr.length - 1 - i; j++) {
+      if (arr[j] > arr[j + 1]) {
+        let tmp = arr[j];
+        arr[j] = arr[j + 1];
+        arr[j + 1] = tmp;
+      }
+    }
+  }
+  return arr;
```
frosas: 👍

PR #467, opened 2020-03-04
Create review.js

colevandersWands:
this script sits at the top level of the directory and creates review files for each subdirectory to help students study their exercises, and coaches to review their work.

It creates a REVIEW.md & an index.html in each sub-directory.  the REVIEW.md is helpful for reviewing on GitHub and the index.html's make it easy for students to study their code in the devtools or Js Tutor

This script is designed to work with console.assert.  if any files contain asserts, it will be flagged as passing if all assertions pass, or failing if one fails.  In the REVIEW.md's, all assertion messages are rendered above the code with PASS or FAIL

At the top level directory (or in whichever directory you call the script from) a report.json is also created.  this is useful for centralizing review on a single web page or automating parts of the review process

here's [a little demo repository](https://github.com/colevandersWands/review-script) to show what a reviewed repo is like

hope it helps!


ChrisOwen101:
Thanks @colevandersWands! I'll take a look

colevandersWands:
@ChrisOwen101 , I've come up with a few other (simpler) versions of the script since this PR. we'll start testing it them week.  if you like the idea we can find a version that works for this repo

sansaid:
This is super @colevandersWands - thank you for doing this! We created a minimal testing framework that doesn't use `console.assert`, but I would be open to changing it so we can create the review doc. This is especially great because I've actually discovered one or two students who may have submitted their homework before allowing the tests to pass.

colevandersWands:
Thank you @sansaid!  it's taken a few years to get here.  

I'd used small dedicated testing functions like yours before but found that console.assert is much better in the long run.  because it's native it's more portable and plays nicely with the devtools and JS Tutor.  It's also easy to overwrite the assert method to evaluate the same code in different study tools

We're using a simpler version of this script for the first time in two exercises for next week, i'll let you know how it goes: https://github.com/HackYourFutureBelgium/getters-and-setters/, https://github.com/HackYourFutureBelgium/functions-to-methods

PR #469, opened 2020-03-04
Homework mandatory is done

AlexeyBrambalov:
Can`t understand the extra task. Need examples, and the function name is not correct.


week-2/F-logical-operators/test.js
```diff
@@ -0,0 +1,23 @@
...
+
+}
+
+
+
+
+assert(Check("Al23", "managr")==="Username valid");
+assert(Check("Alexey23", "managr")==="Username valid");
+assert(Check("alexel23", "managr")==="Username invalid");
+
```
jamesgorrie: Could you explain to me why you think these tests should pass?

week-2/F-logical-operators/test2.js
```diff
@@ -0,0 +1,12 @@
+function Sum(n){
```
jamesgorrie: `function` names are usually [`lowerCamelCase`](https://wiki.c2.com/?CamelCase). Class names are [`PascalCase` / `UpperCamelCase`](https://wiki.c2.com/?PascalCase).

week-2/F-logical-operators/test2.js
```diff
@@ -0,0 +1,12 @@
...
+    let i=0
+    let sum=0
+    while (i <= n){
+        
+        sum+=i
+        i++
+    }
+    return sum
+}
```
jamesgorrie: Can you think of a way to do this with `reduce`?

while is useful but can introduce infinite loops quite easily, especially if not tested.

week-2/Homework/extra/4-radio-stations.js
```diff
@@ -15,18 +15,56 @@
 
 // `getAllFrequencies` goes here
 
+let arr=[];
```
jamesgorrie: Do you think this should be in the global scope? If I had another function, it might be able to mutate this, which could cause bugs. Where might it better live?

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -1,21 +1,19 @@
 // The below functions are syntactically correct but not outputting the right results.
 // Look at the tests and see how you can fix them.
+let isHappy = true;
```
jamesgorrie: Can you explain what you're doing here?

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -1,21 +1,19 @@
...
   if (isHappy) {
     return "I am happy";
   } else {
     return "I am not happy";
   }
 }
 
+
 function greaterThan10() {
   let num = 10;
```
jamesgorrie: If we do it this way, do we need to keep this `variable`?

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -24,21 +22,23 @@ function greaterThan10() {
 
 function sortArray() {
   let letters = ["a", "n", "c", "e", "z", "f"];
-  let sortedLetters;
-
+  let sortedLetters = letters.sort().toString();
```
jamesgorrie: What does running `toString()` on an array do?

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -63,9 +63,9 @@ test(
...
 test(
   "sortArray function works",
-  sortArray() === ["a", "c", "e", "f", "n", "z"]
+  sortArray() === [ 'a', 'c', 'e', 'f', 'n', 'z' ].toString()
 );
-test("first5 function works", first5() === [1, 2, 3, 4, 5]);
+test("first5 function works", first5() === [1, 2, 3, 4, 5].toString());
 
 test(
   "get3rdIndex function works - case 1",
```
jamesgorrie: You run `toString()` on a lot of arrays here - can you tell me why and what you think it's doing?

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -5,7 +5,14 @@ Write a function that:
 - removes any forward slashes (/) in the strings
 - makes the string all lowercase
 */
-function tidyUpString(strArr) {}
+function tidyUpString(strArr) {
+  for( let i=0; i< strArr.length; i++){
+    strArr[i]= strArr[i].trim().replace('/','').toLowerCase()
```
jamesgorrie: What if there is more than 1 `/` e.g. `/path/to/files`?

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -26,7 +40,9 @@ The function must:
 */
 
 function remove(arr, index) {
-  return; // complete this statement
+  let res;
+  res = arr.filter(item => item !== arr[index]);
```
jamesgorrie: You could just say
```js
const res = arr.filter(item => item !== arr[index]);
```

or 
```
return arr.filter(item => item !== arr[index]); 
```

I would say they are all very clear.

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -14,7 +14,33 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
 "think" like a computer and practice your knowledge of basic JavaScript.
 */
 
+let newArr=[];
```
jamesgorrie: This array could live in function scope to help prevent bugs 🐛

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -14,7 +14,33 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
...
+      {
+          if(newArr[i] > newArr[j]) //comparing first and next element
+          {
+              let tempValue = newArr[j];   
+              newArr[j] = newArr[i];
+              newArr[i] = tempValue;
+          }
+      }
+  }
+
```
jamesgorrie: Try using `map` and `sort` here.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

week-2/F-logical-operators/test.js
AlexeyBrambalov: First test will be failed, there should be "manager"  usertype to pass.
Second  test got a proper username so it will pass all username is valid
Third got the wrong username with lowercase first letter and will be invalid  and test will be passed

week-2/Homework/mandatory/4-sorting-algorithm.js
AlexeyBrambalov: Task says:  (without using the built-in javascript
sort method 😎)

week-2/Homework/mandatory/2-function-creation.js
AlexeyBrambalov: strArr[i].trim().replace(/\//g,'').toLowerCase()

week-2/Homework/mandatory/1-fix-functions.js
AlexeyBrambalov: To compare arrays.  It turns the array into a string

week-2/Homework/mandatory/1-fix-functions.js
AlexeyBrambalov: Without it test will fail as arrays are not able to compare

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -63,9 +63,9 @@ test(
 );
 test(
   "sortArray function works",
-  sortArray() === ["a", "c", "e", "f", "n", "z"]
+  sortArray() === [ 'a', 'c', 'e', 'f', 'n', 'z' ].toString()
```
jamesgorrie:
```suggestion
  sortArray().toString() === [ 'a', 'c', 'e', 'f', 'n', 'z' ].toString()
```

week-2/Homework/mandatory/1-fix-functions.js
jamesgorrie: I see. It would be better to write the test turning it into a string then, and leave these functions to be more accurate.

[This is an example](https://github.com/CodeYourFuture/js-exercises/pull/469/files#r394408489) - not your fault, the tests just needed to be written better.

week-2/Homework/mandatory/2-function-creation.js
jamesgorrie: I've just tried to run that, I get an error
https://repl.it/repls/DeepIllustriousBrackets

week-2/Homework/mandatory/4-sorting-algorithm.js
jamesgorrie: Thanks! I should have read it first 😸

PR #474, opened 2020-03-06
Ellie-week2-js

ellietms:
Mandatory HW


week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -37,9 +45,21 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(arr) {
-  
-}
+function formatPercentage(arr){
+function mystr(strarr) {
```
illicitonion: This works well, well done! One thing worth thinking about a bit is naming here.
`mystr` isn't the most descriptive name - perhaps this could be called `formatOnePercentage` or something similar?
`strarr` is actually a little misleading as a parameter name, as are `Arr` and `newArr` for variable names, as they suggest that the variables are arrays when they're not. Some more descriptive names may be `number`, `numberString`, and `formattedPercentage`, or something along those lines... Something to think about :)

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -7,12 +7,13 @@
...
 
   1. This program throws an error. Why? (If you can't find it, try executing it).
+   //answer :console.log(b);
   2. Remove the line that throws the error.
-  3. What is printed to the console?
-  4. How many times is "f1" called?
-  5. How many times is "f2" called?
-  6. What value does the "a" parameter take in the first "f1" call?
-  7. What is the value of the "a" outer variable when "f1" is called for the first time?
+  3. What is printed to the console?//answer: b is not defined
```
illicitonion: This is right while that line is still in the program. After removing it, what would it print?

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -7,12 +7,13 @@
...
+  4. How many times is "f1" called? //answer : 3 times
```
illicitonion: Unfortunately 4-7 are all slightly off - can you give them another go? It may help to get some paper and write down each of the loop iterations, filling in the ?s for each iteration:
```
i = 0: a = ?, i % 2 = ?, d = f2(?, ?) = ?, e = f1(?, ?) = ?, console.log(?)
i = 1: a = ?, ...
```

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -15,9 +15,23 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
 */
 
 function sortAges(arr) {
-
+  for(var i=0;i<arr.length;i++){
```
illicitonion: This works, and looks good - well done!

I'm curious - is there a reason you chose to use `var` instead of `let` for these loops?

PR #475, opened 2020-03-06
Updating week3 exercises

sansaid:
* Reformatted existing exercises to mandatory exercises
* Added tentative reading material

**Question:** is this homework enough or can we copy from Codewars to create more "extra" homework?


ChrisOwen101:
Just a question - why are we archiving the old homework? Why not just include all of it?

PR #476, opened 2020-03-06
Week2 03

cockochan:


week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -12,9 +12,14 @@ function mood() {
...
-
+  let num = 12;
+  let isBigEnough = function() {
+    if (num > 10) {
+      return true;
+    } else {
+      return false;
+    }
+  };
   if (isBigEnough) {
```
illicitonion: This is really interesting! `isBigEnough` here is the actual function, not the value of calling the function (because it isn't `isBigEnough()`) - so what we've learnt here is that like how `0` is treated like `false` in an `if` condition, and `1` is treated like `true`, a function is also treated like `true`. So here, this condition will always evaluate to `true`, not matter what value `num` has.

You probably wanted to put:
```javascript
if (isBigEnough()) {
```
but a little simpler than that would be to skip the function completely; you can even write:
```javascript
let isBigEnough = num > 10;
```
and that should work fine :)

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -5,7 +5,17 @@ Write a function that:
 - removes any forward slashes (/) in the strings
 - makes the string all lowercase
 */
-function tidyUpString(strArr) {}
+
+function tidyUpString(strArr) {
+  let i = 0;
```
illicitonion: This works, but I'd expect the `let` to be in the `for` loop to limit its scope, i.e.
```javascript
for (let i= 0; i < strArr.length; i++) {
```
is there a reason you're declaring it outside?

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -36,11 +55,20 @@ Write a function that:
 - the numbers must be rounded to 2 decimal places
 - numbers greater 100 must be replaced with 100
 */
-
 function formatPercentage(arr) {
-  
+  let strArr = [];
+  let numRound = [];
```
illicitonion: This works, and could be useful if we also want the number values as well as the strings, but we only actually use each element of this array during one iteration of the loop.

Instead of creating an array which we fill in (which may become very large, and use lots of memory, if the input array was large), we could instead use a single variable scoped to inside the loop, maybe called `numRound` which is a number type, rather than an array.

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -40,3 +41,11 @@ for (let i = 0; i < 5; ++i) {
     console.log(e);
   }
 }
+// there is no variable b declaration, i would add it to escape error,
+//but it is required to remove a line in this case i would also remove a declaration
+//and pass both variables to the function as parameters. i wrote this and then understood thatthisprogram is not only a function.
+//I have the only solution to declare a and b to run it.
```
illicitonion: This is one solution :) A few other ways we could solve this are:
* Remove the `console.log(b);` line
* Move the `console.log` calls to live inside `f2` rather than be outside it
Can you think of when each may be the best idea?

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -40,3 +41,11 @@ for (let i = 0; i < 5; ++i) {
...
+// i think f2 is called 3 times because 0 gives 0 remainder when devided by 2
+//at the same time f1 is called in 2 cases when i=1 and i= 3.
+//I confirmed it by counting logged lines 3 logs for tree log lines before calling loop with f functions and 5
+//for 3X*2+2*f1 calls.
```
illicitonion: This looks great :) Can you add your answers to questions 6 and 7 too?

week-2/Homework/mandatory/1-fix-functions.js
cockochan: Thank you. I am correcting it now,but just added () and did not go to the final nicer way because of limited time. I still need to finish my 4th exercise.Almost done but my bubble sort is not working and i figure out how to fix it i will return and try the third way:)

week-2/Homework/mandatory/2-function-creation.js
cockochan: i actually did it this way at first but during the call my colleagues seem to be unaware of possibility to declare the variable inside a loop and i later changed it.

week-2/Homework/mandatory/3-playing-computer.js
cockochan: i will do new pull request now.

week-2/Homework/mandatory/3-playing-computer.js
cockochan: when i moved it inside f2 visual studio if i want to remove unreachable code. At least one point goes to "remove console.log" version

illicitonion:
This all looks great, good job!

PR #477, opened 2020-03-06
Afshin js week 2

AKaramifar:
I finished all Exercises


week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -5,7 +5,22 @@ Write a function that:
...
+    var container = ""; // empety container for keep arry items without "/"!
+    for (let j = 0; j < newArr[i].length; j++) { // this is a loop for check all char in all items in arry!
+      if (newArr[i].charAt(j) !== "/") {
+        container += newArr[i].charAt(j); // if char is not equal "/" so add to container!
+      }
+    }
+    newArr[i] = container.toLocaleLowerCase(); // replace old value of newArr[i] to new value without any space at the begining and end of string and without any "/"! 
+  }
+  return newArr;
+}
```
theseafaringturtle: Good work. There is also a method called replace() that makes this easier, you'll learn about this on Sunday.

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -38,7 +64,13 @@ Write a function that:
 */
 
 function formatPercentage(arr) {
-  
+  var newArr = [];
+  for (let i = 0; i < arr.length; i++) { // Loop for reed items in aryy 
+    let greaterThan100 = (arr[i] > 100); // If number greater than 100 return true if not return false
+    let checkdecimalNumber = arr[i] % 1 == 0; // If number hase decimal (0.372) return true if not return false
+    newArr.push(`${(greaterThan100? 100 : checkdecimalNumber? arr[i]:arr[i].toFixed(2))}%`); // if number greater than 100 return 100, if hase decimal fixed to two decimal number if not return number
```
theseafaringturtle: I think it's better to break this up, chaining two ternary operators makes this harder to read for anyone else.

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -63,16 +58,16 @@ test(
 );
 test(
   "sortArray function works",
-  sortArray() === ["a", "c", "e", "f", "n", "z"]
+  JSON.stringify(sortArray()) === JSON.stringify(["a", "c", "e", "f", "n", "z"]) // you can not compare two arry only with "===" with JSON.stringify is possible
 );
-test("first5 function works", first5() === [1, 2, 3, 4, 5]);
+test("first5 function works", JSON.stringify(first5()) === JSON.stringify([1, 2, 3, 4, 5])); // you can not compare two arry only with "===" with JSON.stringify is possible
```
theseafaringturtle: You're right, we missed this. By the way, the usual way to do this is to compare them by checking every element in the array with a for loop.

PR #479, opened 2020-03-06
taskWeek2

Hamidreza-Fakharzadeh:


week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -38,9 +49,16 @@ Write a function that:
...
+  persentageArray = [];
+  for(i = 0; i < arr.lenght; i++) {
+     checkArray = arr[i] > 100;
+     checkDecimal = arr[i] % 1 == 0;
+     if(checkArray || checkDecimal) {
+       persentageArray.push(arr[i].toFixed(2))
+     }
+    }
+    return persentageArray;
+  }
```
frosas: This looks good but I'd pay attention to a few things:

- Don't forget to use `let` or `const`, otherwise the variables you use will be global, which is not a good idea.
- `arr.lenght` has a typo and will evaluate to `undefined`
- I don't think "numbers greater 100 must be replaced with 100" is implemented correctly
- What's the purpose of `checkDecimal`?
- Make sure to indent your code to make it more readable

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -15,7 +15,21 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
...
+  }
+  for(k = 0; k < ageSorted.length; k++) {
+    for(l = 0; l < ageSorted.length; l++) {
+      if(ageSorted[k] > ageSorted[l]) {
+        let temp = ageSorted[k];
+        ageSorted[k] = ageSorted[l];
+        ageSorted[l] = temp;
+      }
+    }
+    return ageSorted;
```
frosas: Isn't there a missing `}` here?

Don't forget to use `let` or `const` and indent your code here too!

PR #482, opened 2020-03-06
improvement 1

banirezaie:


week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -45,17 +45,91 @@ Very doubtful.
 
 // This should log "The ball has shaken!"
 // and return the answer.
-function shakeBall() {}
-
-// The answer should come from shaking the ball
+const number8 = Math.ceil(Math.random() * 20);
+function shakeBall() {
+  console.log(`The ball has shaken!`);
+  if (number8 === 1) {
```
illicitonion: Good job, this is a lot better!

A couple of things to think about to make this code even simpler:
1. It may be useful to add some comments in your big chain of `if` statements to show where each category starts/ends, i.e.
```javascript
// Very positive:
if (number8 === 1) {
  ...
...
}
// positive:
else if (number8 === 6) {
  ...
...
```
2.Instead if:
```javascript
if (number8 === 1 || number8 === 2 || number8 === 3 || number8 === 4 || number8 === 5)
```
is there a simpler expression you could use to do the same check? I think you can express this with _two_ comparisons, rather than 5.

week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -45,17 +45,91 @@ Very doubtful.
 
 // This should log "The ball has shaken!"
 // and return the answer.
-function shakeBall() {}
-
-// The answer should come from shaking the ball
+const number8 = Math.ceil(Math.random() * 20);
```
illicitonion: There's an interesting edge-case here :) `Math.random()` may return `0` (it's very unlikely, but it can happen), and `Math.ceil(0)` is `0`. That won't hit your `if` branches because they start at `1`.

You probably want to use `Math.floor(Math.random() * 20)` and start your checks at `0`, rather than using `ceil` and starting at `1` :)

PR #483, opened 2020-03-06
Alex week 2

AlexandruBudaca:


week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -24,22 +24,22 @@ function greaterThan10() {
...
+  let sortedLetters = letters;
   return sortedLetters;
 }
 
 function first5() {
   let numbers = [1, 2, 3, 4, 5, 6, 7, 8];
-  let sliced;
-
+  numbers.length = 5;
+  let sliced = numbers;
```
laky: Good job!

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -24,22 +24,22 @@ function greaterThan10() {
...
   return sliced;
 }
 
 function get3rdIndex(arr) {
-  let index = 3;
-  let element;
-
+  let index = arr[3];
+  let element = index;
```
laky: Almost perfect! Try to think about what the variables are trying to express. The `index` is generally used to index the array. In you implementation you now made `index` the same as the `element`. Could you think of a way to change these 2 lines so that the variables keep their meaning?

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -5,7 +5,18 @@ Write a function that:
...
+let namesFriends = [];
+
+function tidyUpString(namesFriends) {
+  for (i = 0; i < namesFriends.length; i++) {
+    namesFriends[i] = namesFriends[i]
+      .replace(/\//g, "")
+      .trim()
+      .toLowerCase();
+  }
+  return namesFriends;
```
laky: Good job, really nice!

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -15,7 +26,14 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
+function validate(num) {
+  let nums = num;
```
laky: Do you think this line is necessary? Could we go without it?

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -15,7 +26,14 @@ Complete the function to check if the variable `num` satisfies the following req
...
+  if (nums % 2 == 0 && nums <= 100) {
+    return true;
+  } else {
+    return false;
```
laky: Nice! One thing you can think about to improve the answer further would be whether all requirements for the `num` variable are expressed well in this `if` statement. Could you think of a way to make the `is a number` be part of the condition?

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -26,7 +44,9 @@ The function must:
 */
 
 function remove(arr, index) {
-  return; // complete this statement
+  let array = arr;
+  arr.splice(index, 1);
```
laky: Careful here. As you can see in https://www.w3schools.com/jsref/jsref_splice.asp, the splice method modifies the original array. This is something we do not want according to the requirements. To make this perfect, could you think of a way to make sure we do not modify the original array?

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -38,7 +58,19 @@ Write a function that:
 */
 
 function formatPercentage(arr) {
-  
+  let myArray = arr;
```
laky: Note that it might not be necessary to do this in this case. You can end up mixing `myArray` and `arr`, which did indeed happen. It might be cleaner to remove this line and avoid the potential confusion.

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -38,7 +58,19 @@ Write a function that:
...
+  let newArray = [];
```
laky: Just a small comment about naming that might help in the future. It is always good to use variable names that convey information about what they contain. Could you think of a way to change the name to express that this array contains the numbers rounded to 2 decimal points?

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -38,7 +58,19 @@ Write a function that:
...
+  let finalArray = [];
+  for (let i = 0; i < myArray.length; i++) {
+    newArray[i] = Math.round(arr[i] * 100) / 100;
+    if (newArray[i] > 100) {
+      newArray[i] = 100;
+    }
+    formatArray.push(newArray[i]);
+    finalArray[i] = `${formatArray[i].toString()}%`;
+  }
+  return finalArray;
```
laky: Nice, good job!

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -38,7 +58,19 @@ Write a function that:
...
+  let myArray = arr;
+  let newArray = [];
+  let formatArray = [];
+  let finalArray = [];
+  for (let i = 0; i < myArray.length; i++) {
+    newArray[i] = Math.round(arr[i] * 100) / 100;
+    if (newArray[i] > 100) {
+      newArray[i] = 100;
+    }
+    formatArray.push(newArray[i]);
```
laky: Do we need the `formatArray` here? Does it differ from the `newArray`? Try printing them both out and comparing to see.

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -7,12 +7,19 @@
   Answer the following questions:
 
   1. This program throws an error. Why? (If you can't find it, try executing it).
+  // b it.s not defined
```
laky: Indeed, nice job!

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -7,12 +7,19 @@
...
   2. Remove the line that throws the error.
+  // The line 38 throws the error because we don't have a defined variable b.
   3. What is printed to the console?
+  // The console prints first the variabels x and a, then the numbers: 4,9,6,13,8.
   4. How many times is "f1" called?
+  //Three times
   5. How many times is "f2" called?
+  //Two times
```
laky: Are you sure about these? Could you write down for which values `f1` gets called and for which values `f2` gets called? Note that `f2` is used before `f1` in the loop, so it might be a bit confusing.

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -7,12 +7,19 @@
   Answer the following questions:
 
   1. This program throws an error. Why? (If you can't find it, try executing it).
+  // b it.s not defined
   2. Remove the line that throws the error.
+  // The line 38 throws the error because we don't have a defined variable b.
```
laky: Nice!

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -7,12 +7,19 @@
...
   3. What is printed to the console?
+  // The console prints first the variabels x and a, then the numbers: 4,9,6,13,8.
```
laky: Nice!

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -13,45 +13,85 @@ sort method 😎)
...
+  let filterArray = [];
 
-function sortAges(arr) {
+  for (let i = 0; i < arr.length; i++) {
+    if (typeof arr[i] === "number") {
+      filterArray.push(arr[i]);
+    }
+  }
+  return filterArray;
+}
```
laky: Nice, good job!

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -13,45 +13,85 @@ sort method 😎)
...
+      if (sortedArray[i] > sortedArray[i + 1]) {
+        //swamp
+        let temp = sortedArray[i];
+        sortedArray[i] = sortedArray[i + 1];
+        sortedArray[i + 1] = temp;
+        swamp = true;
+      }
+    }
+  } while (swamp);
+  return sortedArray;
```
laky: Ok, this works, nice! 

Some questions you can think about: Do we need the `swamp` variable here? Is there a way we can do this in 2 `for` loops if we do not worry about using unnecessary computation?

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -13,45 +13,85 @@ sort method 😎)
 You don't have to worry about making this algorithm work fast! The idea is to get you to
 "think" like a computer and practice your knowledge of basic JavaScript.
 */
+function filterArrays(arr) {
```
laky: One thing I wanted to highlight and forgot to mention on the first pass. Great job on extracting this into a separate function! Things like this make the code easier to read and maintain and will be crucial later down the line! Keep it up! 🥇

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -4,7 +4,7 @@
 function mood() {
   let isHappy = true;
 
-  if (isHappy) {
+  if (isHappy === "I'm not happy") {
```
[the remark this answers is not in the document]
AlexandruBudaca: Should I define the variable isHappy = "I am happy"? Thank you

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -15,7 +15,7 @@ function greaterThan10() {
   let num = 10;
   let isBigEnough;
 
-  if (isBigEnough) {
+  if (isBigEnough => 10) {
```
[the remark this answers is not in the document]
AlexandruBudaca: If we put 9 it will return "num is not big enough". I will write in condition if(isBigEnough => num) Thank you

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -24,22 +24,22 @@ function greaterThan10() {
 
 function sortArray() {
   let letters = ["a", "n", "c", "e", "z", "f"];
-  let sortedLetters;
-
+  letters.sort();
+  letters.toString();
```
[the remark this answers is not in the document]
AlexandruBudaca: Ups. No because they are already strings.

week-2/Homework/mandatory/1-fix-functions.js
AlexandruBudaca: I will set the variable index equals with the arr and the element will be equals with index[3]

week-2/Homework/mandatory/2-function-creation.js
AlexandruBudaca: Yes, I don't need another variable

week-2/Homework/mandatory/2-function-creation.js
AlexandruBudaca: I can use the Number.isNan(num)

week-2/Homework/mandatory/2-function-creation.js
AlexandruBudaca: ok

week-2/Homework/mandatory/2-function-creation.js
AlexandruBudaca: toDecimalsArr

week-2/Homework/mandatory/4-sorting-algorithm.js
AlexandruBudaca: thank you

PR #484, opened 2020-03-07
for-the-task

yusuf963:


week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -1,116 +1,133 @@
...
+/*
+Complete the function to check if the variable `num` satisfies the following requirements:
+- is a number
+- is even
+- is less than or equal to 100
+Tip: use logical operators
+*/
+
+function validate(num) {
+if(typeof num === "number" && num % 2 === 0 && num <= 100){
```
ChrisOwen101: This would also work if you just wrote

`return typeof num === "number" && num % 2 === 0 && num <= 100`

ChrisOwen101:
In the future, make sure that you're working on your work on a branch instead of on master. At the top of this page it should say something like:

image

It should say `yusuf963:[YOUR_BRANCH]` where YOUR_BRANCH is the name of your branch. This makes it much easier for us :)

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -14,15 +14,59 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
 "think" like a computer and practice your knowledge of basic JavaScript.
 */
 
-function sortAges(arr) {
 
+function removeNoneNumbers(arr){
+  const onlyNumbers = [];
+  for(index = 0; index < arr.length; index++) {
+    const element = arr[index];
+     if(!isNaN(element)){     // isNaN returns a boolean, the function checks if the element passed in "is NOT a Number"
```
ChrisOwen101: Nice!

PR #487, opened 2020-03-07
Week 3 - Use let instead of var

frosas:


ismaelgt:
Hey Francesc. Thanks for submitting this PR. Please bear with us while we decide what to do with #475 as these exercises could be archived.

frosas:
Np!

PR #490, opened 2020-03-07
Ellie week2 extra

ellietms:
the test has a problem


week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -5,8 +5,12 @@ Write a function that:
 - removes any forward slashes (/) in the strings
 - makes the string all lowercase
 */
-function tidyUpString(strArr) {}
-
+function tidyUpString(arr) {
+  function stringarr(strArr) {
```
ChrisOwen101: It's normally better to put functions outside of each other so this code would be like this:

```
function tidyUpString(arr) {
  return arr.map(stringarr);
}

function stringarr(strArr) {
  return strArr.replace(/\//g,'').trim().toLowerCase();
}
```

ChrisOwen101: This can get really confusing if you start putting functions inside each other!

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -15,19 +19,23 @@ Complete the function to check if the variable `num` satisfies the following req
 Tip: use logical operators
 */
 
-function validate(num) {}
-
+function validate(num) {
+  if (typeof num === "number" && num % 2 === 0 && num<= 100 ){
```
ChrisOwen101: You can actually just write 

`return typeof num === "number" && num % 2 === 0 && num<= 100`

because that statement will `return` true or false anything. Give it a try!

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -37,9 +45,21 @@ Write a function that:
 - numbers greater 100 must be replaced with 100
 */
 
-function formatPercentage(arr) {
-  
-}
+function formatPercentage(arr){
+function mystr(strarr) {
```
ChrisOwen101: Move this function outside of the other one - similar to above :)

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -37,9 +45,21 @@ Write a function that:
...
+    if(strarr > 100){
+        return "100%";
+    }
+    if(Number.isInteger(strarr)){
```
ChrisOwen101: This is really close to being right but will only work for Integers (e.g. 1,2,3,4) but not for Floats (e.g 1.1, 31.21, 3133.131).

Do you know how you'd fix this?

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -37,9 +45,21 @@ Write a function that:
...
+      return `${strarr}%`
+    }
+    let Arr = strarr.toFixed(2);
+    let newArr = `${Arr}%`;
```
ChrisOwen101: This is really nice!

PR #491, opened 2020-03-08
Week2/cyf/bani

banirezaie:


week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -24,22 +22,21 @@ function greaterThan10() {
 
 function sortArray() {
   let letters = ["a", "n", "c", "e", "z", "f"];
-  let sortedLetters;
-
+  letters.sort();
+  let sortedLetters = letters;
```
ChrisOwen101: This could be shortened to 

`let sortedLetters = letters.sort()`

You dont need to re-assign the letters array to the sortedLetters

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -24,22 +22,21 @@ function greaterThan10() {
...
+  numbers.length = 5;
+  let sliced = numbers;
   return sliced;
 }
 
 function get3rdIndex(arr) {
-  let index = 3;
-  let element;
-
+  let index;
```
ChrisOwen101: This could be removed if it's not needed anymore

week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -24,22 +22,21 @@ function greaterThan10() {
...
+  let element = arr[3];
```
ChrisOwen101: If you didn't change the line about this could be written as

`let element = arr[index];`

Since you can pass in a variable to the array to get the value

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -15,43 +15,68 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
 */
 
 function sortAges(arr) {
```
ChrisOwen101: Seems like the tests still aren't passing here. Do you know why that might be?

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -15,43 +15,68 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
 */
 
 function sortAges(arr) {
-
+  let newArr = []; //to make an array with numbers
+  for (i = 0; i <= arr.length; i++) {
+    if (typeof arr === "number") {
+      newArr = arr.push[i];
+    }
+    for (j = 0; j <= newArr.length; j++) {
```
ChrisOwen101: This will break because newArr.length will return (for example), the number 7

However because arrays start at zero (e.g. arr[0], arr[1], arr[2]) it will break when it gets to 7.

You can fix this by just using 

`i < arr.length;`

PR #497, opened 2020-03-08
Ferhat week 3

ferhatu:


week-3/Homework/mandatory/1-oxygen-levels.js
```diff
@@ -8,33 +8,59 @@ To be safe, they need to land on the first unamed planet that has Oxygen levels
 
 Write a function that finds the oxygen level of the first safe planet - Oxygen between 19.5% and 23.5%
 */
+// function safeLevels(yourFunc) {
+//
+//   const abracaOutput = yourFunc(abracaArray);
 
-function safeLevels() {
+//   return abracaOutput;
```
ChrisOwen101: It's normally best if you remove commented code before you submit it

week-3/Homework/mandatory/7-password-validator.js
```diff
@@ -23,47 +23,54 @@ PasswordValidationResult=  [false, false, false, false, true]
 */
 
 function validatePasswords(passwords) {
+  //var myS = ["Kbn.i", "Nvsk?", "csdR!"]
+  var strongRegex = /^(?=.*\d)(?=.*[a-z])(?=.*[A-Z])\w{5,}$/;
 
+  // var access = strongRegex.test(myS);
+  var access = passwords.map(x => strongRegex.test(x));
```
ChrisOwen101: You can actually just return here like this
`return passwords.map(x => strongRegex.test(x));`

Give it a try!

week-3/Homework/mandatory/5-journey-planner.js
```diff
@@ -7,18 +7,24 @@
   NOTE: only the names should be returned, not the means of transport.
 */
 
-function journeyPlanner() {
-
+function journeyPlanner(select, vehicle) {
```
ChrisOwen101: So this is correct but it could be made even better by using the filter function that we learnt in class

```
select.filter(select=>select.includes(vehicle)).map(select[0])
```

PR #500, opened 2020-03-09
I want to start js-week-3

naderakhgari:


week-3/Homework/mandatory/1-oxygen-levels.js
```diff
@@ -8,11 +8,12 @@ To be safe, they need to land on the first unamed planet that has Oxygen levels
 
 Write a function that finds the oxygen level of the first safe planet - Oxygen between 19.5% and 23.5%
 */
-
-function safeLevels() {
-
+function safeLevels(oxygenLevels1) {
+  let safeLevel1 = oxygenLevels1.find(Element => Element > "19.5%" && Element < "23.5%");
```
40thieves: You're really close, but I don't think this is quite right.

The problem here is that you're using the greater than and less than operators with strings not numbers.

Try changing one of the values in the test to "100.5%". Does it do what you expected? How could you fix it?

week-3/Homework/mandatory/2-bush-berries.js
```diff
@@ -10,9 +10,11 @@
   Use the tests to confirm which message to return
 */
 
-function bushChecker() {
-
+function bushChecker(bushes) {
+  let bushBerryColours = bushes.every(Element => Element === "pink");
+  return bushBerryColours >0 ? "Bush is safe to eat from" : "Toxic! Leave bush alone!"
```
40thieves: You're really close but I don't think this is quite right.

What is the type of the value that `.every()` returns? (Hint: try looking at the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every)). Think carefully about how you would use this value in the ternary operator (like you have here) or an `if` statement.

week-3/Homework/mandatory/2-bush-berries.js
```diff
@@ -10,9 +10,11 @@
   Use the tests to confirm which message to return
 */
 
-function bushChecker() {
-
+function bushChecker(bushes) {
+  let bushBerryColours = bushes.every(Element => Element === "pink");
```
40thieves: Just a little thing, but remember that we always name variables with `camelCase`. So the first letter should always be lowercase, then every word after that starts with an uppercase letter.

week-3/Homework/mandatory/3-space-colonies.js
```diff
@@ -8,10 +8,20 @@
   NOTE: don't include any element that is not a "family".
 */
 
-function colonisers() {
-
+function colonisers(voyagers) {
+  let result = voyagers.filter(Element => Element.indexOf("family") >= 0 && Element.charAt(0) === "A");
```
40thieves: This is correct, great work!

However there is a new string method that is a bit nicer than `indexOf` for checking whether a string contains another string: [`.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes). How would this improve your code?

week-3/Homework/mandatory/4-eligible-students.js
```diff
@@ -4,12 +4,22 @@
...
-  - Returns an array containing onl the names of the who have attended AT LEAST 8 classes
+  - Returns an array containing only the names of the who have attended AT LEAST 8 classes
 */
 
-function eligibleStudents() {
-
+function eligibleStudents(attendances) {
+  let newArr = [];
+  newArr= attendances.filter(Element => Element[1] >= 8);
+  return newArr.map(Element => Element[0]);
```
40thieves: Really nice 💪

week-3/Homework/mandatory/5-journey-planner.js
```diff
@@ -7,16 +7,25 @@
   NOTE: only the names should be returned, not the means of transport.
 */
 
-function journeyPlanner() {
 
+function journeyPlanner(londonLocations, transportMode){
+  let bus =[];
+    bus = londonLocations.filter(Element => Element.includes(transportMode) );
```
40thieves: Is `bus` the best name for this variable? How could we improve it?

Otherwise, this is looking great!

week-3/Homework/mandatory/6-lane-names.js
```diff
@@ -4,10 +4,11 @@
   Write a function that will return all street names which contain 'Lane' in their name.
 */
 
-function getLanes() {
+function getLanes(streetNames) {
+    return streetNames.filter(Element => Element.indexOf("Lane") >=0);
```
40thieves: Like exercise 3, how could we improve this code (hint: try not using `.indexOf()`)

week-3/Homework/mandatory/7-password-validator.js
```diff
@@ -21,15 +21,88 @@ Expected Result:
 PasswordValidationResult=  [false, false, false, false, true]
 
 */
-
+// let alph = /[a-z]/;
+// let specialChar = /[".", "!", "#", "%", "$"]/
 function validatePasswords(passwords) {
+  let newArr = passwords.map(Element =>
+    (Element.indexOf("!") >= 0 ||
```
40thieves: This is an interesting solution! It is correct, but how could we improve it?

Hint: you could try researching how to use a *regex*: here's some [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) to get you started.

week-3/Homework/mandatory/7-password-validator.js
```diff
@@ -19,17 +19,24 @@ PreviousPassword = ["fhD8!yrjj", "ttkTu.wer3", "dvyyeyY!5", "qwbfj76%", "tytT372
...
 PasswordValidationResult=  [false, false, false, false, true]
-
 */
-
-function validatePasswords(passwords) {
-
+let lowercase = /[a-z]/;
+let uppercase = /[A-Z]/;
+let numbers = /[0-9]/;
+let nonAlphanumeric = /[.!$#%]/;
```
40thieves: Looks great! Well done 💪

week-3/Homework/mandatory/5-journey-planner.js
```diff
@@ -7,16 +7,38 @@
...
+  const londonLocations = [
+  ["Angel", "tube", "bus"],
+  ["London Bridge", "tube", "river boat"],
+  ["Tower bridge", "tube", "bus"],
+  ["Greenwich", "bus", "river boat"]
+]
+  let bus =[];
+  let tube =[];
+  let riverBoat =[];
+  if(transportMode == "bus"){
```
AlexeyBrambalov: ooooor you can do it in one line

function journeyPlanner(londonLocations, transport) {
return londonLocations.filter(location => location.includes(transport)).map(location => location = location[0]);
   
  
}

week-3/Homework/mandatory/2-bush-berries.js
```diff
@@ -10,8 +10,16 @@
   Use the tests to confirm which message to return
 */
 
-function bushChecker() {
-
+function bushChecker(bushes) {
+// clearImmediate;
+  let bushBerryColours = bushes.every(Element => Element === "pink");
+  let message;
+  if(bushBerryColours){
```
AlexeyBrambalov: oooor you can do it with one line 

 return  colours.every(colour => colour === "pink") ? "Bush is safe to eat from" : "Toxic! Leave bush alone!"

PR #499, opened 2020-03-08
Js-1/week_3/Min Ko

minwwgo:

PR #501, opened 2020-03-09
Week3 js 279 roxana

279Roxana:


week-3/Homework/mandatory/2-bush-berries.js
```diff
@@ -10,10 +10,31 @@
   Use the tests to confirm which message to return
 */
 
-function bushChecker() {
-
+function bushChecker(listOfBerryColours) {
+  for(let i=0; i < listOfBerryColours.length; i++){
```
40thieves: This works, but how could we do it with an array method? It looks like you were really close in the commented out code below!

week-3/Homework/mandatory/2-bush-berries.js
```diff
@@ -10,10 +10,31 @@
...
+    
+}
+  return "Bush is safe to eat from";
 }
 
+console.log(bushChecker(["pink", "pink", "pink", "neon", "pink", "transparent"]))
+console.log(bushChecker(["pink", "pink", "pink", "pink"]))
+// function bushChecker(arr) {
+
+//  if(arr.some(element => element !== "pink")){
```
40thieves: This is really close! Think carefully about what array method we want to use. The goal is to find out if **all** of the items in the array pass the check. (Hint: what's another way of saying "all of the items in the array"?)

week-3/Homework/mandatory/3-space-colonies.js
```diff
@@ -8,8 +8,10 @@
   NOTE: don't include any element that is not a "family".
 */
 
-function colonisers() {
+function colonisers(voyagers) {
+ let findFamily = voyagers.filter(voyager => voyager.includes("family") && voyager.charAt(0) === "A");
```
40thieves: Great work!

week-3/Homework/mandatory/4-eligible-students.js
```diff
@@ -4,13 +4,24 @@
...
     (see tests to confirm how this data will be structured)
-  - Returns an array containing onl the names of the who have attended AT LEAST 8 classes
+  - Returns an array containing only the names of the students who have attended AT LEAST 8 classes
 */
 
-function eligibleStudents() {
-
+function eligibleStudents(arr) {
+  
+  let students = arr.filter(x => x[1] >= 8).map(x => x.splice(0, 1).join(" "))
```
40thieves: Excellent work! I love the *chaining* here! 💪 

There's 1 little thing that you could improve. If you look at exercise 3, you'll see there's a different (and I think better) way of getting the first letter from a string. Could we use that here instead of the `.splice()` and `.join()`?

week-3/Homework/mandatory/5-journey-planner.js
```diff
@@ -7,16 +7,20 @@
   NOTE: only the names should be returned, not the means of transport.
 */
 
-function journeyPlanner() {
+function journeyPlanner(londonLocations, transport) {
+
+
+ return londonLocations.filter( x => x.includes(transport)).map(x => slice(0, 1).join(" "))
```
40thieves: I don't think everything here is quite right.

In the `.map()` function, you're calling a function named `slice` that I don't think exists, so I think this will cause an error. Instead I think you mean `x.slice(0, 1)` instead.

However, I think this is a bit overly complicated. Inside the map function we only need to get the first element from the array. The simplest way to do this would be to use array index syntax. How could we change our code here to do that?

One more tip: *chained* methods like you have here (`.filter(...).map(...)`) are a bit easier to read if you put them on newlines like this:

```js
return londonLocations
  .filter(x => x.includes(transport))
  .map(x => slice(0, 1).join(" "));
```

week-3/Homework/mandatory/6-lane-names.js
```diff
@@ -4,8 +4,9 @@
   Write a function that will return all street names which contain 'Lane' in their name.
 */
 
-function getLanes() {
-
+function getLanes(streetNames) {
+let street = streetNames.filter(x => x.includes("Lane"))
+return street
```
40thieves: Great 🎉

week-3/Homework/mandatory/7-password-validator.js
```diff
@@ -1,35 +1,38 @@
...
+*/ 
 
-*/
+ 
+function validatePasswords(arr) {
 
-function validatePasswords(passwords) {
 
-}
+    let validatePass = /(?=.{5,})(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[!#$%.])/
```
40thieves: Wow, this is a complicated regex! Do you feel like you understand what is happening here? If you do, could you explain how it works?

If not, I would recommend reading up on Regular Expressions (aka *regexes*). [This looks like a good place to start reading](https://codeburst.io/javascript-learn-regular-expressions-for-beginners-bb6107015d91). To see an example, Min Ko's homework has some really nice usage of regexes: https://github.com/CodeYourFuture/js-exercises/pull/499/files#diff-548cf5a2c7a92b75e67ed7cf8afdbbfbR26-R29 (scroll down to the bottom, it should be highlighted in yellow).

week-3/Homework/extra/cardvalidator.js
```diff
@@ -0,0 +1,28 @@
+
+
+function validateNumber(arr) {
+let validation = /(?=.{16})(?=.*[0-9]{2})(?=.([02468]$))/
```
40thieves: Similar to the password validator exercise, do you feel like you could explain this code?

week-3/Homework/mandatory/2-bush-berries.js
279Roxana: The commented code was supposed to be my final version but committed the task after trying out other methods. I have committed the  intended version.

week-3/Homework/mandatory/2-bush-berries.js
279Roxana: Another method would be "every".

week-3/Homework/mandatory/4-eligible-students.js
279Roxana: it would be more concise if I would have x[0] instead of using splice and join

week-3/Homework/mandatory/7-password-validator.js
279Roxana: I initially tested the code on https://regex101.com/ and thought it would be easy to break down by following each parenthesis.  Min's code looks more clear and it would be easier for people to read it and spot if there are mistakes along the way. Thank you for the resource

PR #505, opened 2020-03-10
Yohannes week 3

byohannes:


week-3/Homework/mandatory/2-bush-berries.js
```diff
@@ -1,34 +1,35 @@
...
-
-  Use the tests to confirm which message to return
-*/
-
-function bushChecker() {
-
+// this function checks if every string in the array is pink then bush is safe otherwise it is toxic.
+
+function bushChecker(bushBerries) {
+  let bushBerryColours = bushBerries.every(
```
ChrisOwen101: Its good to have a variable name that reflects what is inside it - maybe a better name for this would be "bushContainsPinkBerries"?

week-3/Homework/mandatory/7-password-validator.js
```diff
@@ -1,69 +1,68 @@
...
-
-Example 1:
-PreviousPassword = ["fhD8!yrjj", "ttkTu.wer3", "dvyyeyY!5", "qwbfj76%", "tytT3729."];
-
-Expected Result:
-PasswordValidationResult=  [false, false, false, false, true]
-
-*/
+// this function does validate any password within the password array
+function validatePasswordSimple(password) {
```
ChrisOwen101: This is really nice!

byohannes: Thanks, Chris, all this is due to the hard work of the mentors and course leads.

PR #508, opened 2020-03-12
Week 2 homework - Ola 

leealex88:
Please review my homework. Thank you.


ChrisOwen101:
Assigned to Lukas

week-2/mandatory/2-function-creation.js
```diff
@@ -0,0 +1,145 @@
...
+- removes any forward slashes (/) in the strings
+- makes the string all lowercase
+*/
+// ["/Daniel ", "irina ", " Gordon", "ashleigh "]
+function tidyUpString(strArr) {
+    for (let i = 0; i < strArr.length; i++) {
+        strArr[i] = strArr[i].trim().toLowerCase()
+        strArr[i] = strArr[i].replace("/", "")
+    }
+    return strArr
```
laky: Nice! Good job! 

A little extra thing to consider here that is asked later down the file as well. Does `strArr` that is passed as input change in this case?  In case it does, how would you change the function so that it does not change the input? There were not requirements about it this time, but it might be an interesting thing to consider.

week-2/mandatory/2-function-creation.js
```diff
@@ -0,0 +1,145 @@
...
+- is less than or equal to 100
+Tip: use logical operators
+*/
+
+function validate(num) {
+    if (typeof num == "number" && num % 2 == 0 && num <= 100) {
+        return true
+    } else {
+        return false
+    }
```
laky: 👍

week-2/mandatory/2-function-creation.js
```diff
@@ -0,0 +1,145 @@
...
+The function must:
+- NOT change the original array
+- return a new array with the item removed
+- remove the item at the specified index
+*/
+// let arr = [10, 293, 292, 176, 29]
+
+function remove(arr, index) {
+    arr.splice(index, 1)
+    return arr
```
laky: Nice, almost perfect! Have a look at the note in https://www.w3schools.com/jsref/jsref_splice.asp.

The tests didn't check the first requirement `The function must NOT change the original array`. Is it satisfied in this case? Try playing around a bit with this. Create a variable containing the array. Call your method on this variable and print it afterwards. Does the array stay unchanged? Could you think of a way to make sure the original array does not change?

week-2/mandatory/2-function-creation.js
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// console.log(remove(arr1, 3))
+// console.log(arr1)
```
laky: Exactly like here! These two logs print the same thing, however, remember, we did not want the original to change.

week-2/mandatory/2-function-creation.js
```diff
@@ -0,0 +1,145 @@
...
+Write a function that:
+- takes an array of numbers as input
+- returns an array of strings formatted as percentages (e.g. 10 => "10%")
+- the numbers must be rounded to 2 decimal places
+- numbers greater 100 must be replaced with 100
+*/
+const arr = [23, 18, 187.2, 0.372]
+
+function formatPercentage(arr) {
+    let procentArr;
```
laky: You've defined a new array here, but have not used it in the end. It seems like the intention was good to copy over the values rather than manipulate the original, you just forgot to use it in the end.

week-2/mandatory/2-function-creation.js
```diff
@@ -0,0 +1,145 @@
...
+    for (let i = 0; i < arr.length; i++) {
+        if (arr[i] > 100) {
+            arr[i] = 100
+        }
+        if (Math.floor(arr[i]) !== Math.ceil(arr[i])) {
+            arr[i] = arr[i].toFixed(2)
```
laky: A tiny bonus question: What would happen with a value like `19.2`. What do you think should happen? This isn't very clearly defined in the question, but it might be a fun thing to consider.

week-2/mandatory/2-function-creation.js
```diff
@@ -0,0 +1,145 @@
...
+        }
+        arr[i] = arr[i] + "%"
+
+    }
+    // console.log(arr)
+    return arr
```
laky: 👍

week-2/mandatory/3-playing-computer.js
```diff
@@ -1,42 +1,42 @@
 /*
   You have to predict the output of this program WITHOUT EXECUTING IT.
-
   In order to do this, try writing down the value that all variables take
   during each step of the program execution.
-
   Answer the following questions:
-
   1. This program throws an error. Why? (If you can't find it, try executing it).
+  - tha value b is not defind
```
laky: 👍

week-2/mandatory/3-playing-computer.js
```diff
@@ -1,42 +1,42 @@
...
   2. Remove the line that throws the error.
+  // console.log(b);
```
laky: 👍

week-2/mandatory/3-playing-computer.js
```diff
@@ -1,42 +1,42 @@
...
   3. What is printed to the console?
   4. How many times is "f1" called?
   5. How many times is "f2" called?
```
laky: Seems like you've missed out a couple of these. It might be worth to try and go through them still!

week-2/mandatory/3-playing-computer.js
```diff
@@ -1,42 +1,42 @@
...
   6. What value does the "a" parameter take in the first "f1" call?
+  - 6
```
laky: This might be a bit confusing as the outer `a` is passed into the function's `a` parameter. If we wanted to know what the inner `a`'s value was, what would the answer be?

week-2/mandatory/4-sorting-algorithm.js
```diff
@@ -0,0 +1,54 @@
...
+- takes an array of mixed data types as input
+- removes any non-number data types without using the built-in javascript filter method
+- returns an array of sorted ages in ascending order (without using the built-in javascript
+sort method 😎)
+You don't have to worry about making this algorithm work fast! The idea is to get you to
+"think" like a computer and practice your knowledge of basic JavaScript.
+*/
+
+function sortAges(arr) {
+
```
laky: Seems like you've missed this one. It would be great if you still gave it a shot! It's a good exercise to implement some of these algorithms.

PR #510, opened 2020-03-13
JS1/Week-3/Mo

mnashawati:


week-3/Homework/mandatory/7-password-validator.js
```diff
@@ -21,49 +21,59 @@ Expected Result:
...
-      validatePasswords(passwords2), [true, true, false, false, false]
-   )
- );
+  "validatePasswords function works - case 1",
+  arraysEqual(validatePasswords(passwords1), [false, false, true, false, false])
+);
+
+test(
+  "validatePasswords function works - case 2",
+  arraysEqual(validatePasswords(passwords2), [true, true, false, false, true])
```
frosas: I realised you shouldn't change that last `false` to `true`, notice the exercise says:

> Passwords must not be any previous password in the passwords array.

mnashawati: Ohhh I see! Thanks for pointing this out @frosas I will fix it.

week-3/.archive/C-array-every/exercise.js
```diff
@@ -5,9 +5,11 @@
 var students = ["Omar", "Austine", "Dany", "Swathi", "Lesley", "Rukmini"];
 var group = ["Austine", "Dany", "Swathi", "Daniel"];
 
-var groupIsOnlyStudents; // complete this statement
+var groupIsOnlyStudents = function(students, group) {
+  return group.every(element => students.includes(element));
```
ChrisOwen101: Element can be named something else so that this is more easy to understand - maybe calling it `student` would be be better?

week-3/Homework/mandatory/4-eligible-students.js
```diff
@@ -4,11 +4,14 @@
...
+  - Returns an array containing only the names of the students who have attended AT LEAST 8 classes
 */
 
-function eligibleStudents() {
-
+function eligibleStudents(arr) {
+  let eligibleNames = arr
+    .filter(element => element[1] >= 8)
+    .map(element => element[0]);
+  return eligibleNames;
```
ChrisOwen101: This can be simplified to just return the array. For example:

```
return arr
    .filter(element => element[1] >= 8)
    .map(element => element[0]);
```

week-3/Homework/mandatory/5-journey-planner.js
```diff
@@ -7,18 +7,23 @@
   NOTE: only the names should be returned, not the means of transport.
 */
 
-function journeyPlanner() {
-
+function journeyPlanner(arr, str) {
+  let stationNames = arr
+    .filter(element => element.includes(str))
+    .map(element => element[0]);
+  return stationNames;
```
ChrisOwen101: Same as above

mnashawati: Thank you @ChrisOwen101 noted and refactored.

week-3/.archive/C-array-every/exercise.js
mnashawati: Yes it makes sense.

week-3/Homework/extra/1-card-validator.js
```diff
@@ -0,0 +1,82 @@
...
+
+function validateCreditCard(cardNum) {
+  // Changing the number to an array of single digit strings.
+  let digitsAsStringsArr = String(cardNum).split("");
+  let len = digitsAsStringsArr.length;
+  let isNumber = /[0-9]/;
+  // If any of the following conditions is true the number will be invalid.
+  if (
+    len !== 16 ||
+    digitsAsStringsArr.some(element => !isNumber.test(parseFloat(element))) ||
```
illicitonion: Regular expressions are things we test strings against - here you're converting the string to a number, and then testing it against the regular expression. You don't need this `parseFloat `here - that said, javascript will convert that float back to a string to do the test for you, but it's extra work that could be avoided :)

week-3/Homework/extra/1-card-validator.js
```diff
@@ -0,0 +1,82 @@
...
+    digitsAsStringsArr.every(element => element === digitsAsStringsArr[0]) ||
```
illicitonion: Very nice!

week-3/Homework/extra/1-card-validator.js
```diff
@@ -0,0 +1,82 @@
...
+    parseFloat(digitsAsStringsArr[len - 1]) % 2 !== 0 ||
+    digitsAsStringsArr
+      .map(digit => parseFloat(digit))
+      .reduce((a, b) => a + b, 0) < 16
+  ) {
+    return `The credit card number: ${cardNum} is invalid.`;
```
illicitonion: One thing to think about is that this string, and the string below on line 54, are almost exactly the same. Can you think of a way to avoid that repetition, so that e.g. if you wanted to put the card number in quotes in the string, you'd only need to add the quotes in one place not two?

mnashawati: Maybe assign this bit `The credit card number: ${cardNum} is` to a variable and use it in return instead?

mnashawati:
> I only looked at the credit card validator, but it's looking really good :) I left a few comments for really slight improvements, but you should feel very happy with what you've produced!

Thanks a lot for looking @illicitonion I really appreciate it! :)

week-3/Homework/mandatory/4-eligible-students.js
mnashawati: Done.

week-3/Homework/extra/1-card-validator.js
mnashawati: Ohh, that is very good to know. I deleted the unnecessary parseFloat.

PR #512, opened 2020-03-13
Week 2 -- Meisam

Erwin6997:
Ok, I got this as well.

### DIFF
--- a/week-2/Homework/mandatory/4-sorting-algorithm.js
+++ b/week-2/Homework/mandatory/4-sorting-algorithm.js
@@ -17,18 +17,28 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
 function sortAges(arr) {
  let maxarray = [];
  for ( let i = 0 ; i <= arr.length ; i ++ ){
-   if (typeof arr[i] != `number` || typeof arr[i] !== `boolean`){
+   if (typeof arr[i] !== `number` ||
+       typeof arr[i] === `boolean`||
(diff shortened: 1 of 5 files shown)


week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -15,12 +15,30 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
 */
 
 function sortAges(arr) {
+ let maxarray = [];
+ for ( let i = 0 ; i <= arr.length ; ++i ){
+   if (typeof arr[i] == `number`){
+     maxarray.push(arr[i]);
+   }
+ }
+ sortAge(maxarray);
```
laky: The name of this function is very similar to the one above and could be easily confused. You can try renaming to avoid confusion. In the previous part you already removed any non-numeric values.

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -15,12 +15,30 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
 */
 
 function sortAges(arr) {
+ let maxarray = [];
+ for ( let i = 0 ; i <= arr.length ; ++i ){
+   if (typeof arr[i] == `number`){
+     maxarray.push(arr[i]);
+   }
```
laky: Nice, this is exactly what was needed!

PR #516, opened 2020-03-14
Alex week3

AlexandruBudaca:


week-3/Homework/mandatory/7-password-validator.js
```diff
@@ -1,69 +1,78 @@
...
- test(
-   "validatePasswords function works - case 2",
-   arraysEqual(
-      validatePasswords(passwords2), [true, true, false, false, false]
-   )
- );
\ No newline at end of file
+test(
+  "validatePasswords function works - case 2",
+  arraysEqual(validatePasswords(passwords2), [true, true, false, false, true])
```
frosas: Hey Alex, the exercise looks good but you shouldn't change the test as it works as expected. From the exercise text:

> Passwords must not be any previous password in the passwords array.

AlexandruBudaca: Hy Francesc,
I.m gone remake it. Thank you!

week-3/Homework/mandatory/1-oxygen-levels.js
```diff
@@ -9,32 +9,33 @@ To be safe, they need to land on the first unamed planet that has Oxygen levels
 Write a function that finds the oxygen level of the first safe planet - Oxygen between 19.5% and 23.5%
 */
 
-function safeLevels() {
-
+function safeLevels(planet) {
+  let oxyPlanet = planet.find(x => x > "19.5%" && x < "23.5%");
```
40thieves: You're really close, but I don't think this is quite right.

The problem here is that you're using the greater than and less than operators with strings not numbers.

Try changing one of the values in the test to "100.5%". Does it do what you expected? How could you fix it?

week-3/Homework/mandatory/2-bush-berries.js
```diff
@@ -10,14 +10,18 @@
   Use the tests to confirm which message to return
 */
 
-function bushChecker() {
-
+function bushChecker(pinkColor) {
+  let bePink = pinkColor.every(item => item == "pink");
+  if (bePink == true) {
```
40thieves: This is absolutely correct (well done!), but the `== true` is actually unnecessary. To see why, let's play computer:

1. The computer runs the `bushes.every(...)` code
2. This evaluates to the boolean `true`
3. This value is then assigned to the `bePink` variable
4. The computer reads the `if` statement and sees that the comparison is `bePink == true`
5. It evaluates `bePink == true`
6. The `bePink` variable is assigned to `true` so the expression becomes `true == true`
7. This evaluates to a single `true` value (because `true` is equal to `true`!)
8. The `if` statement runs the body (returning `"Bush is safe to eat from"`) because the value is `true`

If we changed the code to just `if (bePink)` we could skip steps 5 - 7! This is very common when writing JS.

---

Additionally, in a situation like this where we only have 1 condition in the `if` statement, we usually don't separate it out into a separate variable (`bePink` in this case). Instead we just *inline* the comparison into the `if` statement. Do you think you could change that?

week-3/Homework/mandatory/3-space-colonies.js
```diff
@@ -8,8 +8,12 @@
   NOTE: don't include any element that is not a "family".
 */
 
-function colonisers() {
-
+function colonisers(voyagers) {
+  let colonisersName = voyagers
+    .filter(n => n[0] == "A")
+    .filter(name => name.includes("family"));
```
40thieves: Great 👍

week-3/Homework/mandatory/4-eligible-students.js
```diff
@@ -7,8 +7,12 @@
   - Returns an array containing onl the names of the who have attended AT LEAST 8 classes
 */
 
-function eligibleStudents() {
-
+function eligibleStudents(students) {
+  let studentsName = students.filter(([a, b]) => b >= 8);
+  let filterStudentsName = studentsName.map(x => x.slice(0, 1));
```
40thieves: Looking good! Since you're only getting 1 element from the array here, I think it would be simpler to just use array index syntax (`[...]`)?

week-3/Homework/mandatory/4-eligible-students.js
```diff
@@ -7,8 +7,12 @@
...
+  let finalNames = [].concat(...filterStudentsName);
```
40thieves: This is interesting 🙂Could you explain why this line is necessary?

week-3/Homework/mandatory/5-journey-planner.js
```diff
@@ -7,18 +7,22 @@
   NOTE: only the names should be returned, not the means of transport.
 */
 
-function journeyPlanner() {
+function journeyPlanner(location, mode) {
+  let allLocations = location.filter(x => x.includes(mode));
+  let yourLocation = mo.map(x => x[0]);
```
40thieves: Does this code run correctly? How could you fix the error here?

week-3/Homework/mandatory/6-lane-names.js
```diff
@@ -4,43 +4,46 @@
   Write a function that will return all street names which contain 'Lane' in their name.
 */
 
-function getLanes() {
-
+function getLanes(londonStreets) {
+  let streets = londonStreets.filter(x => x.includes("Lane"));
```
40thieves: Correct 🚀

week-3/Homework/mandatory/7-password-validator.js
```diff
@@ -1,69 +1,83 @@
...
 Expected Result:
 PasswordValidationResult=  [false, false, false, false, true]
-
 */
 
-function validatePasswords(passwords) {
+let allUpper = /[A-Z]/;
+let allLower = /[a-z]/;
+let allNumbers = /[0-9]/;
+let allSymbols = /[!#$%.]/;
```
40thieves: Nice work! Can you explain what these lines mean? What are you using here?

week-3/Homework/mandatory/1-oxygen-levels.js
40thieves: Note that this appears to work because `>` does actually work with strings. It just checks **alphabetically** instead of numerically, which is why 100.5% will break it.

40thieves:
Oh sorry @frosas, I didn't see that you'd reviewed this too 🙂

week-3/Homework/mandatory/1-oxygen-levels.js
AlexandruBudaca: Hi, yes

> Well done @AlexandruBudaca! 🎆 🌻 🏅
> 
> Have you tried the extra exercise?

I think about it, but I didn't try :( I will

AlexandruBudaca: Yes we remade the exercise in class. we used parseFloat(), but I didn't remake the exercise.

week-3/Homework/mandatory/2-bush-berries.js
AlexandruBudaca: Yes I can write like
bePink ? "Bush is safe to eat from" : "Toxic! Leave bush alone!"

week-3/Homework/mandatory/4-eligible-students.js
AlexandruBudaca: because I have an 2d array and I nedeed to make 1d  so I concatinated an empty array with my 2d array

week-3/Homework/mandatory/5-journey-planner.js
AlexandruBudaca: Oh, sorry. I write the wrong variable on line 11.
 It will be:
let yourLocation = allLocations.map(x => x[o]);

week-3/Homework/mandatory/7-password-validator.js
AlexandruBudaca: Yes, it's the regular expression,  they test/(match )  if the passwords contains the characters that is in the expression. Thank you

PR #522, opened 2020-03-18
week3-mandatory

ellietms:
mandatory week3 hw


week-3/Homework/mandatory/2-bush-berries.js
```diff
@@ -9,10 +8,21 @@
 
   Use the tests to confirm which message to return
 */
-
-function bushChecker() {
-
+function checkColor(colour){
+    if(colour === "pink"){
+      return colour;
```
ChrisOwen101: This is really close!

You should actually return true or false from this function so that it the function below works in all cases.

Technically this is working because you compare `result` to what is returned below and a none empty string will return null.

week-3/Homework/mandatory/3-space-colonies.js
```diff
@@ -7,11 +7,23 @@
 
   NOTE: don't include any element that is not a "family".
 */
+function findNamesWithA(names){
+  if(names.charCodeAt(0) === 65){
+    return names;
```
ChrisOwen101: Again, this should return `true` or `false`  to make this work in all cases

week-3/Homework/mandatory/3-space-colonies.js
```diff
@@ -7,11 +7,23 @@
 
   NOTE: don't include any element that is not a "family".
 */
+function findNamesWithA(names){
+  if(names.charCodeAt(0) === 65){
```
ChrisOwen101: 65 does equal the letter `A` but you can just write "A" here to make this easier to read.

65 equals the letter A because that is what it is in the `ascii table` you can read more about that here.
http://www.asciitable.com/

week-3/Homework/mandatory/3-space-colonies.js
```diff
@@ -7,11 +7,23 @@
...
+  }
+}  
 
-function colonisers() {
-
+function colonisers(arrayOfNames) {
+  let filterNames = arrayOfNames.filter(findNamesWithA);
+  console.log(filterNames);
+  let results =[];
+  for(let i=0;i<filterNames.length;i++){
```
ChrisOwen101: It would be even better if you used a `forEach` here

week-3/Homework/mandatory/4-eligible-students.js
```diff
@@ -7,8 +7,15 @@
   - Returns an array containing only the names of the who have attended AT LEAST 8 classes
 */
 
-function eligibleStudents() {
-
+function eligibleStudents(studentsname) {
+  let eligibleOnes = [];
+  for(let i = 0 ;i < studentsname.length; i++){
```
ChrisOwen101: It'd be better if you used a `filter` here so that you use what you learnt in the lesson :)

week-3/Homework/mandatory/5-journey-planner.js
```diff
@@ -7,8 +7,14 @@
   NOTE: only the names should be returned, not the means of transport.
 */
 
-function journeyPlanner() {
-
+function journeyPlanner(londonLocations,vehicle) {
+  let locations=[];
+  for(let i=0;i<londonLocations.length;i++){
+    if(londonLocations[i].includes(vehicle)){
```
ChrisOwen101: Again it would be better to use a filter here :)

PR #525, opened 2020-03-19
JS1- week3-Alisina

Alisina92:


week-3/Homework/mandatory/1-oxygen-levels.js
```diff
@@ -9,10 +9,21 @@ To be safe, they need to land on the first unamed planet that has Oxygen levels
...
-function safeLevels() {
-
+function safeLevels(oxygenLevels) {
+  /*loops through the array using map return converted float array */
+  const floatConversion = oxygenLevels.map(element=>parseFloat(element));//used arrow fuction as map argument.
+  
+  /*finds oxygen level of the floated array it is important to note '.find returns first element that meets the requirement'*/
+  const findOxyLevel = floatConversion.find(element => element > 19.5 && element < 23.5); //used arrow fuction as find argument.
+   
+  /*Converts the returned required oxygen to string*/
```
illicitonion: This works well, but turning it into a string is slightly more complicated than it could be :) There are a couple of things here:
1. String interpolation will implicitly convert things to strings, so you don't need to call `toString()` (even though you _can_ - it's not broken to, just more code)
2. We tend to _either_ use interpolation _or_ concatenation, but not both.
So you could either write this:
Using interpolation: 
```javascript
const percentify = `${findOxyLevel}%`;
```
or using concatenation:
```javascript
const percentify = findOxyLevel + '%';
```

Like I said - your code works, and is correct, these are just some suggestions for how to make it a little simpler :)

week-3/Homework/mandatory/2-bush-berries.js
```diff
@@ -9,10 +9,22 @@
 
   Use the tests to confirm which message to return
 */
+ 
+
+function bushChecker(bushBerryColours) {
+  if (bushBerryColours.includes('neon', 'transparent')){
```
illicitonion: This works for our test cases, but isn't quite solving the problem as it was described in the comments :) Rather than looking to see whether the array contains particular bad values, can you try re-doing this looking to see whether _everything_ in the array is pink?

week-3/Homework/mandatory/3-space-colonies.js
```diff
@@ -8,9 +8,12 @@
   NOTE: don't include any element that is not a "family".
 */
 
-function colonisers() {
-
-}
+function colonisers(voyagers) {
```
illicitonion: Like with the berries, this code is much more geared towards the specific test-cases we gave (which are meant to be examples) rather than the real problem to solve (described in the comments at the start). The specification of the problem talks about strings containing "family", rather than strings not being Asimov. Can you give this another go, trying to solve the problem specified in the comments, rather than the specific test-cases?

week-3/Homework/mandatory/4-eligible-students.js
```diff
@@ -4,13 +4,19 @@
...
+  - Returns an array containing only the names of the who have attended AT LEAST 8 classes
 */
 
-function eligibleStudents() {
-
-}
-
+function eligibleStudents(attendances) {
+     const filteredArr =[];
+     attendances.forEach(element=>{
```
illicitonion: This works, well done :) To practice using the `filter` function, could you try also solving this using `filter`, rather than `forEach`?

week-3/Homework/mandatory/5-journey-planner.js
```diff
@@ -7,16 +7,30 @@
   NOTE: only the names should be returned, not the means of transport.
 */
 
-function journeyPlanner() {
-
+function journeyPlanner(londonLocations,transportType) {
```
illicitonion: Like with the last two, this looks like it's really over-fitting to the test cases, rather than the intent of the problem - can you try again? We can help you out on slack if you get stuck :)

PR #527, opened 2020-03-20
W3 02

cockochan:


cockochan:
I hope its ok

week-3/Homework/mandatory/1-oxygen-levels.js
```diff
@@ -9,32 +9,38 @@ To be safe, they need to land on the first unamed planet that has Oxygen levels
 Write a function that finds the oxygen level of the first safe planet - Oxygen between 19.5% and 23.5%
 */
 
-function safeLevels() {
-
+function safeLevels(arr) {
+  let found = arr.filter(element => element > "19.5%" && element < "23.5%");
+
+  console.log(found);
+  if (found[0] === undefined) {
```
illicitonion: This is an interesting check to be doing - can you explain what it's achieving / why it's here?

week-3/Homework/mandatory/3-space-colonies.js
```diff
@@ -8,8 +8,10 @@
   NOTE: don't include any element that is not a "family".
 */
 
-function colonisers() {
-
+function colonisers(arr) {
+  let famArr = arr.filter(x => x.includes("family"));
+  let colonisers = famArr.filter(i => i[0] === "A");
```
illicitonion: This works well, may mean that we're going to look through the array twice rather than once; can you think of how to rewrite this code to only be one filter rather than two?

week-3/Homework/mandatory/5-journey-planner.js
```diff
@@ -7,18 +7,26 @@
   NOTE: only the names should be returned, not the means of transport.
 */
 
-function journeyPlanner() {
-
+function journeyPlanner(arr, transport) {
```
illicitonion: This logic is doing the right thing (yay!) but the tests are failing - can you work out why that is?

Also, for practice, can you think of a way to write this using filter and map, rather than two for loops? (This code totally works, but as we were learning about filter and map, it would be good to practice using them too :))

week-3/Homework/mandatory/3-space-colonies.js
cockochan: I found this article and decided to try to experiment and compare "as is" with % and " and received positive results. later on the call teachers found some exceptions and why this may not work in certain cases, but i was very busy with exercise 7 and did not have time to come back and do a better version...

PR #531, opened 2020-03-21
Js week3/sabia

sabia-m:
unable to pass test 2 of ex 7


week-3/Homework/mandatory/6-lane-names.js
```diff
@@ -4,10 +4,11 @@
   Write a function that will return all street names which contain 'Lane' in their name.
 */
 
-function getLanes() {
-
+function getLanes(arr) {
+   let newArr = arr.filter(e => e.includes ("Lane"))
```
illicitonion: This works, but is slightly more code than it needs to be. You don't really need the `newArr` variable here - it isn't really providing any value, and so it may be a bit clearer to instead write:
`return arr.filter(e => e.includes("Lane"));`

week-3/Homework/mandatory/7-password-validator.js
```diff
@@ -21,10 +21,54 @@ Expected Result:
...
+      checkSymbol.test(testPassword) &&
+      testPassword.length >= 5
+      ){
+          return true
+        }  else {
+          return false
+        }
+  } 
+  function validatePasswords(arr){
+    let newArr = arr.map(testPassword)
```
illicitonion: Like with the previous exercise, this works, but could just be `return arr.map(testPassword);` :)

week-3/Homework/mandatory/6-lane-names.js
Alisina92: Sabia please write element instead of e would be more clear

PR #535, opened 2020-03-21
Week 4 homework updates.

BexB1:


week-4/Homework/mandatory/3-freecodecamp.md
```diff
@@ -0,0 +1,31 @@
+## FreeCodeCamp Lessons
```
makanti: As discussed let's move those to extras directory, as the syllabus doesn't include a lot of what the tutorials will be talking about.

PR #538, opened 2020-03-21
Kyaw js-ex-week3 done

kiowd:


week-3/Homework/extra/1-card-validator.js
```diff
@@ -0,0 +1,26 @@
+function validateCreditCard(cardno){
```
illicitonion: This is looking great, but we're missing one of the checks - that all of the digits aren't all the same. Could you add a check for that, too?

week-3/Homework/mandatory/2-bush-berries.js
```diff
@@ -10,14 +10,18 @@
   Use the tests to confirm which message to return
 */
 
-function bushChecker() {
-
+function bushChecker(arr) {
+  let pink = arr.every(name => name.includes("pink"));
```
illicitonion: This looks mostly right, but `includes` is an interesting check here... What would happen if our array was:

`["pink", "not pink", "pink"]`

?

week-3/Homework/mandatory/3-space-colonies.js
```diff
@@ -8,7 +8,9 @@
   NOTE: don't include any element that is not a "family".
 */
 
-function colonisers() {
+function colonisers(arr) {
+  let stay = arr.filter(col => col.charAt(0) === "A" && col.includes("family"));
```
illicitonion: This works, but is slightly more code than we need to write to solve the problem. We're not really using the `stay` variable here, so we may not want to declare it at all, and instead just write:
`return arr.filter(col => col.charAt(0) === "A" && col.includes("family"));`
because it's a little shorter. Your code is doing the right thing, so the next level up is to think about how we can make our code as simple and easy to read as possible :)

You've done the same thing in most of the exercises, which is fine, but probably worth thinking about in the future :)

week-3/Homework/mandatory/7-password-validator.js
```diff
@@ -22,8 +22,20 @@ PasswordValidationResult=  [false, false, false, false, true]
...
 */
 
-function validatePasswords(passwords) {
+// function validPassword(password, index, arr) {
+//   let pwd = arr.map(password => password.includes(password.match(/^(?=.*[\d])(?=.*[A-Z])(?=.*[a-z])(?=.*[!@#$%^&*])[\w!@#$%^&*]{5,}$/)));
+//   let prePwd = arr.indexOf(password) === index;
+//   return ( pwd && prePwd);
+// }
+function validPassword(password, index, arr) {
+  let pwd = /^(?=.*[\d])(?=.*[A-Z])(?=.*[a-z])(?=.*[!@#$%^&*])[\w!@#$%^&*]{5,}$/;
```
illicitonion: This is a pretty complicated looking regular expression - can you talk through what it does? What does each part of it do?

It can often be easier to use a few small regular expressions, rather than one big one - that can make it easier to see what's going on to people reading the code. Could you think of a few different regular expressions, the `test` of which you could `&&` together to make sure they're all matched by the string?

PR #539, opened 2020-03-22
Js1/week3/bani

banirezaie:


week-3/Homework/mandatory/7-password-validator.js
```diff
@@ -21,45 +21,50 @@ Expected Result:
...
 */
-
+function test(passwords) {
+  let strongRegex = /^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[.!@#$%^&*])(?=.{5,})/;
+  return strongRegex.test(passwords);
+}
 function validatePasswords(passwords) {
-
+  let array = passwords.map(test);
+  let array2 = array.reverse(); //just to pass the test!!!
```
illicitonion: Oh no! This isn't quite right! There are a few things here:

1. The tests were wrong!! The test arrays should be:
```
const passwords1 = ["Se%5", "TktE.TJTU", "384#HsHF", "dvyyeyy!5", "tryT3729"];
const passwords2 = ["StUFf27%", "Pl3nty!", "Jai33", "shajsaUA**&&", "Pl3nty!"];
```
2. The regular expression isn't quite right, and it's pretty hard to debug why because it's big and complicated. If you break it down into a few small really simple regular expressions (one for testing for upper-case letters, one for testing for numbers, and so on), it should be a lot easier to debug and understand.
3. We're not checking to see whether the password appeared previously in the array.

Can you give this one another go? You can always reach out on Slack if you get stuck :)

PR #555, opened 2020-03-24
JS2-core-Week1

tayfuntayfun:


week-4/Homework/mandatory/3-water-bottle.js
```diff
@@ -11,13 +11,13 @@ We made a start on this for you here:
...
   volume: 0,
   fill: function() {
-    // calling this function should make you bottles volume = 100;
+    this.volume = 100; // calling this function should make you bottles volume = 100;
   },
   drink: function() {
-    // calling this function should decrease your bottles volume by 10;
+    this.volume -= 10;// calling this function should decrease your bottles volume by 10;
   },
   empty: function() {
```
ferhatu: What would be if you use bottle.volume ? what is the difference with this.volume?

week-4/Homework/mandatory/2-writers.js
```diff
@@ -39,6 +39,8 @@ let writers = [
   }
 ];
 
+message = writers.filter(x => x.alive).forEach(x => console.log(`Hi, my  name is ${x.firstName} ${x.lastName}. I am ${x.age} years old, and work as a ${x.occupation}`))
```
ferhatu: You can use map instead of for.each. map is more chain friendly.

week-4/Homework/mandatory/3-water-bottle.js
tayfuntayfun: YEs, map could do as well

week-4/Homework/mandatory/2-writers.js
tayfuntayfun: Yes, map could do as well, even better

tayfuntayfun: replacing with map

PR #560, opened 2020-03-25
JS2-Week 1/(Week-4) - Mo

mnashawati:


week-4/Homework/mandatory/2-writers.js
```diff
@@ -42,3 +42,11 @@ let writers = [
 /*
 If you want an extra challenge, only `console.log()` the writers that are alive.
 */
+
+writers.forEach(writer => {
+  if (writer.alive) {
```
AlexeyBrambalov: you can use the filter array method instead of if

dahfool: Yes, agree with @AlexeyBrambalov try `filter`

week-4/Homework/mandatory/3-water-bottle.js
```diff
@@ -12,12 +12,15 @@ let bottle = {
...
     // calling this function should make you bottles volume = 100;
+    this.volume = 100;
   },
   drink: function() {
     // calling this function should decrease your bottles volume by 10;
+    this.volume -= 10;
   },
   empty: function() {
     // this function should return true if your bottles volume = 0
+    this.volume === 0 ? true : false;
```
dahfool: use of ternary operate 👍

week-4/InClass/D-methods/exercise-5.js
```diff
@@ -20,16 +20,25 @@ let coffeeMachine = {
...
-      return `Sorry you don't have enough money for a ${coffee}`;
-    } else {
-      return `Not available!`;
-    }
+    let hasPaidEnough = this.insertedAmount >= this.prices[coffee];
+    return !this.prices.hasOwnProperty(coffee)
+      ? `Not available!`
+      : hasPaidEnough
+      ? `Please take your ${coffee}`
+      : `Sorry you don't have enough money for a ${coffee}`;
```
279Roxana: Nice use of the nested ternary operator!

week-4/Homework/mandatory/3-water-bottle.js
vivekjain202: @dahfool  i certainly disagree here, as the Equality operator always return Boolean value so there's not need to explicit `true` or `false` in first place

PR #562, opened 2020-03-26
W4 3

cockochan:
I create this pull although i  think i need to finish etc. but i cant see some of my commits on my other 2 computers so i want this to be backed up here


week-4/Homework/mandatory/2-writers.js
```diff
@@ -41,4 +41,13 @@ let writers = [
...
+// */
+//writers.forEach(element => console.log(element.firstName));
+
+
+
+  writers.forEach(function (writer) {
+    if (writer.alive === true) {
+      console.log(`Hi, my name is ${writer.firstName} ${writer.lastName}. I am ${writer.age} years old, and work as a ${writer.occupation}.`);
+    }
+  });
```
byohannes: this part only solves the extra challenge only but you have missed solving the main question which is on the top:
1-loop through the array of the objects and console out Hi, my name is ${writer.firstName} ${writer.lastName}. I am ${writer.age} years old, and work as a ${writer.occupation}.`
2-extra challenge, only `console.log()` the writers that are alive.

cockochan: I suppose it is because you only see the lines that have bin changed in current commit. To see entire exercise solution you will probably need to open the repo and see raw code or something like that. I suppose i solved the main task in previous commits and added "alive" section later as an extra commit

makanti: I can't see that part either. I have checked out the repo.

PR #571, opened 2020-03-28
Kyaw's  js-core2-w1

kiowd:


week-4/Homework/mandatory/2-writers.js
```diff
@@ -38,7 +38,8 @@ let writers = [
     alive: true
   }
 ];
-
+let allWriters = writers.forEach(all => console.log(`Hi, my name is ${all.firstName} ${all.lastName}. I am ${all.age} years old, and work as a ${all.occupation}.`))
 /*
 If you want an extra challenge, only `console.log()` the writers that are alive.
 */
+let writersAlive = writers.filter(person => person.alive === true).map(person => console.log(`${person.firstName}  ${person.lastName}`))
```
SCavus: Hi Kyaw,

Nice work. You could do **(person => person.alive === true)**like this (**person => person.alive**) both are the same

week-4/Homework/mandatory/4-groceries.js
```diff
@@ -6,7 +6,12 @@
...
-  item1: "",
-  item2: "",
-  item3: ""
+  item1: "Potatoes",
+  item2: "Orange Juice",
+  item3: "Rice"
 };
+
+for(let glist in groceryList){
+  groceriesToBuy.push(groceryList[glist]);
```
279Roxana: Good usage of for in loop :) . Do you think it would have worked with a method? Which one?

PR #579, opened 2020-03-29
LON_Babak class6-JS-week4

babakjahangiri:
LON_Babak class6-JS-week4


week-4/Homework/mandatory/2-writers.js
```diff
@@ -39,6 +39,20 @@ let writers = [
...
+writers.forEach(function (item){
+console.log(`Hi, my name is ${item.firstName} ${item.lastName}. I am ${item.age} years old, and work as a ${item.occupation}.`)
+});
+
+
 /*
 If you want an extra challenge, only `console.log()` the writers that are alive.
 */
+console.log("---- Alive Writers ----")
+writers.forEach(function (item){
```
ChrisOwen101: This works but have you thought about using a Filter here instead?

week-4/Homework/mandatory/3-water-bottle.js
```diff
@@ -12,12 +12,15 @@ let bottle = {
...
     // calling this function should make you bottles volume = 100;
+    return (this.volume = 100);
   },
   drink: function() {
     // calling this function should decrease your bottles volume by 10;
+    return (this.volume -= 10);
   },
   empty: function() {
     // this function should return true if your bottles volume = 0
+    if (this.volume == 0) return true;
```
ChrisOwen101: This will work for some of the cases but all of them. What if the volume is not 0?

PR #590, opened 2020-03-29
Js2-week4-exercises-tm

tanianm:


week-1/Homework/mandatory/4-tax.js
```diff
@@ -5,7 +5,15 @@
   Sales tax is 20% of the price of the product
 */
 
+<<<<<<< Updated upstream
```
illicitonion: It looks like something weird happened with git here - You probably want to make sure to edit your master branch to make sure this doesn't have these <<>>> markers in it - we can talk about it on slack if you need a hand :)

week-4/Homework/mandatory/4-groceries.js
```diff
@@ -6,7 +6,20 @@
...
-  item2: "",
-  item3: ""
+  item1: "Potatoes",
+  item2: "Orange juice ",
+  item3: "Rice"
+
 };
+
+Object.values(groceryList).forEach(e => {
+  return groceriesToBuy.push(e);
```
illicitonion: This is great, good job!

Can you explain what the `return` here is doing? I think if you remove it, your code will do the same thing, so it would be great to think about what it means :)

week-4/InClass/C-more-complex-objects/exercise-2.js
```diff
@@ -8,6 +8,7 @@
 let house = {
     address: "1 Kinning Park",
     previousOwners: ["Claire M.", "John A."],
+    isForSale = false,
```
illicitonion: This works, but can you achieve the same thing by modifying the object _after_ it was written down (on line 30, like you did with `previousOwners`), rather than by changing how we defined the object in the first place?

week-4/InClass/C-more-complex-objects/exercise-3.js
```diff
@@ -32,7 +32,8 @@ let parkAvenueHouse = {
 
 // returns the full name (first name + last name) of the owner of the house
 function getOwnerFullName(house) {
-
+  let name = kinningParkHouse.currentOwner.firstName;
+  this.
```
illicitonion: Can you finish this one off? :)

week-4/InClass/E-arrays-of-objects/exercise-2.js
```diff
@@ -40,11 +40,25 @@ WRITE YOUR CODE BELOW
...
-let destinationNamesWithin500Kms = // Complete here
+let destinationNamesWithin500Kms = travelDestinations
+.filter(destination=>destination.distanceKms <500)  // Complete here
+.map(destination=>destination.destinationName);
 
-let destinationNameReachableByFerry = // Complete here
+console.log(destinationNamesWithin500Kms);
 
-let destinationNamesMoreThan300KmsAwayByTrain = // Complete here (PRINT THE RESULT IN THE CONSOLE USING FOREACH)
+let destinationNameReachableByFerry = travelDestinations
```
illicitonion: It looks like you didn't quite finish off this file - can you give it another go? We're always around on slack to talk things through if you need a hand :)

week-4/InClass/F-object-keys/exercise-part-1.js
```diff
@@ -15,9 +15,8 @@ let mentorsAges = {
 
 // ONLY EDIT BELOW THIS LINE
 
-let mentorsNames = ;
-
-let mentorsNamedUppercased = ;
+let mentorsNames = Object.keys(mentorsAges);
+let mentorsNamedUppercased = mentorsNames.toUpperCase();
```
illicitonion: This one doesn't quite work like this - `mentorNames` is an `Array`, and `toUpperCase` is a method on things with type `string` - can you think how to rewrite this so that you're applying `toUpperCase` to each element _inside_ `mentorNames`, rather than trying to apply it to the array itself?

You've done this before, so hopefully you can work it out again! :)

PR #595, opened 2020-03-30
week-4 mahmutSel

mahmutsel:


week-4/Homework/mandatory/2-writers.js
```diff
@@ -38,6 +38,11 @@ let writers = [
     alive: true
   }
 ];
+function getName(writers){
```
illicitonion: This works really well, good job!

One thing to think about that could make this code a little clearer is the naming. Let's look at this line - both of the names could be a little more clear. Let's start with the parameter - this function is actually only called with _one_ writer, not all of the writers, so `writer` may be a better name than `writers`.

And then the function name - `getName` - can you think about what this function does, and accordingly what may be a better name for it than `getName`?

week-4/Homework/mandatory/3-water-bottle.js
```diff
@@ -12,12 +12,18 @@ let bottle = {
...
   fill: function() {
     // calling this function should make you bottles volume = 100;
+    this.volume = 100;
   },
   drink: function() {
     // calling this function should decrease your bottles volume by 10;
+    this.volume -= 10;
   },
   empty: function() {
     // this function should return true if your bottles volume = 0
```
illicitonion: This is technically doing what the question asked, but I think implicit in this is that it should return `false` if volume _isn't_ `0` - can you do that too? And can you try to do so by making the function _smaller_ rather than _bigger_? (i.e. without adding an `else` clause?)

week-4/Homework/mandatory/4-groceries.js
```diff
@@ -6,7 +6,9 @@
...
 let groceryList = {
-  item1: "",
-  item2: "",
-  item3: ""
+  item1: "Potatoes",
+  item2: "Orange Juice",
+  item3: "Rice"
 };
+
+console.log();
```
illicitonion: Part one is done here, but can you finish off part 2 of this, to loop through and add the items to the array?

Thanks!

PR #600, opened 2020-04-01
JS Week 2 homework

melanieet:


melanieet:
Belated thank you, @ChrisOwen101!

PR #619, opened 2020-04-04
Week5 updated

abcelen:


week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,69 @@
...
+        jumbo.style.backgroundColor = "#87ca8a";
+        donateBtn.style.backgroundColor = "#black"; //error
+        volunteerBtn.style.backgroundColor = "8c9c08";
+    }
+
+
+    //Part 2 Special Thanks to Ferhat
+    let submitBtn = document.querySelector("form .btn");
+
+    function CheckForm() {
```
ChrisOwen101, line 46: Tiny thing: It's better if you use `checkForm()` using camelcase 

https://en.wikipedia.org/wiki/Camel_case

week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,69 @@
...
+
```
ChrisOwen101, line 47: I've tested this and while it works it refreshed the page onClick so you can't tell.

You have to use preventDefault
https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault

PR #647, opened 2020-04-11
Ellie-week6-js

ellietms:
# Your Details

Your Name: Ellie Tahmasebi
Your City: London
Your Slack Name: Ellie

# Homework Details

Module: js
Week: 6


week-6/Homework/mandatory/1-alarmclock/style.css
```diff
@@ -1,15 +1,85 @@
...
+  margin-bottom:5%;
+  margin-right: 5%;
 }
 
-h1 {
-  text-align: center;
+#timeRemaining{
+  margin-bottom: 15%;
+  font-family: 'Rye', cursive;
+  font-size: 200%;
```
LucyMac, line 28: Because `font-size: 200%;` is not inside a media query, it will only take effect for the screen widths you have left out in your media queries. 

The media queries you've used are:
`max-width:800px` --> from 0 to 800px
`(min-width:800px) and (max-width:1500px)` --> from 800px to 1500px

So what is left is 1500px and above. Which means `font-size: 200%;` will be applied to 1500px and above. Is this what you intended? If so, that's fine.

The "mobile first" approach, which is a very good practice, does things the other way around:

1.  First, you define some CSS for mobile devices, **without using a media query**. 

2. Then, you write some media queries starting at a certain size, for example 800px. This would mean using `(**min**-width:800px)` (not max-width). 

3. Don't set an upper limit in your media queries: only use `min-width: ....px` and don't set a max-width.

4. Keep your styles in ascending order: styles for smaller screens at the top, followed by medium screens, then the  biggest at the end.  This will allow you to keep better control of your styles, and nothing will overwrite anything else.

week-6/Homework/mandatory/1-alarmclock/style.css
```diff
@@ -1,15 +1,85 @@
...
+}
+#buttons{
+  margin-top: 10%;
 }
+@media(max-width:800px){
+  #timeRemaining{
+    font-size:x-large;
```
LucyMac, line 35: Try and avoid these terms like `x-large`, it's unclear how big that is. 
Stick to `px` or `rem` values, they are more predictable.

week-6/Homework/mandatory/1-alarmclock/style.css
```diff
@@ -1,15 +1,85 @@
...
+    margin-top:5%; 
+    font-size: medium; 
+  }
+  .settime_font{
+    display: inline-flex;
+    justify-content: space-around;
+    font-size:medium;
+  }
+  div.settime_font{
+    display:inline-flexbox;
```
LucyMac, line 49: This should be `inline-flex`, not `inline-flexbox`

week-6/Homework/mandatory/1-alarmclock/style.css
```diff
@@ -1,15 +1,85 @@
...
+  }
+}
+@media(min-width:800px) and (max-width:1500px){
+  #timeRemaining{
+    font-size:110%;
+    margin-bottom: 3rem;
+    display: inline-flex;
+  }
+  #buttons{
```
LucyMac, line 58: Please avoid using IDs in your stylesheets. If you have an ID for use in JavaScript, that's fine. But you should only use classes for CSS. So you can add one like this: 
`<div id="buttons" class="alarm-control-buttons">`

and then in your CSS:
```
.alarm-control-buttons {
   // your styles here
}
```

week-6/Homework/mandatory/1-alarmclock/style.css
```diff
@@ -1,15 +1,85 @@
...
+    margin-top:5%; 
+    font-size: large; 
+  }
+  .settime_font{
+    display: inline-flex;
+    justify-content: space-around;
+    font-size:large;
+  }
+  .centre{
+    display:grid;
```
LucyMac, line 69: Do you really mean to use `display: grid` here? What we learned in class was `display: flex`, and you are not using any of the other rules that work with `display: grid` here. It's fine to use `display: grid` if you understand how it works and if you use it alongside the other `grid` rules.  It think for this exercise it's a little over-kill to use this - I recommend `display: flex`

week-6/Homework/mandatory/1-alarmclock/style.css
```diff
@@ -1,15 +1,85 @@
...
+    display:inline-flexbox;
+  }
+  #set{
+    margin-right: 0.5rem;
+  }
+  #pause{
+    margin-right: 0.5rem;
+  }
+  #restart{
+    margin-right: 0.5rem;
```
LucyMac, line 81: Repeating styles like this is usually a sign that you could be re-using a CSS class.

For example: 
```
.alarm-control-button {
   margin-right: 0.5rem;
}
```

week-6/Homework/mandatory/2-quotegenerator/style.css
```diff
@@ -1 +1,143 @@
...
+
+#container {
+  background-color: #c8ccc9;
+  height: 20em;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-evenly;
+  align-items: center;
+  margin-top: 12em;
+  margin-left: 10em;
```
LucyMac, line 23: Avoid setting a specific `margin-left` if what you want to achieve is a centered element. 
Set `margin-left` and `margin-right` to `auto` and the element should be centered. Alternatively, wrap it all inside a flex container and set it to `justify-content: center`

LucyMac: (for this to work, the flex container would have to be the width of the screen, and the grey box would need to be **less** wide than its flex container.)

week-6/Homework/mandatory/3-slideshow/style.css
```diff
@@ -1 +1,90 @@
...
+}
+}
+@media(min-width:300px) and (max-width:600px){
+  body{
+    margin-right: 15%;
+    margin-top: 30%;
+  }
+  #images{
+    width:60%;
+  }
```
LucyMac, line 59: Your image would look better if it was bigger on mobile. Try
```
    width: 90%;
    margin: 5%;
```

PR #653, opened 2020-04-14
Nader js 2 week 2

naderakhgari:
# Your Details

Your Name: Nader Akhgari
Your City: 
Your Slack Name:

# Homework Details

Module:
Week: week 5

### DIFF
--- a/week-5/Homework/mandatory/2-exercises/exercises.js
+++ b/week-5/Homework/mandatory/2-exercises/exercises.js
@@ -14,7 +14,15 @@
  * </div>
  */
 function exerciseOne(arrayOfPeople) {
-  let content = document.querySelector("#content");
+  let content = document.querySelector("#content");//defined a variable and asigned element with Id 'content' to it.
+  arrayOfPeople.forEach(person => {//loop through arrayOfPeople
+    let names = document.createElement("h1");//made a h1 and asigned it to a variable
+    let jobs = document.createElement("h2");//made a h2 and asigned it to a variable
+    names.innerHTML = person.name;//asigned people name to h1 tag 
+    jobs.innerHTML = person.job;//asigned people job to h2 tag
+    content.appendChild(names);// 
+    content.appendChild(jobs);
+  })
 }
 
 /**
@@ -26,6 +34,14 @@ function exerciseOne(arrayOfPeople) {
  */
 function exerciseTwo(shopping) {
   //Write your code in here
+  let unorderedList = document.createElement("ul");
+  let content = document.querySelector("#content");
+  content.appendChild(unorderedList);
+  shopping.forEach(item => {
+    let items = document.createElement("li");
+    items.innerHTML = item;
+    unorderedList.appendChild(items);
+  })
 }
 
 /**
@@ -59,6 +75,21 @@ function exerciseTwo(shopping) {
 **/
 function exerciseThree(books) {
   //Write your code in here
+  
+  let unord = document.createElement("ul");
+  document.body.appendChild(unord);
+  books.forEach(book => {
+    let pEls = document.createElement("p");
+    pEls.innerHTML = `${book.title}, ${book.author}`;
+    console.log(pEls);
+    let listEls = document.createElement("li");
+    let imgEls = document.createElement("img");
+    unord.appendChild(listEls);
+    listEls.appendChild(pEls);
+    listEls.appendChild(imgEls);
+    (book.alreadyRead)? listEls.style.backgroundColor = "green" : listEls.style.backgroundColor = "red"; 
+  })
+  
 }
 
 //
--- a/week-5/Homework/mandatory/3-project/js/main.js
+++ b/week-5/Homework/mandatory/3-project/js/main.js
@@ -1,0 +1,79 @@
+////////Part 1
+/*
+Clicking on the buttons should change the "theme" of the website:
+- When clicking **blue** it should change:
+  - **Jumbotron** background color to `#588fbd`
+  - **Donate a bike** button background color to `#ffa500`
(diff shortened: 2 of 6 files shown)


week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,79 @@
...
+- For the **Email** field also check if it contains the `@` character
+For all the fields that invalid, it should make their background color `red`.
+IF all the fields are valid, when you click **Submit** it should:
+- Display an alert to thank you for filling out the form
+- Blank out (make empty) all the text fields
+**Important hint:** In your function that handles clicks on the `Submit` button you will need to call `event.preventDefault()` to stop the browser from refreshing the page. To read more on how to do this: https://developer.mozilla.org/en/docs/Web/API/Event/preventDefault
+*/
+
+let allValid = 0;//used this variable to check all inputs valid.
+document.querySelectorAll(".btn-primary")[5].addEventListener("click", function(){// main function click the submit.
```
ahmad-ali14, line 59: using `querySelectorAll()[5]` is not recommended. 
imagine if you added another button in the page for some reason. then you have to go all around and change every function in your code.
especially the class `btn-primary`  is widely used.
I would rather give this button an `id` attribute and select it with `document.querySelctor(#id)`

ahmad-ali14, line 59: this is a form submit button, it's more convenient to use `submit` event instead of `click`.
we use `click` when it's a single button without a from.
but when we have `form`, there should be a button that is responsible for this form action, we give it a special type of `submit`.

week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,79 @@
...
+}
+ValidateEmail(emlVal)//call the function validate email.
+let nameVal = document.querySelector("#example-text-input").value;//asigned the value of name input to nameVal.
+(nameVal.length > 0) ? allValid++ : document.querySelector("#example-text-input").style.backgroundColor = "red";
+
+
+let describeVal = document.querySelector("#exampleTextarea").value;
+(describeVal.length > 0) ? allValid++ : document.querySelector("#exampleTextarea").style.backgroundColor = "red";
+
+(allValid === 3) ? alert("thank you for filling out the form") && document.event.preventDefault() : {};//condition to check all inputs are valid.
```
ahmad-ali14, line 73: you are preventing default only when the form is submitted successfully.
it's more convenient to prevent default in case of errors, you can use `preventDefault()` on both sides of the ternary opereator as 

```javascript
(allValid === 3) ? alert("thank you for filling out the form") && e.preventDefault() :  e.preventDefault();
//condition to check all inputs are valid.
```

week-5/Homework/mandatory/3-project/js/main.js
ahmad-ali14, line 59: we have to pass `event` for this function, otherwise, the browser will not know which event we are talking about.

``` javascript
document.querySelectorAll(".btn-primary")[5].addEventListener("click", function(){//
```

week-5/Homework/mandatory/3-project/js/main.js
ahmad-ali14, line 73: `document.event` is not specific, we need to pass event for the function of `addEventListener()`

ahmad-ali14:
Good Job Nadir, 
the form error handling is perfect.
keep going

week-5/Homework/mandatory/3-project/js/main.js
naderakhgari: Hey Ahmad, thanks for your feedback, I've corrected it.

week-5/Homework/mandatory/2-exercises/exercises.js
```diff
@@ -59,6 +75,21 @@ function exerciseTwo(shopping) {
...
 function exerciseThree(books) {
   //Write your code in here
+  
+  let unord = document.createElement("ul");
+  document.body.appendChild(unord);
+  books.forEach(book => {
+    let pEls = document.createElement("p");
+    pEls.innerHTML = `${book.title}, ${book.author}`;
+    console.log(pEls);
+    let listEls = document.createElement("li");
```
ahmad-ali14, line 85: - you lift the images empty. you can do this by using `imgEls.src=' path/to/image '`
- you did not remove the bullets from ul. you can do this by using css property: 
```css
ul {
       list-style-type: none;
}
```
- you can use css: `width: 25%;`  to control each book container.
- you can use css: `float:left;` to align the book containers horizontally , You might find other options for this one.
- compared with this file: https://hyf-js2-week1-makeme-ex1-demo.herokuapp.com/index.html 

- resources:
list style : https://www.w3schools.com/css/css_list.asp
float: https://www.w3schools.com/css/css_float.asp
another float: https://developer.mozilla.org/en-US/docs/Web/CSS/float

week-5/Homework/mandatory/3-project/js/main.js
naderakhgari: thank you; I got it, its done

week-5/Homework/mandatory/3-project/js/main.js
naderakhgari: thank you, its done.

naderakhgari: Great, I got the point

week-5/Homework/mandatory/2-exercises/exercises.js
naderakhgari: You are right. I didn't style the page.

PR #664, opened 2020-04-18
Js2 week3/sabia

sabia-m:
all ex done with minimal / no styling - will continue to work on styling


week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,9 +1,34 @@
-function setAlarm() {}
+function setAlarm() {
```
illicitonion, line 1: This has the interesting effect that when the alarm is going off, the pages shows `00:01` - I would have expected it to show `00:00` - can you give that a go? :)

week-6/Homework/mandatory/1-alarmclock/test.js
```diff
@@ -0,0 +1,14 @@
+function formatSeconds(seconds) {
```
illicitonion, line 1: Can you talk through what this file is doing?

PR #712, opened 2020-04-22
Buchra atkeh week1

Buchraatkeh1984:
# Your Details

Your Name: Buchra Atkeh
Your City: Manchester
Your Slack Name: Buchra Atkeh

# Homework Details

Module:java script
Week:1


week-1/Homework/mandatory/3-function-output.js
```diff
@@ -1,14 +1,15 @@
...
 function s(w1, w2) {
     return w1.concat(w2);
-}
+}//returns a new string containing the text of the joined w1 and w2.
 
 function concatenate(firstWord, secondWord, thirdWord) {
+    return firstWord.concat(secondWord.concat(thirdWord))
     // Write the body of this function to concatenate three words together
     // Look at the test case below to understand what to expect in return 
 }
```
Ashaghel, line 15: your solution is correct. You can also simplify it by just adding more arguments to concat
like
```firstWord.concat(" ", secondWord, " ", thirdWord)```

File: week-1/Homework/mandatory/4-tax.js [hunk 2]
```diff
@@ -17,7 +19,13 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function formatCurrency() {}
+function formatCurrency(num) {
+  num = num*1.20;
+  num = num.toFixed(2)
+
+  return "£"+num.toString(); 
+
```
Ashaghel, line 27: Your fix is correct, the idea was to apply DRY more. So instead of 
```num=num*1.20 //you're using this in the previous function ```
you can use 
``` num=calculatetax(num) // or better even storing it in different variable```
```var taxedAmount= calculatetax(num) ```

File: week-1/Homework/mandatory/5-magic-8-ball.js [hunk 3]
```diff
@@ -45,17 +45,25 @@ Very doubtful.
...
 // - very negative
-function checkAnswer() {}
+function checkAnswer() {
+  let answer = "positive";
+
+return answer;
+}
 
 /* ======= TESTS - DO NOT MODIFY ===== */
 const log = console.log;
```
Ashaghel, line 69: Your solution stands actually, the exact application of TDD 👍 
However, the target solution is a bit different, I will try to explain it tomorrow.

PR #723, opened 2020-04-26
week7-homework-cynthia

cyeburu:
# Your Details

Your Name: Cynthia E
Your City: London
Your Slack Name: Cynthia E

# Homework Details
Debugging and Think like a programmer homework ex

Module:2
Week:7


illicitonion:
Well done - this is a great start :) There are a few more problems in the page - when I press "Submit", it logs an error to the console and doesn't add, and I'm not quite sure the "Read" tracking is working right, or the delete buttons. Can you give it another look to make sure all of the buttons are working correctly?

illicitonion:
Very nice, well done!

I still see one more bug in the library application, though, around the "Read" checkbox when creating a book - can you experiment with it and work out the one last problem?

PR #730, opened 2020-04-27
js-3-week-8

ferhatu:
# Your Details

Your Name:
Your City:
Your Slack Name:

# Homework Details

Module:
Week:

### DIFF
--- a/week-8/Homework/mandatory/1-practice/1-practice.md
+++ b/week-8/Homework/mandatory/1-practice/1-practice.md
@@ -24,8 +24,14 @@ The following endpoint is publicly available from Github
 
 1. Describe in a sentence what this API endpoint returns.
 
-<!-- -->
+<!--It returns a message which is not found(404) and object message
+and documentation_url ({
+message: "Not Found",
+documentation_url: "https://developer.github.com/v3/pulls/comments/#list-comments-on-a-pull-request"
+}) -->
 
 2. What would you put in the following fields? `{owner}`, `{repo}`, `{pull_number}`?
 
-<!-- Write your answer here -->
+<!-- {owner}: CodeYourFuture
+     {repo}: js-exercises
+    {pull_number}: 730-->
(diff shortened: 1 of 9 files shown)


week-8/Homework/mandatory/3-dog-photo-gallery/script.js
```diff
@@ -0,0 +1,19 @@
+let list = document.createElement("li");
+let image = document.createElement("img");
+
+let ulList = document.querySelector("#dog");
+ulList.appendChild(list);
+list.appendChild(image);
```
illicitonion, line 6: This looks great :) An extra question to think about is _when_ to create elements and put them in the page. Imagine the `fetch` here failed, or took a really long time - it may be kind of weird to have an image element in the page with no source. It may be useful to not call `appendChild` until we have a `src` for the image (or maybe even to have a placeholder text saying "Loading..." which we replace with the image). This applies to your answer to the "humour" question too :)

week-8/Homework/mandatory/3-dog-photo-gallery/script.js
```diff
@@ -0,0 +1,19 @@
...
+
+button.addEventListener("click", function () {
+  fetch("https://dog.ceo/api/breeds/image/random")
+    .then((response) => response.json())
+    .then((data) => {
+      //console.log(data);
+      image.src = data.message;
+    })
+    .catch((error) => {
+      console.error(error);
```
illicitonion, line 17: Logging to the console when there's an error is great for us as developers - can you think what else we may want to do in this function to make errors clear to the user on the website too?

week-8/Homework/mandatory/3-dog-photo-gallery/script.js
illicitonion: Actually, I just re-read the problem statement - each time you click it's meant to add a new dog to the list, rather than replacing each time (note the key word "append" in the question) - can you give that a go? :)

week-8/Homework/mandatory/1-practice/1-practice.md
```diff
@@ -24,8 +24,14 @@ The following endpoint is publicly available from Github
 
 1. Describe in a sentence what this API endpoint returns.
 
-<!-- Write your answer here -->
+<!--It returns a message which is not found(404) and object message
```
illicitonion, line 27: Can you describe what it returns when something _was_ found? e.g. with the parameters you described in part 2?

PR #735, opened 2020-04-27
All passed with Extras

elsahib:
# Your Details

Your Name: Osman Elsahib
Your City: Manchester
Your Slack Name: Osman Eslahib

# Homework Details

Module: JavaScript1
Week: 2

### DIFF
--- a/week-2/Homework/extra/2-array-methods.js
+++ b/week-2/Homework/extra/2-array-methods.js
@@ -5,7 +5,17 @@ Following are set of small functions that you need to create using loops and arr
 //1)Find the Smallest Number in an Array
 //Create a function that takes an array of numbers and return the smallest number in it
 
-function findSmallestNum(arr) {}
+function findSmallestNum(arr) {
+    let min = arr[0];
+    
+    for ( i = 1; i <=arr.length - 1;i ++ ){
+        if (arr[i] < min ){
+            min = arr[i];
+        }   
+    }
+    return min
+}
+// console.log(findSmallestNum([354, 15, 114, 2]));
 
 findSmallestNum([354, 15, 114, 2]); //Expected: 2
 
@@ -13,23 +23,54 @@ findSmallestNum([-76, 1, -79, 1, 0]); //Expected: -79
 
 //2)Checking Even Numbers
 //Create a function that takes in an array and returns true if all its values are even, and false otherwise.
-function checkAllEven(arr) {}
+function checkAllEven(arr) {
+    let result = true;
+    for ( i in arr ) {
+        if (!(arr[i] % 2 == 0)){
+            result = false;
+        }
+    }
+    return result
+}
 
 checkAllEven([1, 2, 2, 6, 9, 4]); //Expected: false
 
 checkAllEven([2, 4, 6]); //Expected: true
 //3)Half, Quarter and Eighth
 //Create a function that takes a number and return an array of three numbers: half of the number, quarter of the number and an eighth of the number.
-function getHalfQuarterEighth(number) {}
+function getHalfQuarterEighth(number) {
+    let result = []
+    result.push(number/2);
+    result.push(number/4);
+    result.push(number/8);
+    return result
+}
+// console.log(getHalfQuarterEighth(6));
+
 getHalfQuarterEighth(6); //Expected:[3, 1.5, 0.75]
 
 //4)Add the index to the number
 //Given an array of numbers, create a function which returns the same array but with each element's index in the array added to itself. This means you add 0 to the number at index 0, add 1 to the number at index 1, etc...
-function addIndexes(arr) {}
+function addIndexes(arr) {
+    let result = [];
+    for (i = 0;i <= arr.length -1;i ++){
+        result.push(arr[i]+i)
+    }
+    return result
+}
+// console.log(addIndexes([1, 4, 3, 4, 5]));
+
 addIndexes([1, 4, 3, 4, 5]); //Expected: [1, 5, 5, 7, 9]
 
 //4)Get the Sum of All Array Elements
 //Create a function that takes an array and returns the sum of all numbers in the array.
-function getSumOfItems(arr) {}
+function getSumOfItems(arr) {
+    let result= 0;
+    for (i in arr){
+        result += arr[i]
+    }
+    return result
+}
+console.log(getSumOfItems([-2, 84, 23]));
 
 getSumOfItems([-2, 84, 23]); //Expected: 105
--- a/week-2/Homework/extra/2-array-methods.js
+++ b/week-2/Homework/extra/2-array-methods.js
@@ -15,7 +15,7 @@ function findSmallestNum(arr) {
     }
     return min
 }
-// console.log(findSmallestNum([354, 15, 114, 2]));
+console.log(findSmallestNum([1]));
 
 findSmallestNum([354, 15, 114, 2]); //Expected: 2
 
@@ -71,6 +71,6 @@ function getSumOfItems(arr) {
     }
     return result
 }
-console.log(getSumOfItems([-2, 84, 23]));
+// console.log(getSumOfItems([-2, 84, 23]));
 
 getSumOfItems([-2, 84, 23]); //Expected: 105
--- a/week-2/Homework/mandatory/1-fix-functions.js
+++ b/week-2/Homework/mandatory/1-fix-functions.js
@@ -11,7 +11,7 @@
 //    Only make edits inside the function
 
 function mood() {
-  let isHappy = true;
+  let isHappy = false;
 
   if (isHappy) {
     return "I am happy";
@@ -26,7 +26,7 @@ function mood() {
 
 function greaterThan10() {
   const num = 10;
-  const isBigEnough;
+  const isBigEnough = true ;
 
   if (isBigEnough) {
     return "num is greater than or equal to 10";
@@ -41,7 +41,7 @@ function greaterThan10() {
 
 function sortArray() {
   const letters = ["a", "n", "c", "e", "z", "f"];
-  let sortedLetters;
+  let sortedLetters = letters.sort();
 
   return sortedLetters;
 }
@@ -52,10 +52,11 @@ function sortArray() {
 
 function first5() {
   let numbers = [1, 2, 3, 4, 5, 6, 7, 8];
-  let sliced;
+  let sliced = numbers.splice(0,5);
 
   return sliced;
 }
+
 
 // 5) get3rdIndex function needs to take an array `arr` and give 
 //    back third element value.
@@ -65,7 +66,7 @@ function first5() {
 
 function get3rdIndex(arr) {
   let index = 3;
-  let element;
+  let element = arr[index];
 
   return element;
 }
--- a/week-2/Homework/mandatory/1-fix-functions.js
+++ b/week-2/Homework/mandatory/1-fix-functions.js
@@ -25,8 +25,9 @@ function mood() {
 //    Variable isBigEnough needs to evaluate to a boolean
 
 function greaterThan10() {
+  let input = 20
   const num = 10;
-  const isBigEnough = true ;
+  const isBigEnough =(input > num) ;
 
   if (isBigEnough) {
     return "num is greater than or equal to 10";
--- a/week-2/Homework/mandatory/2-function-creation.js
+++ b/week-2/Homework/mandatory/2-function-creation.js
@@ -47,12 +47,13 @@ The function must:
 function remove(arr, index) {
   let newArray = [];
   for (i in arr){
-    if (!(arr[i] == arr[index])){
+    if (!(i == index)){
        newArray.push(arr[i]);
     }
-  };
-  return newArray;
+  };  
+  return newArray; 
 }
+console.log(remove(["a", "a", "a", "a", "a", "a", "a"], 5));
 
 /*
 Write a function that:
@@ -77,6 +78,7 @@ function formatPercentage(arr) {
   }
   return finalArr
 }
+// console.log(formatPercentage([23, 18, 187.2, 0.372,12.2]));
 
 /* ======= TESTS - DO NOT MODIFY ===== */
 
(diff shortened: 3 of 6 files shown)


week-2/Homework/mandatory/1-fix-functions.js
```diff
@@ -26,7 +26,7 @@ function mood() {
 
 function greaterThan10() {
   const num = 10;
-  const isBigEnough;
+  const isBigEnough = true ;
```
zoltan-gal, line 29: Well, this is right :), however, the aim would have been to use comparison operation to calculate the boolean value instead of using a baked in value. It could then reflect upon any changes in case content `num` changed.

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -5,7 +5,17 @@ Write a function that:
...
 - makes the string all lowercase
 */
-function tidyUpString(strArr) {}
+function tidyUpString(strArr) {
+  let newArr=[];
+  for (i in strArr){
+    let x = strArr[i].trim();
+    x = x.replace("/","");
+    x = x.toLowerCase();
+    newArr.push(x);
```
zoltan-gal, line 14: A big point for the functional approach that you create a new array as an output rather than modifying the input array, it makes it a  function without having no side effect.

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -15,7 +25,16 @@ Complete the function to check if the variable `num` satisfies the following req
...
 
-function validate(num) {}
+function validate(num) {
+  if (typeof num == 'number' && num % 2 == 0 && num <=100 )
+  {
+    return true;
+  }
+  else {
+    return false;
+  };
```
zoltan-gal, line 35: It's a correct solution but do we really need the if-else statement?

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -26,7 +45,13 @@ The function must:
...
 
 function remove(arr, index) {
-  return; // complete this statement
+  let newArray = [];
+  for (i in arr){
+    if (!(arr[i] == arr[index])){
+       newArray.push(arr[i]);
+    }
+  };
+  return newArray;
```
zoltan-gal, line 54: Are you sure that this function would pass in any case? What if the same element appears in the array more than one time?

week-2/Homework/mandatory/2-function-creation.js
```diff
@@ -37,7 +62,21 @@ Write a function that:
...
+function formatPercentage(arr) {
+  let finalArr = [];
+  for ( i in arr ){
+    let x = arr[i];
+    if (x > 100){
+      x = 100
+    }
+    if ( x % 1 !== 0 ){
+    x = (Math.round(x * 100) / 100).toFixed(2)
+    }
```
zoltan-gal, line 74: Wondering what's the additional value of the if statement?

week-2/Homework/mandatory/2-function-creation.js
elsahib: Hi, if you notice this function takes 2 parameters, the second one is a number (index value ) not a value of the element. I believe indexes are uniqe in an array even if the elements of the array are not.

week-2/Homework/mandatory/2-function-creation.js
elsahib: I didn't understand the question!

week-2/Homework/mandatory/2-function-creation.js
elsahib: we can write it like this I think:
`function validate(num) {
  return (typeof num == 'number' && num % 2 == 0 && num <=100 )
}`

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -14,11 +14,11 @@
...
   1. This program throws an error. Why? (If you can't find it, try executing it).
   2. Remove the line that throws the error.
-  3. What is printed to the console?
-  4. How many times is "f1" called?
-  5. How many times is "f2" called?
-  6. What value does the "a" parameter take in the first "f1" call?
-  7. What is the value of the "a" outer variable when "f1" is called for the first time?
+  3. What is printed to the console? (2 6 4 9 6 13 8 )
+  4. How many times is "f1" called? 2
+  5. How many times is "f2" called? 5
```
zoltan-gal, line 19: Could you please explain how you got this count?

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -14,11 +14,11 @@
...
+  6. What value does the "a" parameter take in the first "f1" call? 6
```
zoltan-gal, line 20: Not sure that's right what do you think?

week-2/Homework/mandatory/3-playing-computer.js
elsahib: I used break points :)

week-2/Homework/mandatory/3-playing-computer.js
```diff
@@ -14,11 +14,11 @@
...
-  3. What is printed to the console?
-  4. How many times is "f1" called?
-  5. How many times is "f2" called?
-  6. What value does the "a" parameter take in the first "f1" call?
-  7. What is the value of the "a" outer variable when "f1" is called for the first time?
+  3. What is printed to the console? (2 6 4 9 6 13 8 )
+  4. How many times is "f1" called? 2
+  5. How many times is "f2" called? 5
+  6. What value does the "a" parameter take in the first "f1" call? 6 
+  7. What is the value of the "a" outer variable when "f1" is called for the first time? 6
```
zoltan-gal, line 21: Based on the initial value of the outer variable "a", can it be 6 if that is incremented in each iteration beforehand calling any function?

week-2/Homework/mandatory/1-fix-functions.js
elsahib: I think I'm puzzeled by this one, because isBigEnough was declared as a const and wasn't initialised, now if I change it to let and compare it in the if statement to num, it will throw an error as it holds no value. and the instructions says that i can only change isBigEnough. 
to use a comparison operator i'll need a third variable or input to assign the result of the comparison to isBigEnough. for example :
`  let input = 20
  const num = 10;
  const isBigEnough =(input > num) ;`
I can't really think of anyother way.

week-2/Homework/mandatory/2-function-creation.js
zoltan-gal: I'll take a 2nd look at, might have missed something.

zoltan-gal: Let me ask another way. When will the expression in the IF statement be FALSE?
Can you say me an example when it wouldn't be true?

week-2/Homework/mandatory/2-function-creation.js
zoltan-gal: I'll take a 2nd look at, might have missed something.

week-2/Homework/mandatory/2-function-creation.js
elsahib: this condition will evaluate to false when a decimal number is passed to it, as the reminder/modulos operator will always return the fractional part of the number if it's devided by 1.
for example if we do (10.5 % 1 ) this will return 0.5 which will evaluate to FALSE in our case here.

week-2/Homework/mandatory/4-sorting-algorithm.js
```diff
@@ -21,7 +21,15 @@ You don't have to worry about making this algorithm work fast! The idea is to ge
...
 function sortAges(arr) {
-  //create function here
+  let newArr = [];
+  for ( i in arr ) {
+    if ( typeof arr[i] == "number" ){
+      newArr.push(arr[i]);
+    }
+  }
+  newArr = newArr.sort(function(a, b){return a-b});
+  return newArr
```
zoltan-gal, line 31: It's right! Just an extra tip:
You can sort the array in-line with the return statement, don't need to reassign it to a variable if you don't use it anymore.
As a result, you can simply comprise into one line:
`return newArr.sort(function(a, b){return a-b});`

week-2/Homework/extra/2-array-methods.js
```diff
@@ -5,31 +5,72 @@ Following are set of small functions that you need to create using loops and arr
 //1)Find the Smallest Number in an Array
 //Create a function that takes an array of numbers and return the smallest number in it
 
-function findSmallestNum(arr) {}
+function findSmallestNum(arr) {
+    let min = arr[0];
+    
+    for ( i = 1; i <=arr.length - 1;i ++ ){
```
zoltan-gal, line 11: It's not necessary to subtract 1, you can simply use `<` operator instead

week-2/Homework/extra/2-array-methods.js
```diff
@@ -5,31 +5,72 @@ Following are set of small functions that you need to create using loops and arr
 //1)Find the Smallest Number in an Array
 //Create a function that takes an array of numbers and return the smallest number in it
 
-function findSmallestNum(arr) {}
+function findSmallestNum(arr) {
+    let min = arr[0];
```
zoltan-gal, line 9: Overall is great, however, we as programmers sometimes we need to think about edge cases or boundary cases in order to make this function rock solid within a sensible limit. We haven't stipulated in our function contract how many elements the array can have so it can be from zero to "infinite".

What if it is empty? What'd occur in that case?

week-2/Homework/extra/2-array-methods.js
```diff
@@ -5,31 +5,72 @@ Following are set of small functions that you need to create using loops and arr
...
 findSmallestNum([-76, 1, -79, 1, 0]); //Expected: -79
 
 //2)Checking Even Numbers
 //Create a function that takes in an array and returns true if all its values are even, and false otherwise.
-function checkAllEven(arr) {}
+function checkAllEven(arr) {
+    let result = true;
+    for ( i in arr ) {
+        if (!(arr[i] % 2 == 0)){
+            result = false;
```
zoltan-gal, line 30: Not sure if you've already learnt it or not, it might be advanced tip at this point but
You can immediately exit the loop using `return false` statement once we know array doesn't consist of only pure even numbers. When you use `return` it will interrupt and leave the function computation at whichever point it was. (You can put as many return points into your function as you want.) In this case, at the end of this function, you can always say `return true` because we can reach that point if and only if the loop hasn't ended the function call by returning a value earlier.

week-2/Homework/mandatory/1-fix-functions.js
zoltan-gal: I've thought of something like this:
`consts isBigEnough = num >= 10`

You don't need to use a third variable, you can use any constant value in your expression.

zoltan-gal: But I agree that this exercise is a bit flaky.

week-2/Homework/mandatory/2-function-creation.js
zoltan-gal: You're right, the index is unique and you get an index number as the 2nd parameter.
So let's revise the criteria once more especially this one:
 * `remove the item at the specified index`

Do you think the condition `if (!(arr[i] == arr[index]))`, you have, is checking the position of the element in the array?

What do you think, what should be the outcome of the following example?
`remove(["a", "a", "a", "a", "a", "a", "a"], 6)`

week-2/Homework/mandatory/2-function-creation.js
zoltan-gal: You're absolutely right. I've missed that it can even be a floating number and not only whole number.

week-2/Homework/mandatory/2-function-creation.js
zoltan-gal: Exactly, as that expression results in a Boolean value. So you don't need to "re-specify" that it is true or false depending on the outcome, you can simply use the outcome of that expression.

week-2/Homework/mandatory/2-function-creation.js
zoltan-gal: Tip: `toFixed` function does not only format the number but also rounds it.
It can make your rounding much simpler.

Here is an example: https://www.w3schools.com/jsref/jsref_tofixed.asp
For more: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed

week-2/Homework/mandatory/3-playing-computer.js
zoltan-gal: It's a good tactic :) but you must have miscounted at some point.
If there is only 5 iteration and we call `f2` only when `if (i % 2 === 0)` is true than `f2` cannot be invoked 5 times.

week-2/Homework/mandatory/2-function-creation.js
elsahib20: Hi, it is returning an empty array !!! after I ran this test I realised that this condition is comparing values not indexes. I'll have to research a bit further on this one

week-2/Homework/extra/2-array-methods.js
elsahib20: I can think of a solution for the Zero case, but for the infinte, my solution would be to wrap the whole body of the function inside of a `if (arr.length >0 ) { body of the function }`

elsahib20:
Zoltan, Thank you for your review, but I wish if you can explain to me some of the parts where I have failed to make the propper function. and specially the "thinking like a computer " excercies as I have failed in that one terribally.

week-2/Homework/mandatory/2-function-creation.js
elsahib20: I figured it out, whenever I write `arr[i]` I'll be accessing the value of the item in that index - which I don't want for the purposes of this function- the correct condition should be `if (!(i == index))` this way I'm only dealing with indexes.
This took me a while to understand. Thanks again Zoltan

week-2/Homework/extra/2-array-methods.js
zoltan-gal: It is a good approach! If array is empty, the function would return `undefined` (I believe so)

There are other technics as well, but it is not out of the scope of the subject.

PR #744, opened 2020-04-29
Yohannes week 8

byohannes:
# Your Details

Your Name: B Yohannes
Your City: London
Your Slack Name: Yohannes

# Homework Details
We have been asked to do fetching data from API endpoints as exercises and projects.
Module: JS 3
Week: 8


week-8/Homework/mandatory/2-fetch-exercise/exercise.js
```diff
@@ -1,26 +1,8 @@
...
-    .then(function(greeting) {
-        // Write the code to display the greeting text here
-    });
\ No newline at end of file
+fetch ('https://codeyourfuture.herokuapp.com/api/greetings')
+  .then (function (response) {
+    return response.text ();
+  })
+  .then (function (greeting) {
+    document.querySelector ('#greeting-text').innerHTML = greeting;
```
illicitonion, line 7: This works, but we should generally be wary of injecting HTML into our page which comes from sources we don't trust. I'd think about setting `textContent` rather than `innerHTML`, so that the browser doesn't try to interpret the `greeting` as HTML :)

week-8/Homework/mandatory/3-dog-photo-gallery/index.html
```diff
@@ -0,0 +1,18 @@
...
+  <head>
+    <meta charset="utf-8" />
+    <title>Getting Dog images From API</title>
+    <link rel="stylesheet" href="style.css" />
+  </head>
+
+  <body>
+    <div id="content">
+      <ul id="uList">
+        <li id="list"><img id="image" /></li>
```
illicitonion, line 12: This works, but on the first load of the page, before clicking the button, it looks a little weird :) Can you think of a way to make the page not have a strange black blob displayed?

week-8/Homework/mandatory/3-dog-photo-gallery/script.js
```diff
@@ -0,0 +1,25 @@
...
+      if (response.status >= 200 && response.status <= 299) {
+        return response.json ();
+      } else {
+        throw new Error (
+          `Encountered something unexpected: ${response.status} ${response.statusText}`
+        );
+      }
+    })
+    .then (data => addImage (data))
+    .catch (error => console.log (error));
```
illicitonion, line 14: This works really well for showing the developer the error information, but it doesn't really tell a normal user that anything has gone wrong. Can you think how to also let the user know what's happening / what to expect? Similarly for the humour question :)

week-8/Homework/mandatory/3-dog-photo-gallery/index.html
illicitonion: Actually, sorry, I just re-read the task - it should be _adding_ an image each time you click, rather than replacing - can you give that a go? :)

week-8/Homework/mandatory/3-dog-photo-gallery/script.js
byohannes: Thanks, Daniel I have updated all the recommendations now.

week-8/Homework/mandatory/3-dog-photo-gallery/index.html
byohannes: Thanks, Daniel I have done it mate.thanks again.

week-8/Homework/mandatory/2-fetch-exercise/exercise.js
byohannes: Thanks, Daniel I have updated all the recommendations now.

week-8/Homework/mandatory/3-dog-photo-gallery/script.js
illicitonion: I think this will still only appear in the console, right? Have a think about (thought up to you whether you want to implement) what kinds of things could you do which would show up to users :)

byohannes: Would you think that writting  if (!response.ok) {
          throw Error(response.statusText);
        } would be better?

illicitonion: Where would you expect that to show up?

In general, the DOM is the mechanism we have to change the page from javascript so that users can see information we want to share

byohannes: So we can create p tags using Dom JS then we can display the error message there instead of the console log I would suggest this is the solution you are expecting isn't?

illicitonion: Exactly :)

PR #749, opened 2020-04-29
Week8 - js

ellietms:
# Your Details

Your Name: Ellie Tahmasebi
Your City: London
Your Slack Name: Ellie

# Homework Details

Module: javascript
Week: 8


week-8/Homework/mandatory/3-dog-photo-gallery/dog-photo.js
```diff
@@ -0,0 +1,35 @@
...
+        .then(checkResponse)
+        .then(response => 
+            response.json()
+        )   
+        .then(data => {
+           const myImageContainer  = document.querySelector("#imageContainer")
+           let  divContainer = document.querySelector(".card-body")
+           const  myLi = document.createElement("li")
+           const myImage = document.createElement("img")
+            myImage.className += "card-img-top w-60 m-auto p-auto"
```
illicitonion, line 14: The `+=` is interesting here - what it does is add this string to the existing `className` string - because you've just created the element, we know that's empty, so this is the same as `myImage.className = ...`

There's also a handy property called `classList` we can use to add classes, so that we don't need to worry about whether there are existing classes, or spaces at the end -you can read about it [here](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) :)

week-8/Homework/mandatory/3-dog-photo-gallery/dog-photo.js
```diff
@@ -0,0 +1,35 @@
...
+myButton.addEventListener("click",getFunctionFetch);
+function getFunctionFetch(){
+    fetch("https://dog.ceo/api/breeds/image/random")
+        .then(checkResponse)
+        .then(response => 
+            response.json()
+        )   
+        .then(data => {
+           const myImageContainer  = document.querySelector("#imageContainer")
+           let  divContainer = document.querySelector(".card-body")
```
illicitonion, line 11: I'm curious - why is this one `let` where the other ones are `const`?

week-8/Homework/mandatory/3-dog-photo-gallery/dog-photo.js
```diff
@@ -0,0 +1,35 @@
...
+           const myImage = document.createElement("img")
+            myImage.className += "card-img-top w-60 m-auto p-auto"
+            myImage.src = data.message
+            myLi.appendChild(myImage)
+            myImageContainer.appendChild(myLi)
+            divContainer.appendChild(myImageContainer);
+            
+        })
+        .catch((err) => {
+            console.log('Error has occurred: ' + err);
```
illicitonion, line 22: A more advanced topic to think about: This is really useful for us as developers, but when the general public use websites, they tend not to look at the console, and may be confused if nothing shows up in their website. It would be interesting to think about how we could also surface errors in a way that tells the user that something has gone wrong.

PR #780, opened 2020-05-01
Hacer's homework

hacersk:
# Your Details

Your Name:Hacer
Your City:Glasgow
Your Slack Name:Hacer

# Homework Details

Module: js
Week: 1

### DIFF
--- a/week-1/Homework/mandatory/3-function-output.js
+++ b/week-1/Homework/mandatory/3-function-output.js
@@ -1,16 +1,24 @@
 // Add comments to explain what this function does. You're meant to use Google!
+
+//Mart.random() returns a random number between 0 (inclusive),  and 1 (exclusive).
+// getNumber() returns a random integer from 0 to 9
 function getNumber() {
     return Math.random() * 10;
 }
 
+
 // Add comments to explain what this function does. You're meant to use Google!
+
+//The concat() method is used to join two or more strings.
 function s(w1, w2) {
+    
     return w1.concat(w2);
 }
 
 function concatenate(firstWord, secondWord, thirdWord) {
     // Write the body of this function to concatenate three words together
     // Look at the test case below to understand what to expect in return 
+    return firstWord.concat(" " +secondWord).concat(" "+thirdWord);
 }
 
 /* ======= TESTS - DO NOT MODIFY ===== */
--- a/week-1/Homework/mandatory/4-tax.js
+++ b/week-1/Homework/mandatory/4-tax.js
@@ -5,7 +5,11 @@
   Sales tax is 20% of the price of the product
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(productprice) {
+
+  return productprice+(productprice*20/100);
+
+}
 
 /*
   CURRENCY FORMATTING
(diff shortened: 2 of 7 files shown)


week-1/Homework/mandatory/3-function-output.js
```diff
@@ -1,16 +1,24 @@
...
+//Mart.random() returns a random number between 0 (inclusive),  and 1 (exclusive).
+// getNumber() returns a random integer from 0 to 9
 function getNumber() {
     return Math.random() * 10;
 }
 
+
 // Add comments to explain what this function does. You're meant to use Google!
+
+//The concat() method is used to join two or more strings.
```
filkovsp, line 12: Here must be comment for the function "s".
Please try doing this in VS Code:
1. put /**  --> this must be just 1 line above the function name
2. press enter button
3. see what VS Code will do for you.
Hint: it must be a multi-line comment, parameterised with function's arguments. 
Where you need to shortly specify what the function does at all, and what each of its arguments are for.

filkovsp, line 12: whatever related to concat() is inside the "s" function, 
so, it good practice to such comment inside the function's body and above the line there where concat() is used, or at the end of that line (less common practice).

week-1/Homework/mandatory/3-function-output.js
```diff
@@ -1,16 +1,24 @@
 // Add comments to explain what this function does. You're meant to use Google!
+
+//Mart.random() returns a random number between 0 (inclusive),  and 1 (exclusive).
```
filkovsp, line 3: Math.random() is inside the function, it's better to comment it there.
Nice comment for a function in javascript will be looking like this: [JS Doc support](https://code.visualstudio.com/docs/languages/javascript#_jsdoc-support)

week-1/Homework/mandatory/4-tax.js
```diff
@@ -5,7 +5,11 @@
   Sales tax is 20% of the price of the product
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(productprice) {
+
+  return productprice+(productprice*20/100);
```
filkovsp, line 10: we only need to calculate SalesTax out of Product Price.
---------- 100%   <-- if this product price
-- 20%                 <-- this is what makes 20% of it
But if we return productprice + 20% of it, it will be:
------------ 120%  <-- this much
but function calculateSalesTax need only to return calculated 20% Tax.

week-1/Homework/mandatory/5-magic-8-ball.js
```diff
@@ -45,17 +45,23 @@ Very doubtful.
 
 // This should log "The ball has shaken!"
 // and return the answer.
-function shakeBall() {}
+function shakeBall() {
+   console.log("The ball has shaken!");
```
filkovsp, line 49: nice try, Hacer.
now when you know arrays and if/else would like to think about how you could implement some random element into this function?

there are 5 positive options, 5 very positive, also bunch of other options, 4 x 5 total.
But we always use here only one: " Outlook good " --> with extra spaced around, and we always git it "positive" (with small 1st letter) value. But what about all the others? 
Ask a question, if you need any tip/hint/help :)

PR #804, opened 2020-05-05
week 3

andsemenov:
# Your Details

Your Name:Andrei
Your City:Glasgow
Your Slack Name: andsemenov

# Homework Details

Module:JS
Week:week 3


week-2/Homework/F-logical-operators/exercise2.js
```diff
@@ -5,8 +5,33 @@
...
+  }
+  return false;
+}
+
+function startsWithD(name) {
+  if (name[0] === "D") {
+    return true;
+  }
+  return false;
+}
```
alastair87, line 34: Correct. However, you could write the each of the bodies for the above functions in one line, without needing to use an if block. Think about what the expression you are using for the if conditional evaluates to, and what you need the function to return.

week-1/Homework/extra/2-piping.js
```diff
@@ -15,44 +14,53 @@
...
   the final result to the variable goodCode
 */
-
-function add() {
-
+let a, b, c, d;
+function add(a, b) {
+  let summa = Math.round((a + b) * 10) / 10;
+  return summa;
 }
```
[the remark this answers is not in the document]
andsemenov: Thanks for giving the feedback! I tried to do the function add (a,b) {let summa=a+b; return summa;} But the result couldn't pass automatic test in 2-piping.js because it add 2.4 to 5.3 and summa equals 7,6(9). I agree that it looks strange, but it works for me.

week-2/Homework/F-logical-operators/exercise2.js
andsemenov: Thank a lot. I understand what you tell about. I will change the code in order to do it better. Thank you.

week-3/Homework/A-array-find/exercise.js
```diff
@@ -4,10 +4,23 @@
...
+  "Andronicus",
+  "Annam",
+  "Mikey",
+  "Anastasia",
+  "Karim",
+  "Ahmed",
+];
 
-var longNameThatStartsWithA = findLongNameThatStartsWithA(names);
+var longNameThatStartsWithA = names.find(findLongNameThatStartsWithA);
```
alastair87, line 24: This gives the right result but it isn't really a good way of doing it. The find function expects a callback that returns a boolean (i.e. true or false) and when it gets true it then returns the corresponding value of the array, not the value that was returned by your callback function works.

The reason your function still works in practice is because Javascript has a concept called 'truthyness' as well as 'falsyness'. It's quite complicated, but in this case what it means is that when it sees any string where it might expect a boolean, it treats it the same as the boolean value of `true`. When nothing is returned at all it sees it as `false`. That's why your code still gives the right result. However you should always seek to return a `true` or `false` when a boolean is expected.

week-3/Homework/E-array-map/exercise.js
```diff
@@ -3,3 +3,9 @@
 
 var numbers = [0.1, 0.2, 0.3, 0.4, 0.5];
 
+function multiple100(number) {
+  return number * 100;
+}
+var numbersMultiplied = numbers.map(multiple100);
+
+console.log(numbersMultiplied);
```
alastair87, line 11: You should write code to cover the other examples. It's important to understand the different ways functions are expressed and used in Javascript.

week-3/Homework/D-array-filter/exercise.js
```diff
@@ -6,14 +6,24 @@
...
+var pairsByIndexRaw = [[0, 3], [1, 2], [2, 1], null, [1, 0], false, "whoops"];
+function filteredPairs(pair) {
+  return (
+    pair !== null &&
+    typeof pair === "object" &&
+    pair.length == 2 &&
+    pair[0] >= 0 &&
+    pair[1] >= 0
+  );
+}
```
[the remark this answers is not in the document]
andsemenov: Thank you. I did not know about this method, it is very useful. Thank you so much.

week-3/Homework/A-array-find/exercise.js
andsemenov: Thanks for your feedback. I understand how it should work. My function will return the result of statement like `return name.length > 7 && name.startsWith("A");`
Thank you so much.

week-3/Homework/E-array-map/exercise.js
andsemenov: Thanks for your feedback. I forgot to write different ways. I will add it asap. Thank you so much.

week-3/Homework/D-array-filter/exercise.js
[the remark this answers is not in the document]
alastair87: You only need to check two of the conditions above - since `Array.isArray()` will be `false` if `pair` is `null`. There's no reason the values couldn't be negative for the indexes, so I wouldn't check for that but if this were a real project it's the sort of thing I would probably want to ask whoever wrote the requirement.

week-3/Homework/F-array-forEach/exercise.js
```diff
@@ -9,6 +9,22 @@
...
+    newName = "Buzz";
+  }
+
+  if (name % 3 === 0 && name % 5 === 0) {
+    newName = "FizzBuzz";
+  }
+
+  console.log(newName);
+});
 /* EXPECTED OUTPUT */
```
alastair87, line 28: Good. This is quite an interesting problem, since its quite simple but there are lots of different ways of doing it. It's quite often used as a technical question in job interviews.

week-3/Homework/G-extra/1-map.js
```diff
@@ -10,10 +10,24 @@ var stationTransportOptionsPairs = [
...
+  }
+  return;
+}
 
-var stationsWithRiverBoat; // <-- Complete this statement
+var stationsWithRiverBoat = stationTransportOptionsPairs
+  .map(riverBoatFiltered)
+  .filter((station) => {
+    return station !== undefined;
+  }); // <-- Complete this statement
```
alastair87, line 31: This works but it's not really what's being looked for. The `.filter()` function is intended for applying a filter, whilst `.map()` is in tended for doing something based on each value of the array, such as only returning part of it. It's also expected that the function with which you are filtering return explicit `true` or `false` values. So, for example, whether it's `true` or `false` that transport includes river boat. Then you can map the filtered values to the names of the stations.

week-3/Homework/G-extra/2-map.js
```diff
@@ -11,10 +11,18 @@ var attendanceCounts = [
...
-var eligibleStudentNames; // TODO: Complete this line.
+var eligibleStudentNames = attendanceCounts
+  .map((pair) => {
+    let name = pair[0];
+    let attendance = pair[1];
+    if (attendance >= 8) return name;
+  })
+  .filter((name) => {
+    return name != undefined;
+  }); // TODO: Complete this line.
```
alastair87, line 26: This is similar to the previous example, that its better to apply filter first, then map the filtered array to the values that you want.

week-3/Homework/G-extra/1-map.js
andsemenov: Thanks for your feedback. Yes, I have changed the code and committed it. It became shorter and more clear. Thanks again.

week-3/Homework/G-extra/2-map.js
andsemenov: You are absolutely right. My first code looks not optimal, I have changed it. Thanks for your feedback.

week-3/Homework/F-array-forEach/exercise.js
andsemenov: Thanks for you feedback. I will find different ways of solving it. It is really interesting task, I enjoyed it.

week-3/Homework/D-array-filter/exercise.js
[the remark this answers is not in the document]
andsemenov: Thanks for your feedback. I removed excess conditions, now code looks more optimal. Thank you.

PR #807, opened 2020-05-06
Yohannes week 9

byohannes:
# Your Details

Your Name: Berhane Yohannes
Your City: London
Your Slack Name: Yohannes

# Homework Details

Module: JS 3
Week: week 9

### DIFF
--- a/week-9/Homework/mandatory/1-practice/2-code-reading.md
+++ b/week-9/Homework/mandatory/1-practice/2-code-reading.md
@@ -1,65 +1,61 @@
-# Code Along
-
-In the following "code along" you'll be building a complete Weather App that makes use of the [Darksky API](https://darksky.net).
-
-Enjoy!
-
-- [Build a Weather App with Vanilla JavaScript Tutorial](https://www.youtube.com/watch?v=wPElVpR1rwA)
-
-
 # Code reading
+
 ## Question 1
+
 Take a look at the following code:
 
 ```
 1    let x = 1;
 2    {
 3        let x = 2;
-4        console.log(x); 
+4        console.log(x);
 5    }
-6    console.log(x); 
+6    console.log(x);
 ```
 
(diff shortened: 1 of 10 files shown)


week-9/Homework/mandatory/1-practice/2-code-reading.md
```diff
@@ -14,6 +14,7 @@ Take a look at the following code:
 ```
 
 Explain why line 4 and line 6 output different numbers.
+The first console will display 2 as x scoped to its block while the second console will display 1 as x is a global in that case.
```
ahmad-ali14, line 17: perfect

week-9/Homework/mandatory/1-practice/2-code-reading.md
```diff
@@ -33,6 +34,8 @@ console.log(y)
 ```
 
 What will be the output of this code. Explain your answer in 50 words or less.
+After the function is called the console inside the function will display 10 then te last console will display undefined as y can not be accessed outside its block.
```
ahmad-ali14, line 37: - the function has been executed inside the console.log.
- f1 does not have return statement so it will log either undefined or log the function statement .
- it will logs x anyway  while it's been executed 
- y undefined.

week-9/Homework/mandatory/1-practice/2-code-reading.md
```diff
@@ -61,3 +64,4 @@ console.log(y);
 ```
 
 What will be the output of this code. Explain your answer in 50 words or less.
+ console.log(x) will display 9 on the screen as x is a global, while console.log(y) will display {x:10}  as val.x will refer to the same reference of x on object y then it will add 1 to it  as y is global.
```
ahmad-ali14, line 67: - correct.
- f1 will return 10.
- console.log x will return 9. 
- x is const 
-  y will log 10

PR #810, opened 2020-05-06
BuchraAtkeh week3 work

Buchraatkeh1984:
# Your Details

Your Name:Buchra Atkeh
Your City:Manchester
Your Slack Name: Buchra Atkeh

# Homework Details

Module:java script
Week:3


week-3/Homework/mandatory/5-journey-planner.js
```diff
@@ -84,7 +88,14 @@ function getTransportModes() { }
...
   Hint: Use the corresponding array method to decide if an element is member of an array.
 */
-function isAccessibleByTransportMode() { }
+function isAccessibleByTransportMode(arraysOfTransport,transportMode) { 
+  if (arraysOfTransport.find(item => item ===transportMode) === transportMode){
+    return true
+  }
+  else{
+    return false
+  }
```
zoltan-gal, line 97: In this case, using `includes` would give more compact code. Or also you can use some, but it doesn't need such an extra logic.

week-3/Homework/mandatory/5-journey-planner.js
```diff
@@ -93,7 +104,11 @@ function isAccessibleByTransportMode() { }
    - Returns the name of the location
      e.g: "Tower Bridge"
 */
-function getLocationName() { }
+function getLocationName(arraysOfLocAndTransport) {
+  let location = arraysOfLocAndTransport.slice(0,1);
+  
+  return location[0];
```
zoltan-gal, line 110: This happens when we learn new staffs and think we should use it in whatsoever places :) and forget that it might not necessary as can be solved simpler without them:
What if we simply read the first element of the incoming array: `arraysOfLocAndTransport[0]`

week-3/Homework/mandatory/5-journey-planner.js
```diff
@@ -120,6 +135,24 @@ function getLocationName() { }
   Advanced challange: try to use arrow function when invoking an array method.
 */
 function journeyPlanner(locations, transportMode) {
+  let isAccessiblePlaces=[];
+  let placesIcanGo=[];
+  let transport = locations . map (getTransportModes);
```
zoltan-gal, line 140: Wonderful! You passed a function as a parameter for map!

week-3/Homework/mandatory/5-journey-planner.js
```diff
@@ -120,6 +135,24 @@ function getLocationName() { }
...
+  
+  for( var i = 0; i < transport.length;i++ ){
+   
+    if (isAccessibleByTransportMode(transport[i],transportMode)) { 
+      isAccessiblePlaces.push(locations[i]);
+     
+    }
+  }
```
zoltan-gal, line 148: Another way to achieve this is to use array methods.
 1) Use `filter` on the `locations` list then inside the filter function you can 
     - get the transport mode of that location: `let transports = getTransportModes(location)
     - then can call isAccessibleByTransportMode(transports,transportMode) with the transport modes above.
 2) Finally, you have to just map over the filtered list of locations, there you can reuse the `getLocationName` function, similarly as you used the `getTransportModes` on map.

Your code would like something like this:
```
locations.filter(function(location) {
 // here your filter logic comes as described in point 1.
}).map(getLocationName);
```

week-3/Homework/mandatory/6-lane-names.js
```diff
@@ -5,8 +5,12 @@
   
 */
 //hint: string and array methods that could be helpful (indexOf, filter)
-function getLanes() {
-
+function getLanes(listOflondonStreet) {
+    let laneStreet = [];
+    laneStreet = listOflondonStreet . filter( street => street.includes ("Lane") );
```
zoltan-gal, line 10: Superb!

zoltan-gal:
Thank you for submitting it! Nice job!
I might be a bit cheeky by saying that some exercises look quite identical to teachers solutions 😄, however, I'd like to make sure that you understand them in any case even if you used them for help.

PR #872, opened 2020-05-19
finished all HW and some of coding wars

elsahib:
# Your Details

Your Name: Osman Elsahib
Your City: Manchester
Your Slack Name: Osman Elsahib

# Homework Details

Module: Js
Week: 4

### DIFF
--- a/week-4/Homework/mandatory/2-writers.js
+++ b/week-4/Homework/mandatory/2-writers.js
@@ -54,7 +54,9 @@ Exercise 1:
   "Hi, my name is {firstName} {lastName}. I am {age} years old, and work as a {occupation}."
 */
 
-
+writers.forEach(function (person) {
+  console.log(`Hi, my name is ${person.firstName} ${person.lastName}. I am ${person.age} years old, and work as a ${person.occupation}.`);
+})
 
 /*
 Exercise 2:
@@ -62,7 +64,14 @@ Exercise 2:
   and not with us anymore, use the below sentence format:
   "Writer {firstName} {lastName} died at {age} years old."
 */
-
+writers.forEach(function(person) {
+  if (person.age <= 49 && person.age >= 40) {
+    if (person.alive == false) {
+      console.log(`Writer ${person.firstName} ${person.lastName} died at ${person.age} years old.`);
+      
+    }
+  }
+})
 
 
 /*
@@ -70,3 +79,11 @@ Exercise 3:
   Only `console.log()` out contemporary (meaning still alive) writers who are in their forties:
   "Hi, my name is {firstName} {lastName}. I am {age} years old."
 */
+writers.forEach(function(person) {
+  if (person.age <= 49 && person.age >= 40) {
+    if (person.alive == true) {
+    console.log(`Hi, my name is ${person.firstName} ${person.lastName}. I am ${person.age} years old.`);
+        
+    }
+  }
+})
--- a/week-4/Homework/mandatory/3-water-bottle.js
+++ b/week-4/Homework/mandatory/3-water-bottle.js
@@ -22,18 +22,33 @@ let bottle = {
   volume: 0,
   fillUp: function() {
     // calling this function should pour your bottle full (volume = 100);
+    this.volume = 100
   },
   pour: function () {
     // calling this function should increase your bottle volume by 10 unit;
+    this.volume += 10
+    if (this.volume > 100) {
+      this.volume = 100      
+    }
   },
   drink: function() {
     // calling this function should decrease your bottle volume by 10 unit;
+    this.volume -= 10
+    if (this.volume < 0) {
+      this.volume = 0
+    }
   },
   isFull: function () {
     // this function should return true if your bottle is empty;
+    if (this.volume == 100) {
+      return true
+    }
   },
   isEmpty: function() {
     // this function should return true if your bottle is full;
+    if (this.volume == 0) {
+      return true
+    }
   }
 };
 
@@ -50,7 +65,7 @@ Extra question:
   Leave your answer below:
 */
 
-// HERE COMES YOUR ANSWER
+// HERE COMES YOUR ANSWER : I don't exactly know why, but I find it clearer and more conveinent to use this rather than the object name when i'm referring to things inside it.
 
 /*
 Once you have completed your object run the following 
--- a/week-4/Homework/mandatory/4-groceries.js
+++ b/week-4/Homework/mandatory/4-groceries.js
@@ -11,7 +11,7 @@ that contains the missing ingredients to your menus. It is stored in the "weekly
 Complete the exercises below.
 */
 
-// Here is your 
+// Here is your    
 let weeklyMealPlan = {
   monday: ["Cheese", "Eggs", "Tomato", "Paprika", "Leek"],
   tuesday: ["Wrap", "Tuna", "Canned beans", "Cheese", "Carrot", "Aubergine"],
@@ -30,6 +30,22 @@ Exercise 1:
 // Gather all week item names into this array
 let weeklyGroceriesToBuy = [];
 
+// weeklyMealPlan.forEach(function (day) {
+//   if (day.length > 0) {
+//     for (let i = 0; i <day.length; i++) {
+//       weeklyGroceriesToBuy.push(day[i])
+//     }
+//   }
+// })
+
+for (day in weeklyMealPlan) {
+  if (weeklyMealPlan[day].length > 0) {   
+    for (let i = 0; i < weeklyMealPlan[day].length; i ++)
+      weeklyGroceriesToBuy.push(weeklyMealPlan[day][i])   
+  }
+}
+console.log(weeklyGroceriesToBuy);
+
 
 
 /*
@@ -39,7 +55,13 @@ Exercise 2:
 */
 // Gather weekend item names into this array
 let weekendGroceriesToBuy = [];
-
+for (day in weeklyMealPlan) {
+  if (day == "saturday" || day == "sunday" && weeklyMealPlan[day].length > 0 ) {   
+    for (let i = 0; i < weeklyMealPlan[day].length; i ++)
+      weekendGroceriesToBuy.push(weeklyMealPlan[day][i])   
+  }
+}
+console.log(weekendGroceriesToBuy);
 
 
 /*
@@ -60,3 +82,13 @@ let numberOfItemsPerWeak = {
   saturday: 0,
   sunday: 0
 };
+
+for (count in numberOfItemsPerWeak) {
+  for (day in weeklyMealPlan){
+    if (count == day) {
+      numberOfItemsPerWeak[count] = weeklyMealPlan[day].length
+    } 
+  }
+}
+
+console.log(numberOfItemsPerWeak);
--- a/week-4/Homework/mandatory/5-codewars.md
+++ b/week-4/Homework/mandatory/5-codewars.md
@@ -6,8 +6,8 @@ Click "ATTEMPT" to test your solution.
 
 Exercises:
 
-- [Training JS #5: Basic data types--Object](https://www.codewars.com/kata/571f1eb77e8954a812000837/train/javascript)
-- [Welcome!](https://www.codewars.com/kata/welcome/train/javascript)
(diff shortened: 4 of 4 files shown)


week-4/Homework/mandatory/3-water-bottle.js
```diff
@@ -22,18 +22,33 @@ let bottle = {
...
+    this.volume -= 10
+    if (this.volume < 0) {
+      this.volume = 0
+    }
   },
   isFull: function () {
     // this function should return true if your bottle is empty;
+    if (this.volume == 100) {
+      return true
+    }
```
zoltan-gal, line 45: Well, although it was explicitly declared, it should return false if not full.
It'd also keep the function outcome more consistent. Because now it can return either `true` or `undefined` instead of `false`.
More compact solution is simply saying `return this.volume == 100`

week-4/Homework/mandatory/3-water-bottle.js
```diff
@@ -22,18 +22,33 @@ let bottle = {
...
   },
   isEmpty: function() {
     // this function should return true if your bottle is full;
+    if (this.volume == 0) {
+      return true
+    }
```
zoltan-gal, line 51: Likewise than above, just the other way around.

week-4/Homework/mandatory/3-water-bottle.js
```diff
@@ -50,7 +65,7 @@ Extra question:
   Leave your answer below:
 */
 
-// HERE COMES YOUR ANSWER
+// HERE COMES YOUR ANSWER : I don't exactly know why, but I find it clearer and more conveinent to use this rather than the object name when i'm referring to things inside it.
```
zoltan-gal, line 68: One reason is that by using `this` keeps the code more enclosed/encapsulated, there is no outside dependency. Because if it refers to itself, `this` makes it more resilient, also it is more pragmatic solution. E.g. if you rename the variable name then you should also rename the corresponding variable name inside. However, by using `this` you don't need to.

week-4/Homework/mandatory/4-groceries.js
```diff
@@ -30,6 +30,22 @@ Exercise 1:
...
+// weeklyMealPlan.forEach(function (day) {
+//   if (day.length > 0) {
+//     for (let i = 0; i <day.length; i++) {
+//       weeklyGroceriesToBuy.push(day[i])
+//     }
+//   }
+// })
+
+for (day in weeklyMealPlan) {
+  if (weeklyMealPlan[day].length > 0) {
```
zoltan-gal, line 42: Don't need to check if array is empty twice. You have the same check implemented in the condition inside the for-loop.

week-4/Homework/mandatory/4-groceries.js
```diff
@@ -39,7 +55,13 @@ Exercise 2:
 */
 // Gather weekend item names into this array
 let weekendGroceriesToBuy = [];
-
+for (day in weeklyMealPlan) {
+  if (day == "saturday" || day == "sunday" && weeklyMealPlan[day].length > 0 ) {
```
zoltan-gal, line 59: This can be both ambiguous and err-prone using `or` and `and` operators on the same level without making the order of precedence explicit. This is something we have yet to teach you all, it is called precedence that determines the evaluation order of expressions.
See more: https://www.dummies.com/web-design-development/javascript-operator-precedence/ 

Your expression is equivalent to this form:
`day == "saturday" || (day == "sunday" && weeklyMealPlan[day].length > 0)`
It means if not `saturday` then condition will be only `true` if day is `sunday and meal plan is empty for that day.
Which is luckily true 😃 in this case. But what if the array weren't empty?

week-4/Homework/mandatory/4-groceries.js
```diff
@@ -60,3 +82,13 @@ let numberOfItemsPerWeak = {
   saturday: 0,
   sunday: 0
 };
+
+for (count in numberOfItemsPerWeak) {
+  for (day in weeklyMealPlan){
+    if (count == day) {
```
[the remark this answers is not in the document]
elsahib20: To ensure that I’m assigning the right count to the right day. As I’m only assigning values to the existing Object not creating a an array to return the count

week-4/Homework/mandatory/2-writers.js
```diff
@@ -54,19 +54,36 @@ Exercise 1:
...
   and not with us anymore, use the below sentence format:
   "Writer {firstName} {lastName} died at {age} years old."
 */
-
+writers.forEach(function(person) {
+  if (person.age <= 49 && person.age >= 40) {
+    if (person.alive == false) {
+      console.log(`Writer ${person.firstName} ${person.lastName} died at ${person.age} years old.`);
+      
+    }
```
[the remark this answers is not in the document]
elsahib20: Will I need to use parentheses as well if I’m using the unary ! Operator?

If !person.alive OR if (!person.alive) OR if !(person.alive) 

Which way is the right way to use it?

week-4/Homework/mandatory/3-water-bottle.js
elsahib20: There is a mistake in line 42, this function is for when the bottle is full not empty.
Its still stuck in my head that I should avoid jumping into the return statement straight away. Someone planted the idea in my head that this will make it harder to maintain the code.
But I like it though

week-4/Homework/mandatory/4-groceries.js
elsahib20: I didn’t understand what you mean here.

week-4/Homework/mandatory/4-groceries.js
elsahib20: Does this make it better ?
if ((day == "saturday" || day == "sunday") && weeklyMealPlan[day].length > 0 )

week-4/Homework/mandatory/4-groceries.js
zoltan-gal: So technically these are the same if you think about that
`if (weeklyMealPlan[day].length > 0)` === `for (let i = 0; i < weeklyMealPlan[day].length; i++)`

If the array is empty then the for-loop will do nothing, it is redundant to check in the if statement too.

week-4/Homework/mandatory/4-groceries.js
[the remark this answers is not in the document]
zoltan-gal: Ahh, all right, I see that. The naming is not improper though. `count` will be the keys of the `numberOfItemsPerWeak` object which will be a string, however, the name `count` assumes that it's going to be some kind of number.
I would rather avoid having nested for loop. The key of the outer for loop (`numberOfItemsPerWeak`) could be immediately used to determine whether the same key exists in the object (`weeklyMealPlan`) used in the nested loop.

Also if you think about the performance, more keys the object has, more comparison you will do needlessly.
The number of comparisons will be `N^2` where `N` is the number of keys. 

This something we describe with the big O notation.
> In computer science, big O notation is used to classify algorithms according to how their run time or space requirements grow as the input size grows.[3]
https://en.wikipedia.org/wiki/Big_O_notation

week-4/Homework/mandatory/4-groceries.js
zoltan-gal: Yes, it is definitely better. Less the chance to fall into a logical pitfall. Parenthesis make the evaluation order of the logical expression unequivocal.

week-4/Homework/mandatory/3-water-bottle.js
zoltan-gal: Yes, it was a copy-paste type.
I think the idea someone wanted to plant was going to refer to something else. There is an argument that a function should have only one return statement instead of multiple. But I'm sceptical about the benefit of that approach. I don't see much harm of having multiple exit point in a function. Well, if the function is too big it might make harder to maintain as someone told you.

week-4/Homework/mandatory/2-writers.js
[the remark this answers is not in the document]
zoltan-gal: Yes, you must use parenthesis on if statement first because it is part of the syntax then everything else should go inside, otherwise, you would get a syntax error. So the middle example is the correct.

week-4/Homework/mandatory/4-groceries.js
elsahib: BODMAS you mean?

week-4/Homework/mandatory/2-writers.js
```diff
@@ -54,19 +54,36 @@ Exercise 1:
   "Hi, my name is {firstName} {lastName}. I am {age} years old, and work as a {occupation}."
 */
 
-
+writers.forEach(function (person) {
+  console.log(`Hi, my name is ${person.firstName} ${person.lastName}. I am ${person.age} years old, and work as a ${person.occupation}.`);
+})
```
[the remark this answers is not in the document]
elsahib: Implemented

week-4/Homework/mandatory/2-writers.js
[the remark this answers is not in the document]
elsahib: Thanks Zoltan.

week-4/Homework/mandatory/2-writers.js
```diff
@@ -54,19 +54,36 @@ Exercise 1:
...
 /*
 Exercise 2:
   Only `console.log()` out the writers who are in their 40s (meaning between 40 and 49)
   and not with us anymore, use the below sentence format:
   "Writer {firstName} {lastName} died at {age} years old."
 */
-
+writers.forEach(function(person) {
+  if (person.age <= 49 && person.age >= 40) {
+    if (person.alive == false) {
```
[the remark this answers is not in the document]
elsahib: '''
const aliveInForties = writers.filter((writer) => {
  return writer.age <= 49 && writer.age >= 40;
});
aliveInForties.forEach((person) => {
  if (person.alive == true) {
    console.log(
      `Hi, my name is ${person.firstName} ${person.lastName}. I am ${person.age} years old.`
    );
  }
});
'''
Is this what you mean?

week-4/Homework/mandatory/3-water-bottle.js
elsahib: So, this is more of a coding preference, it's not a case of right and wrong?

week-4/Homework/mandatory/4-groceries.js
elsahib: fixed it. I think I'm tip-toing when writing code, I hate to get an error message.

week-4/Homework/mandatory/4-groceries.js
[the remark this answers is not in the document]
elsahib: This was very helpful. Thank you loads. here is the re-write of the above code.
```
for (day in numberOfItemsPerWeak) {
  numberOfItemsPerWeak[day] = weeklyMealPlan[day].length;
}
```

week-4/Homework/mandatory/2-writers.js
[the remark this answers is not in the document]
zoltan-gal: yes, it is

week-4/Homework/mandatory/4-groceries.js
zoltan-gal: Never heard this acronym :) Thanks for that!

week-4/Homework/mandatory/2-writers.js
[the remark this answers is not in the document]
zoltan-gal: But also this part `if (person.alive == true)` should go to the filter :)

week-4/Homework/mandatory/3-water-bottle.js
zoltan-gal: Yes, I believe so :). This kind of topic is subjective. I'd say it can be religious :D.
Here you can see an 11 years old question and loads of answers along with examples. It is worth reading. I'm glad that you raised this topic. I never really looked it deeply.
https://stackoverflow.com/questions/36707/should-a-function-have-only-one-return-statement

I'm coping some good points from there:

> Minimize the number of returns in each routine. It's harder to understand a routine if, reading it at the bottom, you're unaware of the possibility that it returned somewhere above.

> Use a return when it enhances readability. In certain routines, once you know the answer, you want to return it to the calling routine immediately. If the routine is defined in such a way that it doesn't require any cleanup, not returning immediately means that you have to write more code.

PR #884, opened 2020-05-25
JS-Hedyeh- week4

Hedyeh65:
# Your Details

Your Name: Hedyeh
Your City:Manchester
Your Slack Name:

# Homework Details

Module:JavaScript
Week:4

### DIFF
--- a/week-4/Homework/mandatory/2-writers.js
+++ b/week-4/Homework/mandatory/2-writers.js
@@ -22,29 +22,29 @@ let writers = [
     lastName: "Woolf",
     occupation: "writer",
     age: 59,
-    alive: false
+    alive: false,
   },
   {
     firstName: "Zadie",
     lastName: "Smith",
     occupation: "writer",
     age: 41,
-    alive: true
+    alive: true,
   },
   {
     firstName: "Jane",
     lastName: "Austen",
     occupation: "writer",
     age: 41,
-    alive: false
+    alive: false,
   },
   {
     firstName: "bell",
     lastName: "hooks",
     occupation: "writer",
     age: 64,
-    alive: true
-  }
+    alive: true,
+  },
 ];
 
 /*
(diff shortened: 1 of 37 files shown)


week-4/Homework/mandatory/2-writers.js
```diff
@@ -53,19 +53,41 @@ Exercise 1:
...
+  if (
+    writers[i].age >= 40 &&
+    writers[i].age < 49 &&
+    writers[i].alive === true
+  ) {
+    console.log(
+      `Hi,my name is ${writers[i].firstName} ${writers[i].lastName}. I am ${writers[i].age} years old.`
+    );
+  }
+}
```
janetrimmer, line 93: This is correct but an alternative is to use the filter method:
writers.filter(writer => !writer.alive && writer.age >= 40 && writer.age <= 49).forEach(writer => {
  console.log(`Writer: ${writer.firstName} ${writer.lastName} died at ${writer.age} years old.`);
});

week-4/Homework/mandatory/3-water-bottle.js
```diff
@@ -20,21 +20,37 @@ You have to implement the missing features according to the specification.
...
   volume: 0,
-  fillUp: function() {
+  fillUp: function () {
+    this.volume = 100;
     // calling this function should pour your bottle full (volume = 100);
   },
   pour: function () {
+    if (this.volume < 100) {
+      return (this.volume += 10);
+    }
```
janetrimmer, line 30: This isn't quite right. If the volume were 95, for example, then adding 10 would increase the volume to 105.

week-4/Homework/mandatory/3-water-bottle.js
```diff
@@ -20,21 +20,37 @@ You have to implement the missing features according to the specification.
...
     // calling this function should decrease your bottle volume by 10 unit;
   },
   isFull: function () {
+    if (this.volume == 100) return true;
+    else {
+      return false;
+    }
+    // this function should return true if your bottle is full;
+  },
```
janetrimmer, line 45: To be absolutely accurate,
 if (this.volume == 100) should be replaced by
 if (this.volume === 100)
A slightly neater way of writing this is
 return this.volume === 100 ? true : false;

PR #908, opened 2020-06-03
Finished week 5 work

elsahib:
# Your Details

Your Name: Osman Elsahib
Your City: Manchester
Your Slack Name: Osman Elsahib

# Homework Details

Module: Java Script
Week: 5

### DIFF
--- a/week-5/Homework/mandatory/2-exercises/exercises.js
+++ b/week-5/Homework/mandatory/2-exercises/exercises.js
@@ -40,6 +40,15 @@
 function insertPeopleData(arrayOfPeople) {
   let content = document.querySelector("#content");
   //Write your code in here
+  
+  for (let i = 0; i < arrayOfPeople.length ; i ++) {
+    let name  = document.createElement("h1");
+    let job = document.createElement("h2");
+    name.innerText = arrayOfPeople[i].name;
+    job.innerText = arrayOfPeople[i].job;
+    content.append(name);
+    content.append(job);
+  }
 }
 
 /**
@@ -53,7 +62,16 @@ function insertPeopleData(arrayOfPeople) {
  */
 function insertShoppingList(shoppingList) {
   //Write your code in here
+  let content = document.querySelector("#content");
+  let myList = document.createElement("ul");
+  content.appendChild(myList)
+  for (let i = 0; i < shoppingList.length ; i ++) {
+    let tempLi = document.createElement("li")
+    tempLi.innerText = shoppingList[i];
+    myList.append(tempLi)
+  }
 }
+
 
 /**
  * I'd like to display my three favorite books inside a nice webpage!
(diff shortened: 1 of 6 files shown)


week-5/Homework/mandatory/2-exercises/exercises.js
```diff
@@ -73,6 +91,29 @@ function insertShoppingList(shoppingList) {
  */
 function insertBooks(books) {
   //Write your code in here
+  let images = ["https://encrypted-tbn2.gstatic.com/shopping?q=tbn:ANd9GcTcej0xfvRf7aB-hIonqVwHAgrxSOyUyEAt_z9w1jKmMPudjZG5ZIqZKXqxhmgoBVmmfk3e3b_e4gPkVfKYmn4bAZj6fy9cfL8Rsv7dgi7-&usqp=CAY",
+  "https://images-na.ssl-images-amazon.com/images/I/41m1rQjm5tL._SX322_BO1,204,203,200_.jpg",
+  "https://imagery.pragprog.com/products/595/tpp20.jpg?1557173963"]
```
zoltan-gal, line 96: This is a neat solution 👍

week-5/Homework/mandatory/2-exercises/style.css
```diff
@@ -10,7 +10,20 @@ html,
...
+    padding: 20px;
+    width: calc( 100% - 41px);
+}
 
+.li {
+    width: calc(25% - 51px);
+    margin: 15px;
+    padding: 10px;
+    min-width: 350px;
+}
```
zoltan-gal, line 26: Great to see that you cared about the appearance too 🌟

week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,54 @@
+// ############### Part One #################
+// ========= Elements to be manipulated =========
+let jumptron  = document.querySelector(".jumbotron");
+let blueBtn = document.querySelector("#blueBtn");
+let orangeBtn = document.querySelector("#orangeBtn");
+let greenBtn = document.querySelector("#greenBtn");
+let dntBtn = document.getElementsByClassName("btn-lrg")[0];
+let vlnBtn = document.getElementsByClassName("btn-lrg")[1];
```
zoltan-gal, line 8: This is fine. General advice, if we can, use more precise selectors to elements you are looking for e.g. here ".jumbotron .btn-primary". That would make your code more resilient/flexible against modification in the HTML page. Changing of the class of an element is less likely than changes of in their text, but the order can change too and in that case, your code would be inverted.

So all I mean it better selector you have, less likely to change the code in case of HTML changes. But it doesn't mean that you have to write something super long very precise selector 😄  that could be also super fragile.

week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,54 @@
...
+            email.style.border = "2px solid red";
+        }else if (name.value.length <= 0 ) {
+            name.style.border = "2px solid red";
+        } else if (textarea.value.length <= 0) {
+            textarea.style.border = "2px solid red";
+            textarea.placeholder = "Enter some Text";
+        } else {
+            alert ("Thank you for filling out this form")
+            document.getElementsByTagName("form")[0].reset();
+        }
```
zoltan-gal, line 52: It is better than an if-else spiral but there is still room for improvements here :)

zoltan-gal: Although it doesn't require such an extra cognitive effort to reason about and read, I'd suggest keeping logics separated. Here we can split it up to two or three main sections to simplify and improve.

For instance:
- store the validation result of each field in an interim variable that can be reused to do the following steps
- set the background colours for each
- then if each of them valid perform the action

zoltan-gal: Some cosmetic issue:
- When I click the submit button, this only paints the colour to red for the first invalid box, the rest of them remain white even if they are empty. It'd be nice to see all errors in one go, elevated user experience :)
- When everything is valid and the form has been cleared, backgrounds are not changed back to white.

PR #921, opened 2020-06-05
Hedyeh week 5

Hedyeh65:
# Your Details

Your Name:
Your City:
Your Slack Name:

# Homework Details

Module:
Week:

### DIFF
--- a/week-5/Homework/mandatory/2-exercises/exercises.js
+++ b/week-5/Homework/mandatory/2-exercises/exercises.js
@@ -1,65 +1,111 @@
+/*
+ In this exercise you will deal with Objects stored in array and DOM manipulation with JavaScript.
+ From now, exercise descriptions don't provide detailed examples about input and output of functions.
+ You should be already able to read and understand the code in order to extract the necessary information
+ you need to implement functions' body.
+ */
+
+/*
+ Optional question:
+
+ Notice how this JavaScript file linked to the index.html file.
+ What do you think when this JavaScript file is being executed?
+
+ When you finished this exercise try to move the script tag around the Div tag with the id "content",
+ refresh the page and observe what happens. (This should answer the question above.)
+ 
+ What do you think why this changes shouldn't work?
+ */
+// LEAVE YOUR ANSWER HERE (THIS IS OPTIONAL)
+
 /**
+ * This function recieves a lists of people. Each object should contain the name and the occupation of a person.
+ * Look for usage of the function in the code and see what variable is passed into it as an argument.
  *
- * For each of the names in the array passed into this function
+ * Create and insert the below HTML elements to the DOM for each of Objects in the Array as follows:
+ *  1. Add a <h1> tag to the website containing the name of the person
+ *  2. Add a <h2> tag to the website containing the job of the person
  *
- * - Add a <h1> tag to the website containing the name of the person
- * - Add a <h2> tag to the website containing the job of the person
+ * All of your HTML elements should go inside the Div tag with the id "content".
  *
- * All of your HTML should go inside the Div tag with the id "content".
- *
+ * An example "content" div should look like:
  * <div id="content">
- *      <h1>{Name Here}</h1>
- *      <h2>{Job Here}</h2>
- *      .....
+ *      <h1>Mario</h1>
+ *      <h2>Plumber</h2>
+ *      <h1>Luigi</h1>
+ *      <h2>Plumber</h2>
  * </div>
  */
-function exerciseOne(arrayOfPeople) {
+function insertPeopleData(arrayOfPeople) {
   let content = document.querySelector("#content");
+  for (let i = 0; i < arrayOfPeople.length; i++) {
+    var h1El = document.createElement("h1");
+    h1El.textContent = arrayOfPeople[i].name;
+    content.appendChild(h1El);
+    var h2El = document.createElement("h2");
+    h2El.textContent = arrayOfPeople[i].job;
+    content.appendChild(h2El);
+  }
 }
 
 /**
  *
- * Create a list of shopping items. You should use an unordered list.
+ * Create a list of shopping items using an unordered HTML list.
+ * The input of this function is a simple Array of Strings, look for the concrete example in the code.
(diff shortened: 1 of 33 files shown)


week-5/Homework/mandatory/2-exercises/exercises.js
```diff
@@ -1,64 +1,119 @@
...
-function exerciseOne(arrayOfPeople) {
+function insertPeopleData(arrayOfPeople) {
   let content = document.querySelector("#content");
+  for (let i = 0; i < arrayOfPeople.length; i++) {
+    var h1El = document.createElement("h1");
+    h1El.textContent = arrayOfPeople[i].name;
+    content.appendChild(h1El);
+    var h2El = document.createElement("h2");
+    h2El.textContent = arrayOfPeople[i].job;
+    content.appendChild(h2El);
```
dsobkow, line 47: Great work here Hedyeh👏

week-5/Homework/mandatory/2-exercises/exercises.js
```diff
@@ -1,64 +1,119 @@
...
-function exerciseTwo(shopping) {
-  //Write your code in here
+function insertShoppingList(shoppingList) {
+  let content = document.querySelector("#content");
+  let listToBuy = document.createElement("ul");
+  for (let i = 0; i < shoppingList.length; i++) {
+    let list = document.createElement("li");
+    list.textContent = shoppingList[i];
+    listToBuy.appendChild(list);
+    content.appendChild(listToBuy);
```
dsobkow, line 67: Well done! I can see you used a for loop but remember that higher order functions such as `forEach` are your friend here, try to use them when possible🙂 Also appending `ul` element (on line 67) to the content should be outside the loop

week-5/Homework/mandatory/2-exercises/exercises.js
```diff
@@ -1,64 +1,119 @@
...
-function exerciseThree(books) {
-  //Write your code in here
+    if (books[i].alreadyRead === true) {
+      liOfBook.style.backgroundColor = "green";
+    } else {
+      liOfBook.style.backgroundColor = "red";
+    }
+  }
+  unorderedList.style.display = "flex";
+  unorderedList.style.justifyContent = "space-around";
```
dsobkow, line 116: Really nice solution and styling is almost right. I would only add some padding to `li` elements and remove list bullet points by adding this line of code:
``` unorderedList.style.listStyleType = "none";```
You could also declare an array with image URLs outside the loop. Images don't have links(anchor tags) to URLs with book covers as per the instruction but you did a good job🙂

File: week-5/Homework/mandatory/3-project/README.md [hunk 4]
```diff
@@ -9,18 +9,18 @@ Clicking on the buttons should change the "theme" of the website:
 
 - When clicking **blue** it should change:
 
-  - **Jumbotron** background color to `#588fbd`
+  - Div tag with the id **Jumbotron** background color to `#588fbd`
   - **Donate a bike** button background color to `#ffa500`
   - **Volunteer** button background color to `black` and text color to `white`
```
dsobkow, line 14: Your solution to this exercise is missing. It should be added to `main.js` file in `3-project/js` folder

week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,54 @@
...
+orangeBtn.addEventListener("click", function () {
+  jumbotron.style.backgroundColor = `#f0ad4e`;
+  donateBtn.style.backgroundColor = `#5751fd`;
+  volunteerBtn.style.backgroundColor = `#31b0d5`;
+  volunteerBtn.style.color = `white`;
+});
+greenBtn.addEventListener("click", function () {
+  jumbotron.style.backgroundColor = `#87ca8a`;
+  donateBtn.style.backgroundColor = `black`;
+  volunteerBtn.style.backgroundColor = `#8c9c08`;
```
dsobkow, line 24: The purpose of this exercise was to query specific DOM elements without modifying anything in html file. You added `id` attributes to a few buttons which made this exercise slightly easier. Beside that, well done! All buttons work correctly🙂 However I can see some logic repetition here as all three event listeners make very similar changes. Can you think of the way how you could refactor the code so there is only one function responsible for making these changes? This is only something for you to think about, I'm not asking you to submit a refactored solution as this one is good enough🙂

week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,54 @@
...
+    yourName.value = "";
+    describe.value = "";
+  } else if (
+    emailAddress.value.length <= 0 ||
+    yourName.value.length <= 0 ||
+    describe.value.length <= 0
+  ) {
+    emailAddress.style.backgroundColor = `red`;
+    yourName.style.backgroundColor = `red`;
+    describe.style.backgroundColor = `red`;
```
dsobkow, line 52: You were very close but something is not right here:
- Background color should change to red only for invalid fields. At the moment it changes for all fields, even the ones that are valid.
- If I input some characters to email field but without `@` character, the background doesn't change to red.
-  When I try to submit the form with empty fields, their background color changes to red. Then I input some values and click submit again. I can see the alert but the background color remains red when it should change back to white.

week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,54 @@
...
+  volunteerBtn.style.backgroundColor = `#8c9c08`;
+});
+
+var emailAddress = document.getElementById("exampleInputEmail1");
+var yourName = document.getElementById("example-text-input");
+var describe = document.getElementById("exampleTextarea");
+var submitBtn = document.getElementById("submitBtn");
+
+submitBtn.addEventListener("click", function (submit) {
+  submit.preventDefault();
```
dsobkow, line 33: There is no need to pass a `submit` argument to the function here. You need to call `preventDefault()` for the `event` like this:
```
submitBtn.addEventListener("click", function () {
  event.preventDefault();
...
```

PR #924, opened 2020-06-05
Vito week 5

JeanPaul18:
# Your Details

Your Name:Vito
Your City:Manchester
Your Slack Name:Vito

# Homework Details

Module:Js2
Week:2

### DIFF
--- a/week-5/Homework/mandatory/2-exercises/exercises.js
+++ b/week-5/Homework/mandatory/2-exercises/exercises.js
@@ -18,17 +18,16 @@
  */
 // LEAVE YOUR ANSWER HERE (THIS IS OPTIONAL)
 
-
-/**
+/**git
  * This function recieves a lists of people. Each object should contain the name and the occupation of a person.
  * Look for usage of the function in the code and see what variable is passed into it as an argument.
- * 
+ *
  * Create and insert the below HTML elements to the DOM for each of Objects in the Array as follows:
  *  1. Add a <h1> tag to the website containing the name of the person
  *  2. Add a <h2> tag to the website containing the job of the person
  *
  * All of your HTML elements should go inside the Div tag with the id "content".
- * 
+ *
  * An example "content" div should look like:
  * <div id="content">
  *      <h1>Mario</h1>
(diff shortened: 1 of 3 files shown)


week-5/Homework/mandatory/2-exercises/exercises.js
```diff
@@ -37,42 +36,94 @@
...
+    liEl.style.listStyleType = "none";
+    liEl.style.margin = "20";
+    if (books[i].alreadyRead === true) {
+      liEl.style.backgroundColor = "green";
+    } else {
+      liEl.style.backgroundColor = "red";
+    }
+  }
+  ulEl.style.display = "flex";
+  info.appendChild(ulEl);
```
dsobkow, line 126: Very nice solution🏅 Is it a good practice to be consistent when writing code so I wouldn't mix `let` and `var`. You could also add some padding to `li` elements and make sure that images have the same width and height. 

Images don't have links to URLs with book covers as per the instruction but you did a good job🙂

week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,65 @@
...
+  document.querySelector(".jumbotron .btn-secondary").style.color = "white";
+});
+
+let greenChanges = document.querySelector("#greenBtn");
+greenChanges.addEventListener("click", function () {
+  document.querySelector(".jumbotron").style.backgroundColor = "#87ca8a";
+  document.querySelector(".jumbotron .btn").style.backgroundColor = "black";
+  document.querySelector(".jumbotron .btn-secondary").style.backgroundColor =
+    "#8c9c08";
+});
```
dsobkow, line 25: Nice one, it works correctly! All three event listeners make very similar changes, therefore there is some repetition in your code. Can you think of the way how you could refactor the code so there is only one function responsible for making these changes? Hint: you can pass colors as function parameters.

week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,65 @@
...
+    email.value.length > 0 &&
+    email.value.includes("@") &&
+    name.value.length > 0 &&
+    text.value.length > 0
+  ) {
+    alert("Thank you for filling out this form");
+    email.value = "";
+    name.value = "";
+    text.value = "";
+  }
```
dsobkow, line 64: Awesome work! The behaviour of the form is correct but there is always room for improvement🙂 

There is an easier way to select the button we need here. You can find a button with a type of submit like this:
```let submitButton = document.querySelector("button[type='submit']");```

I can also see several if else statements checking the same condition, and then you check them again from line 54. It could be extracted to a separate function responsible for the validation. Remember, keep your code [DRY (Don't Repeat Yourself)](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) - you will hear it a lot😀

PR #929, opened 2020-06-05
Week4 homework

mara-ber:
# Your Details

Your Name: Marina
Your City: Glasgow
Your Slack Name:

# Homework Details

Module:
Week:


week-4/Homework/mandatory/1-food.js
```diff
@@ -26,22 +27,29 @@ console.log("I can buy my favorite food from " + getTakeAwayAddress(myFavFood));
...
+
+console.log("Is the food low calorie? " + isLowCalorie(myFavFood));
 //---------------------------------------
 
 // 3. Declare a function "isSafeForNutAllergies" that accepts "food" as a parameter
 // and returns false if it found the word "sesame" in the ingredients, or true otherwise
 
 function isSafeForNutAllergies(food) {
+  return !food.ingredients.includes("sesame seeds");
   //write code here
```
alastair87, line 53: You need to check for the word "sesame" in the ingredients so it's a bit more than checking if the `food.ingredients` array includes sesame seeds. You need to use an array function that will iterate (loop) through the ingredients and then check whether each string contains that function. I would suggest that you will need to use an array function, a function of your own that you use on each element and then a function of string to check what each one includes. From this you should be able to work out whether it's true or not that there is at least one ingredient with the word sesame in it.

week-4/Homework/mandatory/3-translator.js
```diff
@@ -28,9 +28,28 @@ It should default to English if the language is not in your object of languages,
...
+    return "that's not a language";
+  }
+}  
+  
+
+console.log(greet("irish"));
+console.log(greet("FINNish"));
+console.log(greet("russian"));
+console.log(greet(6));
+return languages[language];
```
alastair87, line 52: For testing you should call your function from outside - as it is this code will never execute since your `if` block will have already returned. However if it did there would be a problem as your code would recurse infinitely, as greet got called over and over again by itself, eventually causing node to crash.

week-4/Homework/mandatory/3-translator.js
```diff
@@ -28,9 +28,28 @@ It should default to English if the language is not in your object of languages,
...
 }
```
alastair87, line 53: This return outside the function can be removed as it doesn't do anything.

mara-ber: that's not what I have in my file. there are no lines 32,33 and 52,53. Why are they shown here?

PR #930, opened 2020-06-05
Mohammed week 5

mhallaq:
# Your Details

Your Name:Mohammed
Your City:Manchester
Your Slack Name:malhallaq

# Homework Details

Module:
Week:


week-5/Homework/mandatory/2-exercises/exercises.js
```diff
@@ -1,63 +1,77 @@
...
- *      .....
+ *      <h1>Mario</h1>
+ *      <h2>Plumber</h2>
+ *      <h1>Luigi</h1>
+ *      <h2>Plumber</h2>
  * </div>
  */
-function exerciseOne(arrayOfPeople) {
+function insertPeopleData(arrayOfPeople) {
   let content = document.querySelector("#content");
```
dsobkow, line 41: Hi Mohammed, you haven't added your solutions to exercises in this file

week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,84 @@
...
+  //   alert("Blue Button Clicked");
+  var myDiv = document.querySelector(".Jumbotron");
+  myDiv.style.backgroundColor = "#87ca8a";
+
+  var myBtn1 = document.querySelector(".btn.btn-primary.btn-lrg");
+  myBtn1.style.backgroundColor = "black";
+  //
+
+  var myBtn2 = document.querySelector(".btn.btn-secondary.btn-lrg");
+  myBtn2.style.backgroundColor = "#8c9c08";
```
dsobkow, line 49: Nice work, all buttons work correctly! However I can see some logic repetition here. You could select `myBtn1` and `myBtn2` just once outside your functions. Also all three functions make very similar changes. Can you think of the way how you could refactor the code so there is only one function responsible for making these changes?

week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,84 @@
...
+    descField.style.backgroundColor = "white";
+  } else {
+    if (emailField.value.length === 0 && emailField.value.indexOf("@") < 0) {
+      emailField.style.backgroundColor = "red";
+    }
+    if (nameField.value.length === 0) {
+      nameField.style.backgroundColor = "red";
+    }
+    if (descField.value.length === 0) {
+      descField.style.backgroundColor = "red";
```
dsobkow, line 81: You were very close here but something is not right. At the moment, when fields are empty and I click on the submit button, the page will refresh and red fields will disappear. You need to add `event.preventDefault()` in your `submitInfo` function to stop this behaviour. Another thing is that `addEventListener` method needs only two arguments - event type and a callback function so there is no need to pass a boolean as a third argument. I can also see several if else statements checking if fields are empty. It could be extracted to a single function responsible for validation. Remember, keep your code [DRY (Don't Repeat Yourself)](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) - you will hear it a lot🙂

And just one more minor thing - there is an easier way to select the button we need here. You can find a button with a type of submit like this:
```let submitBtn = document.querySelector("button[type='submit']");```

PR #934, opened 2020-06-06
week-5-done

shahdalhaj:
# Your Details

Your Name:
Your City:
Your Slack Name:

# Homework Details

Module:
Week:

### DIFF
--- a/week-5/Homework/mandatory/2-exercises/exercises.js
+++ b/week-5/Homework/mandatory/2-exercises/exercises.js
@@ -37,8 +37,16 @@
  *      <h2>Plumber</h2>
  * </div>
  */
-function insertPeopleData(arrayOfPeople) {
+function insertPeopleData(People) {
   let content = document.querySelector("#content");
+  for (let i = 0; i < people.length; i++) {
+    let name = document.createElement('h1');
+    let job = document.createElement('h2');
+    content.appendChild(name).textContent = people[i].name;
+    content.appendChild(job).textContent = people[i].job;
+    console.log(content)
+    
+  }
   //Write your code in here
 }
 
@@ -51,7 +59,15 @@ function insertPeopleData(arrayOfPeople) {
  *
  * Hint for type of lists in HTML: https://www.w3schools.com/html/html_lists.asp
  */
-function insertShoppingList(shoppingList) {
+function insertShoppingList(shopping) {
+  let container = document.querySelector('#content');
+  let shoppingList = document.createElement('ul');
+  container.appendChild(shoppingList);
+  shopping.forEach(h => {
+    let items = document.createElement('li');
+     items.innerHTML = h;
+     shoppingList.appendChild(items)
+  });
   //Write your code in here
 }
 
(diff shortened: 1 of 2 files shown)


week-5/Homework/mandatory/2-exercises/exercises.js
```diff
@@ -71,9 +87,56 @@ function insertShoppingList(shoppingList) {
...
+    myBooks.appendChild(para);
+    myBooks.appendChild(image);
+    para.innerText = books[i].title + " - " + books[i].author;
+    booksList.append(myBooks)
+    booksList.style.display = "flex";
+    booksList.style.margin = "30px";
+    //booksList.style.justifyContent = "space-between"
+    if (books[i].alreadyRead === true) {
+      myBooks.style.backgroundColor = "green"
+    } else { myBooks.style.backgroundColor = "red"}
```
dsobkow, line 135: Well done here, really nice DOM manipulation but styling could be improved. It looks like the margin should be applied to `li` elements, not to `ul`. You could also do this to get rid of list bullet points:
``` booksList.style.listStyleType = "none";```

It would be nice to give some padding to `li` elements and swap image height and width values so height is `260px` and width `200px`. There is no need to assign className to `ul` and `li` elements if you don’t use them. Images don't have links(anchor tags) to URLs with book covers as per the instruction but you did a good job🙂

File: week-5/Homework/mandatory/3-project/js/main.js [hunk 2]
```diff
@@ -0,0 +1,62 @@
...
+    jumptron.style.backgroundColor = "#f0ad4e";
+    dntBtn.style.backgroundColor = "#5751fd";
+    vlnBtn.style.backgroundColor = "#31b0d5";
+    vlnBtn.style.color = "white";
+})
+
+greenBtn.addEventListener("click", function () {
+    jumptron.style.backgroundColor = "#87ca8a";
+    dntBtn.style.backgroundColor = "black";
+    vlnBtn.style.backgroundColor = "#8c9c08";
```
dsobkow, line 27: Awesome work! There is an easier way to select both buttons:
```
let dntBtn = document.querySelector(".buttons .btn-primary");
let vlnBtn = document.querySelector(".btn-secondary");
```
It’s more error-proof as the order of array elements returned from `getElementsByClassName` method can change when you add more buttons to the website.
Another thing is that all three event listeners make very similar changes, therefore there is some repetition in your code. Can you think of the way how you could refactor the code so there is only one function responsible for making these changes?

week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,62 @@
...
+    describtionBtn.style.backgroundColor = 'red'
+    } else {
+        describtionBtn.style.backgroundColor = 'white'
+    }
+if(emailBtn.value.length > 0 &&  emailBtn.value.includes("@") &&
+    nameBtn.value.length > 0 && describtionBtn.value.length > 0  ){
+        alert("Thanks for filling out the form!")
+        emailBtn.value = ''
+        nameBtn.value = ''
+        describtionBtn.value = ''
```
dsobkow, line 59: Nice one, the form behaves correctly!🙂
You can find a button with a type of submit like this: 
```let submitButton = document.querySelector("button[type='submit’]");```

I can also see several if else statements checking the same condition, and then you check them again from line 54. It could be extracted to a separate function responsible for the validation. Remember, keep your code [DRY (Don't Repeat Yourself)](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) 🙂

PR #935, opened 2020-06-07
Zision week 5

zision71:
5th week homework submission.


week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,41 @@
...
+});
+
+let greenEl = document.getElementById("greenBtn");
+
+greenEl.addEventListener("click", function () {
+  divEl.style.backgroundColor = "#87ca8a";
+  donEl.style.backgroundColor = "black";
+  volEl.style.backgroundColor = "#8c9c08";
+  volEl.style.color = "White";
+});
```
dsobkow, line 32: Excellent work, buttons work correctly!👏 All three event listeners make very similar changes, therefore there is some repetition in your code. Can you think of the way how you could refactor the code so there is only one function responsible for making these changes? Remember, keep your code [DRY (Don't Repeat Yourself)](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) 🙂 And don't forget to remove commented out code and console logs before committing🙂

week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,41 @@
...
+
+let buttonEl = document.querySelector(
+  "#content > div.row > div.primary-content.col-lg-8 > form > button"
```
dsobkow, line 35: I think you haven't finished second part of the project with form validation. Also you can find a button with a type of submit like this: 
```let buttonEl = document.querySelector("button[type='submit’]");```

PR #941, opened 2020-06-11
JS week-6 with extras

elsahib:
# Your Details

Your Name:Osman Elsahib
Your City: Manchester
Your Slack Name:Osman Elsahib

# Homework Details

Module: Javascript 
Week: 6

### DIFF
--- a/week-6/Homework/mandatory/1-alarmclock/alarmclock.js
+++ b/week-6/Homework/mandatory/1-alarmclock/alarmclock.js
@@ -10,11 +10,11 @@ let timer;
 // This function will have a counter in seconds and will change the display by deducting a second everytime it's called. when the counter is 0 it will call playAlarm function
 function trigger() {
   let counter = parseInt(display.innerText.slice(0,2) * 60) + parseInt(display.innerText.slice(3));
-  let seconds = counter%60;
-  let minutes =  Math.floor(counter/60);
   if (counter > 0 ) {
-    display.innerText = minutes.toString().padStart(2,"0") + ":" + (seconds-1).toString().padStart(2,"0");
     counter--;
+    let seconds = counter % 60;
+    let minutes =  Math.floor(counter/60);
+    display.innerText = minutes.toString().padStart(2,"0") + ":" + seconds.toString().padStart(2,"0");
   } else if (counter === 0) {
     playAlarm();
   }
--- a/week-6/Homework/mandatory/1-alarmclock/alarmclock.js
+++ b/week-6/Homework/mandatory/1-alarmclock/alarmclock.js
@@ -22,16 +22,15 @@ function trigger() {
 
 // This function will read the value inputed by user, convert it to minutes and seconds then display it and call the trigger function every second.
 function setAlarm() {
-  if (timeValue.value <60 && timeValue.value > 0){
-    display.innerText = "00:" + timeValue.value.toString().padStart(2,"0");
-  } else {
-    display.innerText = Math.floor(timeValue.value/60).toString().padStart(2,"0") + ":" + (timeValue.value % 60).toString().padStart(2,"0");
-  }
+  display.innerText = Math.floor(timeValue.value/60).toString().padStart(2,"0") + ":" + (timeValue.value % 60).toString().padStart(2,"0");
+  if(!timer){
   timer = setInterval(trigger,1000);
+}
 }
 // This is an event handler for the stop button that will clear the interval to stop the alarm 
 stopbtn.addEventListener("click", () =>{
   clearInterval(timer); 
+  timer = null;
 })
 
 
--- a/week-6/Homework/mandatory/1-alarmclock/alarmclock.js
+++ b/week-6/Homework/mandatory/1-alarmclock/alarmclock.js
@@ -1,7 +1,40 @@
 // Implement the missing functionality according to the task description.
 // Hint: Remember which function you need to call when it's time to alarm.
+// Global variables
+let setbtn = document.getElementById("set");
+let stopbtn = document.getElementById("stop");
+let display = document.getElementById("timeRemaining");
+let timeValue = document.getElementById("alarmSet");
+let timer;
 
-function setAlarm() { }
+// This function will have a counter in seconds and will change the display by deducting a second everytime it's called. when the counter is 0 it will call playAlarm function
+function trigger() {
+  let counter = parseInt(display.innerText.slice(0,2) * 60) + parseInt(display.innerText.slice(3));
+  let seconds = counter%60;
+  let minutes =  Math.floor(counter/60);
+  if (counter > 0 ) {
+    display.innerText = minutes.toString().padStart(2,"0") + ":" + (seconds-1).toString().padStart(2,"0");
+    counter--;
+  } else if (counter === 0) {
+    playAlarm();
+  }
+}
+
+// This function will read the value inputed by user, convert it to minutes and seconds then display it and call the trigger function every second.
+function setAlarm() {
+  if (timeValue.value <60 && timeValue.value > 0){
+    display.innerText = "00:" + timeValue.value.toString().padStart(2,"0");
+  } else {
+    display.innerText = Math.floor(timeValue.value/60).toString().padStart(2,"0") + ":" + (timeValue.value % 60).toString().padStart(2,"0");
+  }
+  timer = setInterval(trigger,1000);
+}
+// This is an event handler for the stop button that will clear the interval to stop the alarm 
+stopbtn.addEventListener("click", () =>{
+  clearInterval(timer); 
+})
+
+
 
 // DO NOT EDIT BELOW HERE
 /* 
--- a/week-6/Homework/mandatory/1-alarmclock/index.html
+++ b/week-6/Homework/mandatory/1-alarmclock/index.html
@@ -2,7 +2,7 @@
 <html>
   <head>
     <title>Alarm Clock</title>
-    <script src="alarmclock.js"></script>
+    
     <link
       rel="stylesheet"
       href="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css"
@@ -14,7 +14,7 @@
 
   <body>
     <div class="centre">
-      <h1 id="timeRemaining">Time Remaining: 00:00</h1>
+      <h1>Time Remaining: <span id="timeRemaining">00:00</span></h1>
       <div>
         Set Time to:
         <input id="alarmSet" type="number" />
@@ -24,5 +24,6 @@
         <button id="stop" type="button">Stop Alarm</button>
       </div>
     </div>
+    <script src="alarmclock.js"></script>
   </body>
 </html>
--- a/week-6/Homework/mandatory/2-quotegenerator/index.html
+++ b/week-6/Homework/mandatory/2-quotegenerator/index.html
@@ -2,7 +2,7 @@
 <html>
   <head>
     <title>Quote Generator</title>
-    <script src="quotes.js"></script>
+    
     <link
       rel="stylesheet"
       href="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css"
@@ -12,6 +12,22 @@
     <link rel="stylesheet" href="style.css" />
   </head>
   <body>
+    <div class="jumbotron">
+      <h1 class="display-6">Get a random quote</h1>
+      <q class="lead" id="quote">Click the button or tick the box to get a quote</q>
+      <hr class="my-2">
+      <p id="author" class="lead">Author</p>
+      <p>
+        <a class="btn btn-primary" href="#" role="button" id="newbtn">New Quote</a>
+        
+      </p>
+      <p><label id="autolabel" class="form-check-label">
+          <input id="checkbox" class="form-check-input" type="checkbox" value="">
+          Auto quote
+        </label>
+      </p>
+    </div>
     <!-- Write your HTML in here -->
+    <script src="quotes.js"></script>
   </body>
 </html>
--- a/week-6/Homework/mandatory/2-quotegenerator/quotes.js
+++ b/week-6/Homework/mandatory/2-quotegenerator/quotes.js
@@ -7,9 +7,7 @@ let quote = document.getElementById("quote"),
     timer;
 
 // Event listener to call change function.
-newbtn.addEventListener("click", () => {
-change();
-});
+newbtn.addEventListener("click",change);
 
 // Function to call pickFromArray and change the display with the result
 function change(){
--- a/week-6/Homework/mandatory/2-quotegenerator/quotes.js
+++ b/week-6/Homework/mandatory/2-quotegenerator/quotes.js
@@ -1,3 +1,39 @@
+// Global variables
+let quote = document.getElementById("quote"), 
+    author = document.getElementById("author"),
+    newbtn = document.getElementById("newbtn"),
+    label = document.getElementById("autolabel"),
+    checkbox = document.getElementById("checkbox"),
+    timer;
+
+// Event listener to call change function.
+newbtn.addEventListener("click", () => {
+change();
+});
+
+// Function to call pickFromArray and change the display with the result
+function change(){
+  let temp = pickFromArray(quotes);
+      quote.innerText = temp.quote;
+      author.innerText = temp.author;
+};
+
+// event lister attahced to the checkbox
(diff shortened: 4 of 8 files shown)


week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,7 +1,40 @@
...
+}
+
+// This function will read the value inputed by user, convert it to minutes and seconds then display it and call the trigger function every second.
+function setAlarm() {
+  if (timeValue.value <60 && timeValue.value > 0){
+    display.innerText = "00:" + timeValue.value.toString().padStart(2,"0");
+  } else {
+    display.innerText = Math.floor(timeValue.value/60).toString().padStart(2,"0") + ":" + (timeValue.value % 60).toString().padStart(2,"0");
+  }
+  timer = setInterval(trigger,1000);
```
zoltan-gal, line 30: Only one caveat, I've spotted whilst tested your page, when I click the `Set Alarm` button more than once, it doesn't reset the timer already set up :). It makes the countdown a bit glitchy as the previous timer is not terminated. Also, alarm can't be turned off because the timer that got stuck in the background keeps it turning back :)

Apart from this minor issue, your code works pretty well.

week-6/Homework/mandatory/1-alarmclock/index.html
```diff
@@ -24,5 +24,6 @@ <h1 id="timeRemaining">Time Remaining: 00:00</h1>
         <button id="stop" type="button">Stop Alarm</button>
       </div>
     </div>
+    <script src="alarmclock.js"></script>
```
zoltan-gal, line 27: If you don't read elements immediately when the script is being executed first then it isn't necessary to be moved down. If elements are looked up inside functions that are invoked by button events then it won't throw any error because by that time page will have been fully loaded.

week-6/Homework/mandatory/2-quotegenerator/style.css
```diff
@@ -1 +1,27 @@
...
+    background-color: darkgoldenrod;
+}
+.my-2 {
+    background-color: #fca903;
+}
+.lead{
+    color: #fca903;
+    font-weight: 600;
+    font-size: 25px;
+
```
zoltan-gal, line 26: Pretty design!

week-6/Homework/mandatory/2-quotegenerator/quotes.js
```diff
@@ -1,3 +1,39 @@
...
+    author = document.getElementById("author"),
+    newbtn = document.getElementById("newbtn"),
+    label = document.getElementById("autolabel"),
+    checkbox = document.getElementById("checkbox"),
+    timer;
+
+// Event listener to call change function.
+newbtn.addEventListener("click", () => {
+change();
+});
```
zoltan-gal, line 12: In theory, as `change` is a function, it could be also passed as a parameter.
`newbtn.addEventListener("click", change)`

week-6/Homework/mandatory/2-quotegenerator/quotes.js
```diff
@@ -1,3 +1,39 @@
...
+
+// Function to call pickFromArray and change the display with the result
+function change(){
+  let temp = pickFromArray(quotes);
+      quote.innerText = temp.quote;
+      author.innerText = temp.author;
+};
+
+// event lister attahced to the checkbox
+checkbox.addEventListener('click', ()=> {
```
zoltan-gal, line 22: Splendid! You also smashed the additional tasks!

week-6/Homework/mandatory/3-slideshow/style.css
```diff
@@ -1 +1,11 @@
...
+    padding: 1em;
+    margin: 100px auto auto auto;
+    width: 900px;
+    align-content: center;
+    text-align: center;
+}
+.lead{
+    font-weight: 400;
+    font-size: 25px;
+}
```
zoltan-gal, line 11: Nice design! Good to see that you used bootstrap instead of writing everything from scratch. A nice example of how to reuse things 👍

week-6/Homework/mandatory/3-slideshow/slideshow.js
```diff
@@ -1 +1,79 @@
...
+fwdBtn.addEventListener("click", () => {
+forward();
+});
+// Manual backward event handler
+bckBtn.addEventListener("click", () => {
+backward();
+});
+// Automatic forward event handler
+autoFwd.addEventListener("click", ()=>{
+    clearInterval(timer);
```
zoltan-gal, line 39: Good :) you didn't miss here to erase the timer before creating another one.

week-6/Homework/mandatory/3-slideshow/slideshow.js
```diff
@@ -1 +1,79 @@
...
+    autoFwd = document.getElementById("autoForward"),
+    autoBck = document.getElementById("autoBackward"),
+    stop = document.getElementById("stop"), 
+    timer,
+    userPref = document.getElementById("setInterval");
+    
+// Manual forward event handler
+fwdBtn.addEventListener("click", () => {
+forward();
+});
```
zoltan-gal, line 32: In this case, it can be simply `fwdBtn.addEventListener("click", forward)`. Likewise, I've mentioned somewhere earlier.

week-6/Homework/mandatory/3-slideshow/slideshow.js
```diff
@@ -1 +1,79 @@
...
+bckBtn.addEventListener("click", () => {
+backward();
+});
+// Automatic forward event handler
+autoFwd.addEventListener("click", ()=>{
+    clearInterval(timer);
+    if (userPref.value > 0) {
+        timer = setInterval(forward,userPref.value*1000);
+    } else {
+        timer = setInterval(forward,3000);
```
zoltan-gal, line 43: I'd rather you defined constant values like this in global variable e.g. `const defaultTime = 3000` and then you can reuse it as many times as you need it. You have then change it only one place if you wish to modify the default time.

week-6/Homework/mandatory/3-slideshow/slideshow.js
```diff
@@ -1 +1,79 @@
...
+});
+// Stop automation handler
+stop.addEventListener("click",() => {
+    clearInterval(timer);
+});
+
+// added clear interval at the end of the cycle for both functions to make life easier I guess
+// forward function
+function backward() {
+    let current = imagesArr.indexOf(imageDis.src);
```
zoltan-gal, line 64: It might be better if you store the index of the current image in a variable rather than finding based on an element's property. It is a similar issue than the one, I expanded on the alarm clock exercise, saying that code logic unnecessarily depends on HTML elements.

week-6/Homework/mandatory/3-slideshow/slideshow.js
```diff
@@ -1 +1,79 @@
...
+    } 
+);
+// Automatic backward event handler
+autoBck.addEventListener("click", () => {
+    clearInterval(timer);
+    if (userPref.value > 0 ) {
+        timer = setInterval(backward,userPref.value*1000);
+    } else {
+        timer = setInterval(backward,3000);
+    }
```
zoltan-gal, line 54: This code is nearly the same than in forward event handler. Could you generalise this logic by factoring the common bits out from both of them into one new function and configuring the differences via parameters?
As far as I can see it, the only difference is the function you have to call in the timer.

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,7 +1,40 @@
...
+// Global variables
+let setbtn = document.getElementById("set");
+let stopbtn = document.getElementById("stop");
+let display = document.getElementById("timeRemaining");
+let timeValue = document.getElementById("alarmSet");
+let timer;
+
+// This function will have a counter in seconds and will change the display by deducting a second everytime it's called. when the counter is 0 it will call playAlarm function
+function trigger() {
+  let counter = parseInt(display.innerText.slice(0,2) * 60) + parseInt(display.innerText.slice(3));
```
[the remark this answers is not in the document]
elsahib: Thought about this when I revisited the code To correct for a negative value, will try to implement it this week

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
elsahib: You mean that I should clear the interval before assigning it in the code?  same way as line 34?

week-6/Homework/mandatory/2-quotegenerator/quotes.js
elsahib: I got carried away with the arrow functions :D

week-6/Homework/mandatory/1-alarmclock/index.html
elsahib: I haven’t tried it before moving it to be honest, But one of the students posted about a problem she faced while solving the task and the solution was put the script to the bottom of the page. Thats why I went with it.

week-6/Homework/mandatory/3-slideshow/slideshow.js
elsahib: Brilliant suggestion, Thanks.

elsahib:
Thanks for your comments, I’ll apply them and If I failed to make any of them I hope you can be free to guide me through.
Your time is very much appreciated.

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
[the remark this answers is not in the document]
zoltan-gal: It is also good to handle invalid numbers like a negative number.

What I meant is not to read the value from the `display.innerText` but read it from a variable instead. Only read the the `display` element once when you click on the `Set Alarm` button to initiate the alarm clock. 

E.g.
```
let counter = parseInt(timeValue.value);
// then when trigger is called, you would only
//  - deduct one from the counter (counter--)
//  - and set display (display.innerText = ....)
trigger();
```

week-6/Homework/mandatory/1-alarmclock/index.html
zoltan-gal: That's okay. Just wanted to make sure you are aware of when a script fails and when not.

Basically, when the browser sees a script tag, it tries to execute that. But your functions are only called inside when something refers to and invokes them. It can be triggered by an event on a button, deferred call by a timer or of course you can call functions immediately.
But if the script tries to refer to something on the DOM that hasn't been loaded at that time then it will fail.

There are different technics to prevent errors. One can be what you did, moving the script tag to another place inside the HTML. Or you could see the `window.onload` event, it is also another way.

week-6/Homework/mandatory/2-quotegenerator/index.html
```diff
@@ -12,6 +12,22 @@
...
+      <h1 class="display-6">Get a random quote</h1>
+      <q class="lead" id="quote">Click the button or tick the box to get a quote</q>
+      <hr class="my-2">
+      <p id="author" class="lead">Author</p>
+      <p>
+        <a class="btn btn-primary" href="#" role="button" id="newbtn">New Quote</a>
+        
+      </p>
+      <p><label id="autolabel" class="form-check-label">
+          <input id="checkbox" class="form-check-input" type="checkbox" value="">
```
zoltan-gal, line 25: It might be more idiomatic to close a single self-closing tag. But if you don't close it shouldn't cause any harm. My knowledge is staled on this subject, I have to make a research :)

week-6/Homework/mandatory/1-alarmclock/index.html
zoltan-gal: @elsahib 
After reviewing others' code, I've noticed that you were right. It is a better approach, in this case, moving the script tag down because you must attach the event listener to the stop button. (Unfortunately, the `window.onload` event is already specified and it can't be added more than once and moreover the stipulation was to "do not modify the code below" 😄 .

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
elsahib: I solved this with the following:
1. In setAlarm Function added one more condition as follows 
```
if(!timer){
  timer = setInterval(trigger,1000);
}
```

2.  In the eventlistener for Stop Button after clearing the interval I gave the timer a value of null
```
  clearInterval(timer); 
  timer = null;
```

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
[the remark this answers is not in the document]
elsahib: When I tried to define the counter outside of the trigger funtion it had no value assigned to it as it grabs the value as soon as the page is loaded I suppose.

week-6/Homework/mandatory/2-quotegenerator/quotes.js
elsahib: corrected this one.

week-6/Homework/mandatory/3-slideshow/slideshow.js
elsahib: Implemented.

week-6/Homework/mandatory/3-slideshow/slideshow.js
elsahib: created a new automation function that accepts a function as a parameter, but when calling that function in the eventlistener I couldn't pass in the arguments without creating an annonymos function to call the automation function with the desired argument.

elsahib:
I've implemented the bits I've changed and commented on the bits I couldn't change.

week-6/Homework/mandatory/2-quotegenerator/quotes.js
zoltan-gal: No worries, end of the day it did the same.

week-6/Homework/mandatory/3-slideshow/slideshow.js
zoltan-gal: In that case, you need that anonymous function :), otherwise, nothing will pass the desired argument onto it.

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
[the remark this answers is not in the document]
zoltan-gal: it would look stg like this:
 - `let counter = 0;`
 - then in your `setAlarm` function you have to update it `counter = parseInt(timeValue.value);`
 - and finally your `trigger` function will be decreasing it.

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
zoltan-gal: Yes, because what if you don't click the stop button before set alarm again?
Try it:
Set the time to let's say 30 sec and then click set alarm button 5 times. You will see what happens :)

zoltan-gal: Okay I see, you fixed it now, you added and if case saying
```
if(!timer){
  timer = setInterval(trigger,1000);
}
```
Instead of shutting down the timer and creating a new one, instead, you reuse the same one if it already exists and you only update the display time.
It is a good approach too ⭐ Well done ⭐

week-6/Homework/mandatory/2-quotegenerator/quotes.js
```diff
@@ -7,9 +7,7 @@ let quote = document.getElementById("quote"),
     timer;
 
 // Event listener to call change function.
-newbtn.addEventListener("click", () => {
-change();
-});
+newbtn.addEventListener("click",change);
```
zoltan-gal, line 10: Nice :)

week-6/Homework/mandatory/3-slideshow/slideshow.js
```diff
@@ -24,40 +24,36 @@ let imagesArr = ["https://images.unsplash.com/photo-1452860606245-08befc0ff44b",
...
-
+// Automation function
+function automate(callback){
+    clearInterval(timer);
+    if (userPref.value > 0 ) {
+        timer = setInterval(callback,userPref.value*1000);
+    } else {
+        timer = setInterval(callback,defaultTime);
+    }
+}
```
zoltan-gal, line 56: That's awesome 🏆 ! You adapted the callback pattern we talked about. Just an extra theory if you keen on learning more.

This concept when you pass a function as a parameter or return function is called (Higher-order function)[https://en.wikipedia.org/wiki/Higher-order_function]. Another page: https://eloquentjavascript.net/05_higher_order.html

PR #943, opened 2020-06-11
Week5

andsemenov:
# Your Details

Your Name:Andrei Semenov
Your City:Glasgow
Your Slack Name:andsemenov

# Homework Details

Module:JS
Week:Week5 (Inclass and homework)


week-5/Homework/mandatory/1-exercises/exercises.js
```diff
@@ -0,0 +1,164 @@
...
+    hOnes = document.createElement("h1");
+    hOnes.innerHTML = element.name;
+    hTwoes = document.createElement("h2");
+    hTwoes.innerHTML = element.job;
+    let myDiv = document.querySelector("#content");
+    myDiv.appendChild(hOnes);
+    myDiv.appendChild(hTwoes);
+  });
+}
+
```
alastair87, line 28: Make sure that you use an appropriate keyword when declaring a variable for the first time within a given scope - i.e. `var`, `let` or `const`. If you don't have any such keyword then JavaScript will automatically assume that you want the variable initialised with global scope or that you want to update an existing variable accessible within scope. This can be a problem since you can accidentally overwrite data that's needed somewhere else in your code.

So, for example, in `exerciseTwo()`:

```js
const ulEl = document.createElement(ul);
```

Note that it is OK to use `const` here since `const` does not prevent you from modifying the object, such as to, it simply prevents you assigning another value to that variable identifier (`ulEl`) within the same scope. This might be a good thing since it means if you were to unintentionally do this later in your code then you would get an error.

week-5/Homework/mandatory/1-exercises/exercises.js
```diff
@@ -0,0 +1,164 @@
...
+  ulEl.className = "ulstyle";
+  let content = document.querySelector("#content");
+  content.appendChild(ulEl);
+  shoppingItems.forEach((element) => {
+    let liEl = document.createElement("li");
+    liEl.innerHTML = element;
+    ulEl.appendChild(liEl);
+  });
+}
+
```
alastair87, line 48: Following on from what I said in my comment on `exerciseOne()`.

In `exerciseTwo()`:

```js
const ulEl = document.createElement(ul);
```

Note that it is OK to use `const` here since `const` does not prevent you from modifying the object, such as to, it simply prevents you assigning another value to that variable identifier (`ulEl`) within the same scope. This might be a good thing since it means if you were to unintentionally do this later in your code then you would get an error.

week-5/Homework/mandatory/2-project/js/main.js
```diff
@@ -0,0 +1,97 @@
...
+    event.preventDefault();
+    email.style.backgroundColor = "white";
+    yourName.style.backgroundColor = "white";
+    describeYourself.style.backgroundColor = "white";
+    email.value = "";
+    yourName.value = "";
+    describeYourself.value = "";
+  }
+}
+
```
alastair87, line 85: This works, but it's quite complicated. There might well be ways you could simplify it to make it easier to maintain, for example so as you don't have to check any of the conditions twice. In this situation it doesn't matter much, but it's possible you might be doing something in another situation where you have to run a function that takes a while, or accesses network resources, which you'd want to keep to a minimum.

week-5/Homework/mandatory/2-project/js/main.js
```diff
@@ -0,0 +1,97 @@
...
+function greenBtnFunc() {
+  document.querySelector(".jumbotron").style.backgroundColor = `#87ca8a`;
+  document.querySelector(
+    ".btn.btn-primary.btn-lrg"
+  ).style.backgroundColor = `black`;
+  document.querySelector(
+    ".btn.btn-secondary.btn-lrg"
+  ).style.backgroundColor = `#8c9c08`;
+}
+
```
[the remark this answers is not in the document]
andsemenov: Thanks for your feedback. Unfortunately, I can not make it work properly. I supposed that this function setColors should be anonymous, otherwise addEventListener calls it without button pressing. But if I use anonymous function, it will come to nothing the advantage of function. Could you give some advice how can I avoid this issue please? Thanks for you assistance. There is a code for a colour button. `let blueBtn = document.querySelector("#blueBtn");
blueBtn.addEventListener(
  "click",
  setColor("#588fbd", "#ffa500", "black", "white")
);
`

week-5/Homework/mandatory/2-project/js/main.js
andsemenov: Thanks for your feedback. I changed the code and removed dual check-up. But I noticed that code is working differently. It paints in red invalid fields one after another, not simultaneously, as it was early. I am not sure that it is so bad, but it can increase quantity of button pressures. Thanks for your assistance.
`function submitFunc() {
  let email = document.querySelector("#exampleInputEmail1");
  let yourName = document.querySelector("#example-text-input");
  let describeYourself = document.querySelector("#exampleTextarea");
  event.preventDefault();

  if (email.value.length === 0 || email.value.includes("@") === false) {
    email.style.backgroundColor = "red";
  } else if (yourName.value.length === 0) {
    yourName.style.backgroundColor = "red";
  } else if (describeYourself.value.length === 0) {
    describeYourself.style.backgroundColor = "red";
  } else {
    alert("Thank you for filling");
    email.style.backgroundColor = "white";
    yourName.style.backgroundColor = "white";
    describeYourself.style.backgroundColor = "white";
    email.value = "";
    yourName.value = "";
    describeYourself.value = "";
  }
}`

PR #944, opened 2020-06-12
Hedyeh week 6

Hedyeh65:
# Your Details

Your Name:
Your City:
Your Slack Name:

# Homework Details

Module:
Week:

### DIFF
--- a/week-4/Homework/mandatory/2-writers.js
+++ b/week-4/Homework/mandatory/2-writers.js
@@ -22,29 +22,29 @@ let writers = [
     lastName: "Woolf",
     occupation: "writer",
     age: 59,
-    alive: false
+    alive: false,
   },
   {
     firstName: "Zadie",
     lastName: "Smith",
     occupation: "writer",
     age: 41,
-    alive: true
+    alive: true,
   },
   {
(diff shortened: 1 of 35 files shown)


week-6/Homework/mandatory/3-slideshow/slideshow.js
```diff
@@ -1 +1,59 @@
 // Write your code here
+const imageArray = [
```
msachi, line 2: Nice solution, well done! It works well generally but I think there are a few little bugs to do with how you are looping through the images.

You should manipulate the i first, and only set the image at the end. So something like:
```
  i++;
  if (i >= imageArray.length) {
    i = 0;
  }
  getImages.src = imageArray[i];
```

week-4/Homework/mandatory/2-writers.js
```diff
@@ -22,29 +22,29 @@ let writers = [
     lastName: "Woolf",
```
msachi, line 22: Also please try to follow the github process correctly when making a pull request :) Changes from weeks 4 and 5 should not be showing in this pull request.

PR #947, opened 2020-06-12
js2-week5 done

HarrisSheraz20:
# Your Details

Your Name: Harris Sheraz
Your City: London 
Your Slack Name: Harris Sheraz 

# Homework Details

Module: JS2
Week: 2

### DIFF
--- a/week-5/Homework/mandatory/2-exercises/exercises.js
+++ b/week-5/Homework/mandatory/2-exercises/exercises.js
@@ -15,6 +15,16 @@
  */
 function exerciseOne(arrayOfPeople) {
   let content = document.querySelector("#content");
+
+  arrayOfPeople.forEach((x) => {
+    let createH1 = document.createElement("h1");
+    let createH2 = document.createElement("h2");
+    content.appendChild(createH1);
+    content.appendChild(createH2);
+
+    createH1.textContent = x.name;
+    createH2.textContent = x.job;
+  });
 }
 
 /**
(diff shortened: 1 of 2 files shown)


week-5/Homework/mandatory/2-exercises/exercises.js
```diff
@@ -48,17 +65,52 @@ function exerciseTwo(shopping) {
...
+  books.forEach((x) => {
+    let createLi = document.createElement("li");
+    createUl.appendChild(createLi);
+    let createP = document.createElement("p");
+    createLi.appendChild(createP);
+    createP.textContent = `${x.title} - ${x.author} `;
+
+    let createImage = document.createElement("img");
+    createLi.appendChild(createImage);
+    let createSrc = document.createAttribute("src");
```
illicitonion, line 94: This works, but we can actually do something a little simpler, with:
```javascript
createImage.src = "https://image.ebooks.com/previews/095/095767/095767915/095767915-hq-168-80.jpg";
```

then we don't need to call `createAttribute` or `setAttributeNode` or the `.value` :)

Similarly for height (though ideally we'd set that as a CSS property, rather than an HTML attribute, like you do with `backgroundColor`)

week-5/Homework/mandatory/3-project/js/main.js
```diff
@@ -0,0 +1,52 @@
...
+let targetEmail = document.querySelector("div #exampleInputEmail1");
+let targetName = document.querySelector("div #example-text-input");
+let targetDescribeYourself = document.querySelector("div #exampleTextarea");
+
+let emailExpression = /[a-z0-9][@.]/;
+
+targetSubmit.addEventListener("click", (e) => {
+  e.preventDefault();
+
+  validateAll() ? alert("thank you") : alert("no");
```
illicitonion, line 39: Every time I fill in this form it says "no" - can you make it so that it says "thank you" when I fill it in right? :)

PR #957, opened 2020-06-18
Week6 

andsemenov:
# Your Details

Your Name:Andrei
Your City:Glasgow
Your Slack Name:andsemenov

# Homework Details

Module:JS II
Week:6


week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,4 +1,71 @@
-function setAlarm() {}
+function setAlarm() {
+  //get the value of the input field
+  let inputField = document.querySelector("#alarmSet").value;
+
```
alastair87, line 4: You need to do some input validation here. What happens if the input field doesn't have a value or it's unsuitable - e.g. it's a negative number or has letters or spaces?

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,4 +1,71 @@
...
+    document.querySelector(".pauseButton").hidden = true;
+
+    let restartButton = document.createElement("button");
+    restartButton.textContent = "Restart";
+    let divButton = document.querySelector(".centre >:last-child");
+    divButton.prepend(restartButton);
+    restartButton.className = "restartButton";
+
+    //restart functionality
+    document.querySelector(".restartButton").addEventListener("click", () => {
```
alastair87, line 34: What about if I want to unpause the timer? The restart button just seems to reset it. Think about how you could maybe control the flow of execution in the timer in such a way as you could easily toggle the paused state (something like `if (paused)`).

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,4 +1,71 @@
...
+      location.reload();
+    });
+  });
+
+  //update the `Time Remaining` title every second
+  myAlarm = setInterval(() => {
+    if (inputField == 0) {
```
alastair87, line 41: See the commend about validation above. I see you've used `==` rather than `===` to check the equality. This only applies loose equality checking which is why it works in some cases even though the original type of the value is a string. Ideally I would suggest having some function that parses the input field value and converts it to a valid integer or returns some fixed value such as `-1` if it is invalid. This value, if its valid, should be an integer that you can then check to see if it is 0 or not using strict equality (`===`).

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,4 +1,71 @@
...
+        setInterval(function () {
+          if (index % 2) {
+            document.body.style.backgroundColor = "white";
+            index++;
+          } else {
+            document.body.style.backgroundColor = "orange";
+            index++;
+            timeRemaining.textContent = `Time Remaining: ${inputField}:00`;
+          }
+        }, 1000);
```
alastair87, line 60: This is good. If you look up the modulus operator (`%`) you might get an idea of how you could have an array of colours of any length you want and iterate through them on each interval with somewhat less code. This is strictly an extension for if you have time though, rather than something to get too involved in.

See here for an introduction to the modulus: https://riptutorial.com/javascript/example/760/remainder---modulus----

week-6/Homework/mandatory/2-quotegenerator/quotes.js
```diff
@@ -1,3 +1,66 @@
...
+  button.style.backgroundColor = color;
+}
+
+function colorGenerator() {
+  let color = `rgb(${Math.random() * 255},${Math.random() * 255},${
+    Math.random() * 255
+  })`;
+  return color;
+}
+
```
alastair87, line 63: Excellent work. I'm impressed that you thought to generalise the idea of using the `Math.random()` function to your own work, in order to pick the colour.

week-6/Homework/mandatory/3-DOM-practice/main.js
```diff
@@ -0,0 +1,33 @@
...
+let years = [
+  "December 9, 1906 — January 1, 1992",
+  "August 26, 1918 — February 24, 2020",
+  "December 10, 1815 — November 27, 1852",
+];
+document.querySelectorAll("section").forEach((element) => {
+  let paragraph = document.createElement("p");
+  paragraph.textContent = years[index];
+  element.append(paragraph);
+  index++;
```
alastair87, line 32: All lookds good.

week-6/Homework/extra/2-slideshow/slideshow.js
```diff
@@ -0,0 +1,95 @@
...
+
+function setup() {
+  document.querySelector("#backBtn").addEventListener("click", backMove);
+  document.querySelector("#forwardBtn").addEventListener("click", forwardMove);
+  document
+    .querySelector("#autobackBtn")
+    .addEventListener("click", autoBackMove);
+  document
+    .querySelector("#autoforwardBtn")
+    .addEventListener("click", autoForwardMove);
```
alastair87, line 92: Very good, but it would help to have some comments to explain more clearly how your code works.

week-6/Homework/extra/2-slideshow/slideshow.js
```diff
@@ -0,0 +1,95 @@
...
+  clearInterval(timerRemainTime);
+}
+
+let timerImage, timerRemainTime;
+function setTimer(func) {
+  timerImage = setInterval(func, 5000);
+}
+
+function addRemainTime() {
+  let timer = 5;
```
alastair87, line 71: This is where you might need to refactor your code a bit if you wanted the user to be able to change the time, as suggested in the extension task.

week-6/Homework/extra/2-slideshow/slideshow.js
andsemenov: Thanks for your feedback. I have added notes to the functions. Thanks for your assistance.

week-6/Homework/extra/2-slideshow/slideshow.js
andsemenov: Thanks for your feedback. I have added the ability to change the time for user. But I am not sure that it is the best solution because after changing interval user needs to run automatic movement again. Unfortunately, I could not find way to know which function was running at the changing moment in order to make it working again. Could you advise something about it please?  Thanks for your assistance.

week-6/Homework/mandatory/3-DOM-practice/main.js
andsemenov: Thanks for your feedback. Thanks for your assistance.

week-6/Homework/mandatory/2-quotegenerator/quotes.js
andsemenov: Thanks for your feedback. Thanks for your assistance.

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
andsemenov: Thanks for your feedback. I have added function for input validation. It informs that user inputted wrong value and reload web page.  Thanks for your assistance.

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
andsemenov: Thanks for your feedback. I have added the ability to unpause it. Thanks for your assistance.

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
andsemenov: Thanks for your feedback. I have used a function which checks 2 conditions: the value is positive and contains only digits. Thanks for your assistance.

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
andsemenov: Thanks for your feedback. I've read the article about modulus operator, but I've not understood yet. 
But I can form the array of colours just calling function like this:
`function randomColor() {
  let letters = "abcdef1234567890";
  let color = "#";
  for (let i = 0; i < 6; i++) {
    color = color + letters[Math.floor(Math.random() * 16)];
  }
  return color;
}`
But it is interesting question, I will think about it in the background.
Thanks for your assistance.

PR #959, opened 2020-06-18
Improve JS1 week2 exercise 1

illicitonion:
* All functions take arguments - artificial examples which don't involve
  taking input make it less clear to students why functions are useful.
* Show expected/actual - this both helps the students debug, and better
  matches what they should see professionally from tests.
* Test that first5 used slice not splice.


illicitonion:
Re-created in new repo: https://github.com/CodeYourFuture/JavaScript-Core-1-Homework/pull/14

PR #966, opened 2020-06-19
Bekoo week6

Bek0o0:
# Your Details

Your Name:
Your City:
Your Slack Name:

# Homework Details

Module:
Week:


week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,4 +1,39 @@
-function setAlarm() {}
+function setAlarm() {
```
zoltan-gal, line 1: The basic idea is not bad in this function and your structured your code very well, the presented a separation of concern is also great. However, there is a fundamental concept issue saying that you set up your code each time when you click on the Set button. After each clicks when this listener runs:
```
document.getElementById("set").addEventListener("click", () => {
    setAlarm();
  });
```
it attaches a new set of event listeners to your buttons inside your function without removing the old ones. This means that as many times you click the Set button, as many event listeners you will have on each button, as a result of this, when you click a button, all of the attached event listeners on that will be triggered and executed.

The `setAlarm` function should be responsible to either remove the listeners before adding new ones or prevent adding again if they are already attached.

Have you tried to test your code? It is very easy to break. Try to pay attention to the testing side. A good strategy can be defining different test scenarios that will help you focus on one thing at a time. 
For example:
 1. when I set value and click the set button then time should start
 2. when I set a value smaller than 10, seconds should be padded to 2-character long with a leading zero.
 3. when the timer is up, the counter should stop and alarm clock start

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,4 +1,39 @@
...
+    Counter = setInterval(CountDown, 1000);
+  });
+  let TheTime;
+  setIt.addEventListener("click", function () {
+    TheTime = document.getElementById("alarmSet").value;
+
+    if (TheTime <= 60) {
+      count.innerHTML = "Time Remaining: 00:" + TheTime;
+    }
+  });
```
zoltan-gal, line 20: This should also start the countdown. It only works if you double click on the Set button which is a bit dodgy.

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,4 +1,39 @@
...
+  let Counter;
+  pause.addEventListener("click", function () {
+    clearInterval(Counter);
+  });
+  Restart.addEventListener("click", function () {
+    Counter = setInterval(CountDown, 1000);
+  });
+  let TheTime;
+  setIt.addEventListener("click", function () {
+    TheTime = document.getElementById("alarmSet").value;
```
zoltan-gal, line 15: I would rather parse this to a number because even if the type of the input box is number this function will still return a string. JS is quite lenient so let you threat string as a number in given circumstances but it is not a good practice and can lead to a cunning bug later on.

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,4 +1,39 @@
-function setAlarm() {}
+function setAlarm() {
+  let count = document.getElementById("timeRemaining");
+  let setIt = document.getElementById("set");
+  let pause = document.getElementById("p");
+  let Restart = document.getElementById("re");
+  let Counter;
```
zoltan-gal, line 6: The naming of the variables names should be consistent. I'd suggest following the standard naming conventions, for variable names the naming convention is usually camel case that means that the first-word start is traight lowercased and then the first letter of each following words are uppercased.
 
Check this article out:
https://www.freecodecamp.org/news/javascript-naming-conventions-dos-and-don-ts-99c0e2fdd78a/

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,4 +1,39 @@
...
+    clearInterval(Counter);
+  });
+  Restart.addEventListener("click", function () {
+    Counter = setInterval(CountDown, 1000);
+  });
+  let TheTime;
+  setIt.addEventListener("click", function () {
+    TheTime = document.getElementById("alarmSet").value;
+
+    if (TheTime <= 60) {
```
zoltan-gal, line 17: What if the time is bigger than 60?

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,4 +1,39 @@
...
+      count.innerHTML = "Time Remaining: 00:" + TheTime;
+    }
+  });
+  Counter = setInterval(CountDown, 1000);
+  function CountDown() {
+    if (TheTime <= 60) {
+      count.innerHTML = "Time Remaining: 00:" + TheTime;
+    }
+    if (TheTime == 0 || null) {
```
zoltan-gal, line 26: I can't really follow this condition. What does the `or` part mean?

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,4 +1,39 @@
...
+      count.innerHTML = "Time Remaining: 00:00";
+      playAlarm();
+      alert("Time is up");
+      clearInterval(Counter);
+      document.body.style.backgroundColor = "gray";
+    } else {
+      TheTime = TheTime - 1;
```
zoltan-gal, line 33: you can also write: `TheTime -= 1;` it is a shorter format

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,4 +1,39 @@
...
+
+    if (TheTime <= 60) {
+      count.innerHTML = "Time Remaining: 00:" + TheTime;
+    }
+  });
+  Counter = setInterval(CountDown, 1000);
+  function CountDown() {
+    if (TheTime <= 60) {
+      count.innerHTML = "Time Remaining: 00:" + TheTime;
+    }
```
zoltan-gal, line 25: This part is duplicated. Do you remember the DRY rule? Meaning Do not Repeat Yourself. When you have repeated code lines, it is worth considering to factor those out into a separate function and just you have to refer to the function.

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,4 +1,39 @@
...
+  });
+  Restart.addEventListener("click", function () {
+    Counter = setInterval(CountDown, 1000);
+  });
+  let TheTime;
+  setIt.addEventListener("click", function () {
+    TheTime = document.getElementById("alarmSet").value;
+
+    if (TheTime <= 60) {
+      count.innerHTML = "Time Remaining: 00:" + TheTime;
```
zoltan-gal, line 18: Time should be padded to 2-character long. So when time is smaller than 10, a `0` should be prepended e.g. `06`.
You can use the `padStart` built-in JS function to do that.

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
zoltan-gal: Try to use the debugger to inspect and understand better how the code behaves and why. Does it work as expected or not? Alternatively, you can use console log that would also give some clue telling what is happening at a given momentum

PR #967, opened 2020-06-19
Shahd week 6

shahdalhaj:
# Your Details

Your Name:
Your City:
Your Slack Name:

# Homework Details

Module:
Week:


week-6/Homework/mandatory/2-quotegenerator/quotes.js
```diff
@@ -489,3 +490,19 @@ const quotes = [
     author: "Zig Ziglar",
   },
 ];
+
+
+function qoutesGenerator(){
```
msachi, line 495: Nice, you were close! Just two things to fix to make it work:
- move all the code from the function quotesGenerator outside (there is no need to wrap it in a function)
- move the script tag in your html below the body (the problem is that, at the moment, you are running the JS before all the HTML has been loaded, so the JS doesn't find the HTML elements

week-6/Homework/mandatory/1-alarmclock/alarmclock.js
```diff
@@ -1,7 +1,59 @@
...
+  
+counter--;
+alarmSet.innerHTML  =  counter;
+timer.innerText = ("time remaining: 00.0" + counter);
+if (counter === 0){
+  playAlarm()
+}
+} */
+ 
+function setAlarm() {
```
msachi, line 27: Good effort, but I think you overcomplicated it a bit. There is no need for a for loop, or for any additional event listeners.

Have a look at Mohammed's solution, which is quite simple but it does the job. You might find it easier to understand than the official solution:
https://github.com/CodeYourFuture/js-exercises/blob/31c8642d3646e002e62290db41196d787df74aee/week-6/Homework/mandatory/1-alarmclock/alarmclock.js

week-6/Homework/mandatory/3-slideshow/slideshow.js
```diff
@@ -1 +1,84 @@
...
+function changeImg(){
+	document.querySelector(".slides").src = images[i];
+
+	if(i < images.length - 1){
+	  i++; 
+	} else { 
+		i = 0;
+	}
+
+	setTimeout("changeImg()", time);
```
msachi, line 37: No need for setTimeout here - it is the manual forward action.

week-6/Homework/mandatory/3-slideshow/slideshow.js
```diff
@@ -1 +1,84 @@
...
+function changeImg1(){
+	document.querySelector(".slides").src = images[i];
+
+	if(i === 0){
+	  i = images.length - 1; 
+	} else { 
+		i --;
+	}
+
+	setTimeout("changeImg1()", time);
```
msachi, line 50: Again no need for setTimeout here.

week-6/Homework/mandatory/3-slideshow/slideshow.js
```diff
@@ -1 +1,84 @@
...
+images[1] = "https://cdn.pixabay.com/photo/2014/09/14/18/04/dandelion-445228__340.jpg";
+images[2] = "https://cdn.pixabay.com/photo/2018/12/15/02/53/flower-3876195__340.jpg";
+images[3] = "https://cdn.pixabay.com/photo/2013/07/30/12/25/marriage-168831__340.jpg";
+
+let goButton = document.querySelector("#forward");
+goButton.addEventListener("click",changeImg);
+let backButton = document.querySelector("#backward");
+backButton.addEventListener("click",changeImg1)
+let autoForward = document.querySelector("#autoforward");
+autoForward.addEventListener("click",forwardAutoPlay);
```
msachi, line 15: I think you got the name of the functions a bit wrong - this should be `forwardPlay` right? (and same for `backwardPlay`)

week-6/Homework/mandatory/3-slideshow/slideshow.js
```diff
@@ -1 +1,84 @@
...
+   
+        document.querySelector(".slides").src = images[i];
+    
+        if(i < images.length - 1){
+          i++; 
+        } else { 
+            i = 0;
+        }
+    
+        setTimeout("changeImg()", time);
```
msachi, line 65: Here we need a setInterval here instead of the setTimeout, to make sure that the function executes every second. The setInterval should wrap around all the logic in this function.

week-6/Homework/mandatory/3-slideshow/slideshow.js
```diff
@@ -1 +1,84 @@
...
+  function backwardPlay() {
+    document.querySelector(".slides").src = images[i];
+
+	if(i === 0){
+	  i = images.length - 1; 
+	} else { 
+		i --;
+	}
+
+	setTimeout("changeImg1()", time);
```
msachi, line 79: Again we need a setInterval here.

week-6/Homework/mandatory/3-slideshow/slideshow.js
```diff
@@ -1 +1,84 @@
...
+function stoper() {
+    stopButton.addEventListener.addEventListener("click",()=>{
+        clearInterval(clear);
+      });
+    
+}
+ 
+
+function changeImg(){
+	document.querySelector(".slides").src = images[i];
```
msachi, line 29: In all of your functions, you should do the `i` manipulation first, and then set the `src` only at the end. This will ensure correct slideshow behaviour.

week-6/Homework/mandatory/3-slideshow/slideshow.js
```diff
@@ -1 +1,84 @@
...
+	  i = images.length - 1; 
+	} else { 
+		i --;
+	}
+
+	setTimeout("changeImg1()", time);
+}
+
+    
+
```
msachi, line 83: Very good! You got a little confused about setInterval and setTimeout, but you've got the right general idea :)

PR #987, opened 2020-06-26
resubmission in week-7-2 branch.

zision71:
# Your Details

Your Name: Zision(zee)
Your City: Manchester
Your Slack Name: zee

# Homework Details
Debugging exercise.
Module:
Week: 7

### DIFF
--- a/week-7/Homework/mandatory/1-debugging-practice/script.js
+++ b/week-7/Homework/mandatory/1-debugging-practice/script.js
@@ -31,14 +31,17 @@ function submit() {
   if (
     title.value == null ||
     title.value == "" ||
+    author.value == null ||
+    author.value == "" ||
     pages.value == null ||
-    pages.value == ""
+    pages.value <= 0 ||
+    check.checked == false
   ) {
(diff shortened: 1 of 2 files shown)


evanrolfe:
Great work Zee! This is looking very good, I will only ask you to make one change to this: currently when you click submit to create a new book, the javascript code is checking whether or not the "Read" checkbox has been checked, this means that it only lets you add a book if that checkbox is checked. But that doesn't really make sense for this form, because what if I want to add a book that I haven't read? Please can you correct that and let me know. Thanks

evanrolfe:
@Iman-feisali no problem.

evanrolfe:
@zision71 Great work on this homework! Its looking very good. I will only ask you to make one change to your solution: currently when you click submit to create a new book, the javascript code is checking whether or not the "Read" checkbox has been checked, this means that it only lets you add a book if that checkbox is checked. But that doesn't really make sense for this form, because what if I want to add a book that I haven't read? Please can you correct that and let me know. Thanks

PR #1001, opened 2020-07-01
Week6

alastair87:
# Your Details

Your Name:
Your City:
Your Slack Name:

# Homework Details

Module:
Week:


week-6/Homework/extra/2-slideshow/slideshow.js
```diff
@@ -1 +1,61 @@
...
+let stopBtn = document.querySelector("#stop");
+stopBtn.addEventListener("click", stopAutoPlay)
+let counter = 0;
+let interval;
+
+function slideForward() {
+  img.src = images[counter % images.length].src;
+  counter++;
+  console.log(counter % images.length);
+}
```
alastair87, line 35: It isn't ideal if your `slideForward()` function relies on continuing to increase the value of counter like this, since eventually this would cause an exception if the number got to be so large that it exceeded the allowed size of an integer.

You could avoid this problem by doing something like this, similar to how you've dealt with it in `slideBack()`, except you need to do it *before* accessing the `img.src` property:

```js
if (counter === images.length - 1) {
    counter = 0;
} else {
    counter++;
}
```

If you use if-else blocks like these to maintain `counter` then you also don't need to do `counter % images.length` as this value will be the same as `counter`.

PR #1016, opened 2020-07-06
Buchra atkeh week 9

Buchraatkeh1984:
# Your Details

Your Name:Buchra Atkeh
Your City:Nort West
Your Slack Name: Buchra Atkeh

# Homework Details

Module:javascript
Week:9


week-9/Homework/mandatory/1-practice/2-code-reading.md
```diff
@@ -61,3 +66,7 @@ console.log(y);
 ```
 
 What will be the output of this code. Explain your answer in 50 words or less.
+//the output will as following:
+9
+{ x: 10 }
+x did not change because it is varibale of primary type, and if we want to change it we need to reassign it ,for example the previous code would change the value of x if we have put x =f1(x);.However for the variable which is not primary type ,we can change it without reassign it as the code above did.
```
Ashaghel, line 72: You'r correct as a primary type variable will pass to the function as a value. However, objects pass by reference. 
This is due to the data structure itself, as in Objects you don't create new stored value but your point to the same reference. For more you can search passing by reference vs by value. And pointer to object.

week-9/Homework/mandatory/1-practice/script.js
```diff
@@ -0,0 +1,60 @@
...
+
+
+    }
+
+    function setIcons(icon, iconId) {
+        const skycons = new Skycons({ color: "white" })
+        const currentIcon = icon.replace(/-/g, "_").toUpperCase();
+        skycons.play();
+        return skycons.set(iconId, Skycons[currentIcon]);
+    }
```
Ashaghel, line 59: Good work!

week-9/Homework/mandatory/2-exercises/1-shopping-cart.js
```diff
@@ -10,11 +10,18 @@ The output of running your code should be:
...
+        }
+        // Add your code here
+    addItem(item) {
+        this.cart.push(item)
+    }
+    cartContains() {
+        console.log(this.cart);
+
+        // Use console.log() to output everything contained in your cart
+    }
```
Ashaghel, line 24: Right to the point

week-9/Homework/mandatory/2-exercises/2-convertion.js
```diff
@@ -20,11 +20,28 @@
...
+            this.name = name;
+        }
+        // Add your code here
+    greeting() {
+        console.log('Hi! I\'m ' + this.name + '.');
+
+    }
+
+    cartContains() {
+        console.log(this.cart);
```
Ashaghel, line 34: some copy paste went wrong here 🤔 
there is not cart for person

week-9/Homework/mandatory/2-exercises/3-atm.js
```diff
@@ -12,7 +12,28 @@
...
 class ATM {
-    // Add your code here
+    constructor() {
+        this.balance = 100;
+    }
+    make_deposit(amount) {
+        if (amount > 0)
+            this.balance += amount;
+        else
+            console.log("you can not add nagative number,please put positive number");
```
Ashaghel, line 22: good addition to validation 
Other checks will be, data type, referencing right functions, ensuring number in a specific format (digits or not)

week-9/Homework/mandatory/2-exercises/3-atm.js
```diff
@@ -12,7 +12,28 @@
...
+    }
+    check_balance() {
+        console.log(this.balance);
+
+    }
+    make_withdrawl(amount) {
+            if ((this.balance >= amount) && (amount > 0))
+                this.balance -= amount;
+            else
+                console.log("Sorryy,your balance is not sufficient to make this withdrawl");
```
Ashaghel, line 33: Good work

week-9/Homework/mandatory/2-exercises/4-music-player.js
```diff
@@ -20,34 +20,50 @@ This means the order the songs are played in will be random, but each song will
...
-    // Add your code here
+    constructor() {
+        this.playing = [];
+        this.MusicPlayer = [];
+        this.index = 0;
+    }
+    add(playing, player) {
+        this.playing.push(playing);
+        this.MusicPlayer.push(player);
+
```
Ashaghel, line 31: you're creating two arrays that are linked to each other (each song from array 1 will have a singer from array 2) to  exclude any issues, adding an array of object will be easier 
something like 
this.playList=[]
this.playlist.push({song:,  singer:})

week-9/Homework/mandatory/2-exercises/4-music-player.js
```diff
@@ -20,34 +20,50 @@ This means the order the songs are played in will be random, but each song will
...
+    }
+    play() {
+        console.log("Currently playing: " + this.playing[this.index] + " by " + this.MusicPlayer[this.index]);
+
+    }
+    skip() {
+        if (this.index < this.playing.length)
+            this.index++;
```
Ashaghel, line 39: this may throw an error as you're incrementing after the check 
so in array of 5 elements 
if the index is 4, the check will pass but as you're incrementing the index after, it will throw an error

week-9/Homework/mandatory/1-practice/2-code-reading.md
zoltan-gal: I don't want to be pedantic but it is called primitive type 😛

week-9/Homework/mandatory/2-exercises/4-music-player.js
```diff
@@ -20,34 +20,50 @@ This means the order the songs are played in will be random, but each song will
...
+        else this.index = 0;
+        console.log("Currently playing: " + this.playing[this.index] + " by " + this.MusicPlayer[this.index]);
```
zoltan-gal, line 41: If you call the play function here then you don't have to repeat this code again :).

week-9/Homework/mandatory/2-exercises/4-music-player.js
```diff
@@ -20,34 +20,50 @@ This means the order the songs are played in will be random, but each song will
...
+    }
+    previous() {
+        if (this.index < this.playing.length)
+            this.index--;
+        else this.index = this.playing.length - 1;
+
+        console.log("Currently playing: " + this.playing[this.index] + " by " + this.MusicPlayer[this.index]);
```
zoltan-gal, line 48: Like above, after setting the index, you simply need to call the play function that will do the rest for you as it is responsible to play the song. Remember the DRY rule!

PR #1019, opened 2020-07-07
Azad week 3

azadaafo:
# Your Details

Your Name:
Your City:
Your Slack Name:

# Homework Details

Module:
Week:

### DIFF
--- a/week-3/Homework/mandatory/1-oxygen-levels.js
+++ b/week-3/Homework/mandatory/1-oxygen-levels.js
@@ -26,9 +26,17 @@ console.log(newdessert); // returns ice cream
 
 /* + + + + + + + */
 /* Now try the exercise */
-
-function safeLevels() {
-
+function safeLevels(arrStr) {
+  //first solution
+  let levelNumber = arrStr.map((level) => {
+    let removingPercentage = level.substring(0, 4);
+    let stringToNumber = parseFloat(removingPercentage);
+    return stringToNumber;
+  });
+  let findOxyArray = levelNumber.find(
+    (findOxy) => findOxy > 19.5 && findOxy < 23.5
+  );
+  return findOxyArray + "%";
 }
 
 /* ======= TESTS - DO NOT MODIFY ===== */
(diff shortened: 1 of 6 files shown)


week-3/Homework/mandatory/1-oxygen-levels.js
```diff
@@ -1,34 +1,29 @@
...
 let newdessert = dessert.substring(0, 9);
-
 console.log(newdessert); // returns ice cream
 
 /* + + + + + + + */
 /* Now try the exercise */
 
 function safeLevels() {
-
+  if (oxygenLevels1 >= "19.5%" && oxygenLevels1 <= "23.5%") {
```
zoltan-gal, line 22: `oxygenLevels1` should have been an input parameter instead of using the global variable from outside.

zoltan-gal: `oxygenLevels1 >= "19.5%"` comparison won't work because `oxygenLevels1` is an array whereas "19.5%" is a string. Also, you have to parse number form the value of each array element and then compare as a number.

week-3/Homework/mandatory/2-bush-berries.js
```diff
@@ -18,18 +18,23 @@
...
 will have a remainder of 0 i.e. they are even numbers*/
 
-array.every((value) => {return (value % 2 == 0)}); /* this will return false as not ALL 
+array.every((value) => { return (value % 2 == 0) }); /* this will return false as not ALL 
 values will have a remainder of 0 i.e. there are some odd numbers in the array too*/
 
 /* + + + + + + + + + + + + + + */
 
 /* Now try to complete the exercise */
+function bushChecker(safeBush) {
```
zoltan-gal, line 30: Here you recognised here that you need input, but you haven't used it

week-3/Homework/mandatory/2-bush-berries.js
```diff
@@ -18,18 +18,23 @@
...
+  if (bushBerryColours1 === "pink") {
```
zoltan-gal, line 31: Again, it is a comparison of an array and a string so it doesn't make sense logically.

PR #1028, opened 2020-07-10
Tinta week 9

tintax01:
# Your Details

Your Name:Tinta
Your City:Manchester
Your Slack Name:Tinta5000

# Homework Details

Module:JS
Week:9

### DIFF
--- a/week-9/Homework/mandatory/2-exercises/1-shopping-cart.js
+++ b/week-9/Homework/mandatory/2-exercises/1-shopping-cart.js
@@ -10,17 +10,19 @@ The output of running your code should be:
 */
 
 class ShoppingCart {
-  // Add your code here
-
-  cartContains() {
-    // Use console.log() to output everything contained in your cart
-  }
+	constructor(toiletRoll, eggs, pasta) {
+		this.toiletRoll = toiletRoll;
+		this.eggs = eggs;
+		this.pasta = pasta;
+	}
+	cartContains() {
+		console.log("Your Shopping cart has 3 Items:");
+		console.log("Toilet Roll " + this.toiletRoll);
+		console.log("Egg" + this.eggs);
+		console.log("Pasta" + this.pasta);
+	}
 }
 
-let myCart = new ShoppingCart(); // Creates an empty shopping cart
-
-myCart.addItem("Toilet Roll");
-myCart.addItem("Pasta");
-myCart.addItem("Eggs");
+let myCart = new ShoppingCart(" ", " ", " ");
 
 myCart.cartContains();
--- a/week-9/Homework/mandatory/2-exercises/2-convertion.js
+++ b/week-9/Homework/mandatory/2-exercises/2-convertion.js
@@ -22,6 +22,15 @@
 */
 
 // Write your code here
+class Person {
+	constructor(name) {
+		this.name = name;
+	}
+	greeting() {
+		console.log("Hi! I'm " + this.name + ".");
+		// alert("Hi! I'm " + simon.name + ".");
+	}
+}
 
 // Do not edit this section
 const simon = new Person("simon");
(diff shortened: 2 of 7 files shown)


week-9/Homework/mandatory/2-exercises/1-shopping-cart.js
```diff
@@ -10,17 +10,19 @@ The output of running your code should be:
...
+		console.log("Pasta" + this.pasta);
+	}
 }
 
-let myCart = new ShoppingCart(); // Creates an empty shopping cart
-
-myCart.addItem("Toilet Roll");
-myCart.addItem("Pasta");
-myCart.addItem("Eggs");
+let myCart = new ShoppingCart(" ", " ", " ");
```
zoltan-gal, line 26: You don't have to remove `addItem` function. This should have been implemented instead.

week-9/Homework/mandatory/2-exercises/1-shopping-cart.js
```diff
@@ -10,17 +10,19 @@ The output of running your code should be:
...
 class ShoppingCart {
-  // Add your code here
-
-  cartContains() {
-    // Use console.log() to output everything contained in your cart
-  }
+	constructor(toiletRoll, eggs, pasta) {
+		this.toiletRoll = toiletRoll;
+		this.eggs = eggs;
+		this.pasta = pasta;
```
zoltan-gal, line 16: This is also a possible solution. But the problem is that in this way you cast in stone the number of elements the cart can store. Also, each of them has a name that indicates what should be there. The shopping cart should be a generic cart, it should be able to store anything. In this exercise, Shopping cart needs an array to where you will be able to add the items using the `addItem` function that should be also implemented.

week-9/Homework/mandatory/2-exercises/1-shopping-cart.js
```diff
@@ -10,17 +10,19 @@ The output of running your code should be:
...
+	}
+	cartContains() {
+		console.log("Your Shopping cart has 3 Items:");
+		console.log("Toilet Roll " + this.toiletRoll);
+		console.log("Egg" + this.eggs);
+		console.log("Pasta" + this.pasta);
```
zoltan-gal, line 22: If you had an array instead of elements then it would have been easier to iterate it over.

week-9/Homework/mandatory/2-exercises/2-convertion.js
```diff
@@ -22,6 +22,15 @@
 */
 
 // Write your code here
+class Person {
```
zoltan-gal, line 25: Good work!

week-9/Homework/mandatory/2-exercises/3-atm.js
```diff
@@ -12,14 +12,29 @@
...
-
+	constructor() {
+		this.accountBalance = 100;
+		this.deposit = 0;
+		return this.accountBalance;
+	}
+	make_deposit(deposit) {
+		this.deposit = deposit;
+		return this.deposit;
+	}
```
zoltan-gal, line 23: I think the deposit should be added to here instead of in the check balance function.
As an extra, you can also validate that the deposit is a number and it is bigger than zero.

week-9/Homework/mandatory/2-exercises/3-atm.js
```diff
@@ -12,14 +12,29 @@
...
+	check_balance() {
+		this.accountBalance += this.deposit;
+		return this.accountBalance;
+	}
+	make_withdrawl(withdraw) {
+		this.withdraw = withdraw;
+		return this.withdraw;
+	}
```
zoltan-gal, line 31: Like above, some extra validation on the input value would be welcomed. For example if it is number and you have enough balance on your account (balance >= withdraw).

PR #1038, opened 2020-07-16
Zision week 9.2

zision71:
Only attempt Exercises for this week, Didn't try Last exercise. Please review.

### DIFF
--- a/week-9/Homework/mandatory/2-exercises/1-shopping-cart.js
+++ b/week-9/Homework/mandatory/2-exercises/1-shopping-cart.js
@@ -10,10 +10,18 @@ The output of running your code should be:
 */
 
 class ShoppingCart {
-  // Add your code here
+  constructor() {
+    this.total = 0;
+    this.array = [];
+  }
+
+  addItem(item) {
+    this.array.push(item);
+    this.total++;
+  }
 
   cartContains() {
-    // Use console.log() to output everything contained in your cart
+    console.log(`Your shopping cart has ${this.total} items : ${this.array} `);
   }
 }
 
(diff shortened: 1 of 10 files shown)


week-9/Homework/mandatory/2-exercises/1-shopping-cart.js
```diff
@@ -10,10 +10,18 @@ The output of running your code should be:
...
+  }
+
+  addItem(item) {
+    this.array.push(item);
+    this.total++;
+  }
 
   cartContains() {
-    // Use console.log() to output everything contained in your cart
+    console.log(`Your shopping cart has ${this.total} items : ${this.array} `);
```
evanrolfe, line 24: To make the output easier to read you can do this instead:
```js
${this.array.join(', ')}
```

week-9/Homework/mandatory/2-exercises/1-shopping-cart.js
```diff
@@ -10,10 +10,18 @@ The output of running your code should be:
...
 class ShoppingCart {
-  // Add your code here
+  constructor() {
+    this.total = 0;
+    this.array = [];
+  }
+
+  addItem(item) {
+    this.array.push(item);
+    this.total++;
```
evanrolfe, line 20: This works, but the code would be better if you only had one instance variable `this.array` and instead got the total from a function i.e.
```js
total() {
  return this.array.length;
}
```

This is better because you don't have to keep track of two variables, you only need one.

evanrolfe:
Nice work :+1:

